Issue #10, opened 2026-07-26
Epic: Documentation architecture and README honesty


ambroise-leclerc:
Bring the top-level documentation into line with the architecture and with the implementation on `develop` after v0.3.0. This is the documentation track of Wave 3.

## Children

- [x] #108 — Re-baseline ADR numbering and supersede Catch2
- [x] #109 — Current architecture reference
- [x] #110 — Supported getting-started guide
- [x] #111 — Rewrite README from built reality
- [x] #65 — Landed AGENTS.md alignment with the current architecture
- [x] #112 — Retire superseded point-in-time documentation
- [x] #113 — Authorized-source verification of IEC 62366-1 mappings
- [x] #114 — Authorized-source verification of IEC 81001-5-1 mappings
- [x] #115 — Internal-link and retired-path CI lint

## Wave 2 RETEX incorporated here

- Generated indexes and schemas are checked independently from prose.
- Ordinary repository links and retired paths become CI-checked, not manual review only.
- Low-confidence corpus mappings are closure blockers until reviewed against authorized sources.
- Current and planned architecture must be labelled separately; no API may be documented unless it exists.

**Status: complete.** All nine children merged to `develop` in #154 and closed.

Both confidence-flagged corpora now have authorized-source review records, and both reviews found
real errors — which is the argument for having done them rather than waived them:

- **IEC 62366-1:2015** — §4 was titled with its own sub-clause's name, §5.1 used the superseded
  IEC 62366:2007 term, and §5.3/§5.4/§5.5/§5.10 were absent entirely, including *user interface of
  unknown provenance*, the clause an integrator is most likely to apply to MduX.
- **IEC 81001-5-1:2021** — verified against the official IEC preview of the published edition plus
  ISH1:2025. A first attempt using the CDV draft was refused, and comparison shows why: security
  risk management is §7 in the draft and §4.2 as published.

The documentation lint now enforces what this epic fixed by hand: internal links, heading anchors
and retired paths are checked in CI, over the root documents as well as `docs/`.

## Done when

No documented API is absent from the tree, the ADR index is contiguous and authoritative, the two confidence-flagged corpora have authorized-source review records, all live internal references resolve, and AGENTS.md/README describe the v0.3.0+ architecture without certification or production-readiness claims.

ambroise-leclerc:
All nine children merged in #154 and closed.

**Done when**, checked against the criteria in the body:

[...]

PR #67, opened 2026-07-26
Delete dead code and orphaned assets from examples/

### LINKED ISSUE #42: S3 — Delete dead code from examples/
`examples/` contains four `.cpp` files; `examples/CMakeLists.txt` builds two targets. The mismatch is worse than it looks.

| File | Status |
|---|---|
| `SimpleMedicalUiExample.cpp` | Built — as target **`MedicalUiExample`** (name mismatch). A CI smoke test: builds a `VulkanContext` with every handle `VK_NULL_HANDLE`, prints `"Vulkan context valid: No (expected in CI)"`, never instantiates `MedicalUiRenderer`. Links GLFW, never opens a window. **Keep for now**; E7/S9 rewrites it. |
| `VulkanSCTriangleExample.cpp` | Built. The only real renderer in the repository. **Keep.** |
| `MedicalUiExample.cpp` (482 L) | **In no target.** A more ambitious GLFW+Vulkan host, full of `// Simplified` markers. **Delete.** |
| `BasicExample.cpp` (58 L) | **In no target, and uncompilable** — `#include <mdux/mdux.hpp>` (no such file) and uses `mdux::WindowConfig` / `mdux::Window` / `mdux::Graphics`, an API generation deleted when the library went pure-Vulkan. **Delete.** |

**Also delete:**

- `examples/ui.html`, `examples/sample_ui.html`, `examples/medical_interface.html` — decorative browser mockups no parser consumes. `ui.html` even encodes window config as CSS (`title: "..."; resizable: true;` inside `body{}`), a scheme nothing reads.
- `examples/shaders/cube_vertex.glsl`, `cube_fragment.glsl`, `examples/shaders/compiled/*.spv` — no target, no consumer.
- `examples/README_CUBE_DEMO.md` — documents a `mdux::UserRendering` / `preUiCallback` / `postUiCallback` API that does not exist in `mdux.cppm`, and a rotating-cube demo with no source file.

Do not touch `examples/shaders/triangle.vert|.frag` — E7/S4 migrates them into the evidence pipeline.

Part of #11.

ambroise-leclerc:
Closes #42.

Removed `BasicExample.cpp` (uncompilable, references a deleted API), `MedicalUiExample.cpp` (in no CMake target), the three decorative `.html` mockups, and the cube-demo shader/README orphans that have no source file.

Verified via grep that nothing in `CMakeLists.txt`, `examples/CMakeLists.txt`, or any other build file references the deleted paths. Two historical docs still *describe* `MedicalUiExample.cpp`; left alone, they're in scope for the docs-archiving epic (#10) instead.

Could not locally build/test (this session's environment is macOS/AppleClang, below the project's GCC15+/Clang20+/MSVC floor for C++23 modules) — this change is deletion-only so there is nothing to compile-check; CI should confirm no dangling reference.


camille-martin-paris:
Follow-up pushed in `6ac7f5b` to address the stale references found during review:

- removed deleted `medical_interface.html` / `ui.html` arguments from active developer commands
- removed obsolete `MduXExample` command permissions tied to the deleted asset
- made `SimpleMedicalUiExample` explicitly asset-free

Validation: `git diff --check`, JSON syntax validation for `.claude/settings.local.json`, and a scoped reference scan all pass.

Issue #12, opened 2026-07-26
Epic: Evidence kernel


ambroise-leclerc:
Implement the common evidence kernel used by every future baker: canonical artifacts are produced offline into the build tree, deliberately updated under `generated/`, committed, and re-derived in CI for byte identity.

## Completed children

- [x] #49 — ADR-007 evidence pipeline doctrine
- [x] #50 — mdux.evidence.digest
- [x] #51 — mdux.evidence.json and evidence lint
- [x] #52 — mdux.evidence.report
- [x] #53 — cmake/MduXBake.cmake
- [x] #54 — tools/common TOML and CLI
- [x] #55 — evidence-labelled CI on active toolchains

## Final implemented decisions

- Bulk bytes use sidecars; canonical JSON encodes floats as u32 bit patterns.
- Normal builds never write tracked artifacts; only mdux-bake-update performs the deliberate copy.
- BakeReport uses semantic toolVersion and deliberately contains no self-referential commit SHA.
- MSVC and GCC 15/16 run evidence checks; Clang is added when #48 restores that leg.

The first production-shaped consumer is the shader baker in #13. It validates the generic surface end to end but is not unfinished work in this epic.

**Blocked by:** nothing.
**Blocks:** resolved; #13, #14, and #18 may use the kernel.

## Done evidence

All children are on develop and the v0.3.0 release commit passed the active MSVC/GCC evidence and clean-tree CI gates.

ambroise-leclerc:
Evidence kernel complete: children #49-#55 are implemented on develop and the v0.3.0 release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed the MSVC/GCC evidence and clean-tree gates in CI run 30717524381. The first real consumer is now tracked by #13.

PR #68, opened 2026-07-26
Fix .gitignore negations for cmake/ and generated/, add CI clean-tree check

### LINKED ISSUE #43: S4 — .gitignore negations for generated/ and cmake/
The evidence pipeline (E6) depends on **committing** generated artifacts. Today `.gitignore` would silently swallow them.

Two existing rules are the problem:

```gitignore
# CMake generated files
*.cmake

# Compiled shaders
*.spv
```

`cmake/*.cmake` and `examples/shaders/compiled/*.spv` are tracked **only because someone force-added them**. The next `cmake/MduXBake.cmake`, `MduXMedui.cmake`, `MduXTrustZones.cmake` and `MduXHostTools.cmake` will not be, and neither will any baked SPIR-V under `generated/`. The failure is silent: `git status` shows nothing, the file never lands, CI fails somewhere unrelated.

**Fix:**

```gitignore
!cmake/*.cmake
!generated/**
```

Narrow `*.cmake` to the actual targets (`cmake_install.cmake`, build-tree output) rather than relying on negations if that reads more clearly.

**Then add the check that makes this stay fixed:** a CI step asserting `git status --porcelain` is empty after a full build. It catches both an ignored artifact that should have been committed and a build that writes into the source tree — the latter being explicitly forbidden by E6/S5's design (`mdux-bake-update` is the only path allowed to).

Part of #11.

ambroise-leclerc:
Closes #43.

- `!cmake/*.cmake`: the blanket `*.cmake` rule currently means `cmake/*.cmake` is tracked only because it was force-added. Confirmed via `git ls-files '*.cmake'` — all 14 files under `cmake/` are tracked despite the ignore rule.
- `!generated/**`: same problem for future baked evidence artifacts (the evidence-pipeline epic, #12, will commit `*.spv`, `*.cmake`-adjacent sidecars, etc. under `generated/`).
- CI: added a "Verify no source-tree writes" step to both the Windows and Linux build jobs, asserting `git status --porcelain` is empty after build+test.

Validated the workflow YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`) since I can't run this CI locally. Could not exercise the actual Windows/Linux jobs — this session's environment is macOS/AppleClang, below the project's toolchain floor.


camille-martin-paris:
Review follow-up pushed in `f4d648d`.

The new Linux clean-tree check failed after build and tests passed because the GCC container rejected the mounted checkout as a dubious-ownership repository. The fix marks `$GITHUB_WORKSPACE` as a safe Git directory inside the container before running `git status --porcelain`.

Validation performed:
- workflow YAML parses successfully
- `git diff --check` passes
- `git check-ignore` confirms `cmake/*.cmake` and nested `generated/**` artifacts are re-included

The Windows failure occurs earlier during CMake configuration because `cl.exe` is missing from `PATH`; it is unrelated to this PR and its new clean-tree step never runs.

camille-martin-paris:
Second CI follow-up pushed in `95be935`: after the safe-directory fix allowed the clean-tree check to run, it correctly detected that the CMake setup step downloaded `cmake-4.1.1-linux-x86_64.tar.gz` into the repository root. The installer now writes that temporary archive to `$RUNNER_TEMP` before extracting it. Workflow YAML and `git diff --check` pass.

Issue #46, opened 2026-07-26
S7 — Split MduXCore out of MduX and add trust-zone verification


ambroise-leclerc:
Create the governed target and the mechanism that keeps it governed.

**New target `MduXCore`** (alias `MduX::Core`), linking `std` only. **No `Vulkan::Vulkan`.** The existing `MduX` target becomes the adapter umbrella and `PUBLIC`-links `MduXCore`, so consumers keep using `MduX::MduX` and nothing breaks downstream.

Because `MduXCore` never receives Vulkan's include directories, `#include <vulkan/vulkan.h>` in a governed module is a hard compile error on every platform. That property is the whole design; do not add Vulkan to `MduXCore` "temporarily" for anything.

**Initial governed contents** — nothing yet moves out of `mdux.cppm`, which stays an adapter module. `MduXCore` starts with the scaffolding later epics fill:

- `mdux.core.units` — `Px`, `Rect`, `Extent2D`, `ColorRgba8`, `LocaleId`
- `mdux.core.result` — error enums and a `Result<T,E>` alias over `std::expected`

**`cmake/MduXTrustZones.cmake`:**

```cmake
mdux_declare_governed(MduXCore)   # records the target in a global property
mdux_verify_trust_zones()         # called at the end of the top-level CMakeLists
```

`mdux_verify_trust_zones()` transitively walks `LINK_LIBRARIES` and `INTERFACE_LINK_LIBRARIES` of every governed target and `message(FATAL_ERROR)`s on `Vulkan::*`, `glfw`, or any target not itself governed. This is the crate-boundary substitute, and it runs at **configure** time on every build — before anyone compiles anything.

**Also add `include/mdux/core/.clang-tidy`**, inheriting the root config and adding the governed profile from E5/S1 with `WarningsAsErrors: '*'`.

Note the existing `mdux.vulkansc.memory` and `mdux.vulkansc.objects` modules stay in the adapter zone — they take `VkDevice` in their public APIs by design.

Part of #11.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

Issue #41, opened 2026-07-26
S2 — ADR: error handling and exceptions policy


ambroise-leclerc:
The repository has no stated policy, and the current code is inconsistent with the direction this programme needs: `src/mdux.cpp:221` throws `std::runtime_error` from `MedicalUiRenderer`'s constructor, `include/mdux/vulkansc/MemoryPoolManager.cppm:100` documents throwing, and `.clang-tidy:9` *disables* `bugprone-exception-escape`.

**Proposed policy:**

| Zone | Rule |
|---|---|
| Governed | `std::expected` and `noexcept` throughout. No throwing. |
| Adapter | Exceptions permitted at construction boundaries only, never across a `noexcept` render path. |
| Host tools | Exceptions freely — they are not shipped. |

**The rationale that matters more than taste:** Class C and Vulkan SC deployments routinely build with `-fno-exceptions`. A governed library that cannot be compiled that way is not deployable, regardless of how clean its exception hierarchy is.

`std::expected` is available on MSVC 19.33+, GCC 12+ and Clang 16+ — all comfortably below this project's enforced floor of MSVC 19.40 / GCC 15 / Clang 20.

**Make it a link error, not a review comment:** compile `MduXCore` (or at minimum `src/ml/`) into an object library with `-fno-exceptions -fno-rtti` / `/EHs-c- /GR-`. Re-enable `bugprone-exception-escape` in the governed `.clang-tidy`.

Error types should carry evidence, not just a code — see `MlError` in E12/S7, which records the diverging layer, golden vector, element index and both bit patterns. That struct *is* the audit record when a device fails closed.

Existing `MedicalUiRenderer` behaviour is grandfathered until E7/S9 deletes it.

Part of #11.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

Issue #38, opened 2026-07-26
S5 — software_development_file/regulatory/


ambroise-leclerc:
The same documents as E3/S4, filled in **for MduX itself** — TrustSC's `software_development_file/regulatory/` mirror. This is what makes the templates credible: a worked example the manufacturer can read alongside the blank.

```
software_development_file/regulatory/
  IEC_62304/{SAD.md, SDD.md, SOUP.md}
  IEC_62366/Usability_Engineering_File.md
  IEC_81001/Cybersecurity_SAD.md
  ISO_13485/README.md
  ISO_14971/Risk_Management_File.md
```

The SAD is where the trust-zone architecture (E5/S1), the compile-time-only `.medui` boundary (E9/S1), and the evidence pipeline (E6/S1) get written up against IEC 62304 §5.3, with `Justification` blocks citing real clauses and real repository paths.

**The constraint that matters most here**, and it must not be softened for the sake of a nicer-sounding document: the C++ trust zones do **not** reproduce `#![forbid(unsafe_code)]`'s audit property. The defensible claim is that governed modules are compiled without access to platform, graphics or OS headers, are checked by an enforced static-analysis profile, and are covered by determinism tests — **not** that they cannot contain undefined behaviour. Write it that way in the SAD.

Part of #9.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The governance/traceability API, SOUP register, SDF templates, regulatory records, schemas, and tests are present; post-merge CI run 30716625565 passed all seven jobs.

Issue #18, opened 2026-07-26
Epic: Zero-SOUP ML inference


ambroise-leclerc:
Implement zero-SOUP, governed f32 inference using the evidence kernel and one shared kernel implementation for authoring and runtime.

## Current readiness

The former #12 blocker is complete: MduXCore, SHA-256, canonical JSON, BakeReport, TOML/CLI tooling, bake registration, and active-toolchain evidence CI exist. This epic may proceed in parallel with #13.

## Children

- [x] #56 — ADR: zero-SOUP ML inference
- [x] #57 — mdux.ml.schema
- [x] #58 — governed kernels
- [x] #59 — determinism flags and cross-toolchain evidence
- [x] #60 — Safetensors parser and architecture validation
- [x] #61 — golden generation, ML baker, and committed demonstrator
- [x] #62 — fail-closed Classifier1D runtime
- [x] #63 — no-heap verification
- [x] #64 — blob embedding, demonstrator, and weight-swap test

## Updated integration constraints

- #57 is the canonical model; host and runtime consumers import it and never duplicate its records.
- Every baker uses the current BakeReport: semantic toolVersion, resolved options, repository-relative paths, and no commit SHA.
- JSON diagnostics use the shared #19 envelope, including column once available.
- Cross-toolchain evidence currently means MSVC plus GCC 15/16; add Clang after #48 restores it rather than claiming it already runs.
- Dependent PRs follow #11's stack-safe rebase and post-merge-CI policy.

**Status: complete.** All nine children merged to `develop` (PRs #143-#152) and closed.

Delivered: ADR-008; `mdux.ml.schema`, `mdux.ml.kernels` and `mdux.ml.runtime` in the governed zone;
configure-time floating-point determinism enforcement plus a frozen cross-toolchain bit-pattern
contract run on both CI legs; the safetensors importer and architecture validator; `mdux-mlbake`
with the committed, byte-verified `generated/model/ecg-demo/`; `Classifier1D` with its fail-closed
golden self-test; no-heap-in-`predict` verified three independent ways; and `mdux_embed_blob()`
with the ECG demonstrator and the weight-swap proof.

Verified on `develop` from a clean build: 360/360 tests pass under GCC 16.1, including
`ctest -L evidence` (both committed model packages re-bake byte-identically), `-L determinism` and
`-L noheap`.

Two deviations from the issue text, both recorded in ADR-008 and in the PRs:

[...]

PR #69, opened 2026-07-26
Add ADR-004 (trust zones in C++) and ADR-005 (error handling policy)

### LINKED ISSUE #40: S1 — ADR: trust zones in C++
Record the target-family model that substitutes for TrustSC's `crates/` / `adapters/` / `tools/` split (its ADR-005), and — critically — be precise about what it does and does not guarantee.

| Zone | Target | Links | Public API may contain |
|---|---|---|---|
| Governed | `MduX::Core` (`MduXCore`) | `std` only | `std::span`, `string_view`, PODs |
| Adapter | `MduX::MduX` | `MduXCore` + `Vulkan::Vulkan` + `Threads` | `Vk*` handles, owned governed types |
| Host tools | `mdux-*` under `tools/` | `MduXCore` + host-only object libs | n/a |

**Mechanically enforced — real:**

1. **Include-directory absence.** `MduXCore` never receives Vulkan's include dirs, so `#include <vulkan/vulkan.h>` in a governed module is a compile error on every platform, with no tooling required. This is the strongest guarantee available and is worth more than every linter combined.
2. **Configure-time link-graph assertion** (S7) — transitively walks governed targets' link interfaces and `FATAL_ERROR`s on `Vulkan::*` or `glfw`.
3. **Per-directory `.clang-tidy`** with `WarningsAsErrors`, adding the `cppcoreguidelines-pro-type-*`, `-pro-bounds-pointer-arithmetic`, `-no-malloc` and `-owning-memory` checks.
4. **A CI grep lint** over governed directories banning `reinterpret_cast`, `const_cast`, `new`, `malloc`, `<random>`, `std::chrono::*_clock::now`, `std::fma`, `getenv`, `std::filesystem`. Crude — and the only thing that catches `<random>` in a `.cppm` clang-tidy skipped.
5. **Float-determinism flag guards** (see E12/S4).

**Convention only — the ADR must say so:**

- Nothing stops a governed module writing UB inside its own translation unit. There is no `unsafe` keyword and no borrow checker.
- `import std;` gives governed code `std::vector`, threads and I/O. Only the grep lint and review keep them out.
- clang-tidy's analysis of C++23 module interface units is immature; run it in the Clang leg only and treat it as advisory elsewhere.

**The claim, stated exactly:** *governed modules are compiled without access to platform, graphics, or OS headers, are checked by an enforced static-analysis profile, and are covered by determinism tests.* **Not** *governed modules cannot contain undefined behaviour.* Do not let this be written up as equivalent to `#![forbid(unsafe_code)]` in any regulatory document.

Part of #11.

ambroise-leclerc:
Closes #40, #41.

Both ADRs follow the existing template shape (Status/Context/Medical Device Considerations/Decision/Alternatives/Consequences/References/Approval) used by ADR-001 and ADR-003.

ADR-004 fixes the exact wording for the narrower-than-Rust regulatory claim (governed modules compiled without platform/graphics/OS headers + static analysis + determinism tests — **not** UB-free) that #38 and #9's SDF work depend on getting right.

ADR-005 is a prerequisite for #46 (MduXCore split) and #18 (ML inference, whose `Classifier1D::create()` needs `std::expected`).

Also patched `docs/adr/README.md`'s index, which never listed ADR-002/003 and didn't flag the pre-existing ADR-002 filename duplication — noted but not resolved here (that's #10/S1).


camille-martin-paris:
Review follow-up pushed in `af87a67` to correct two regulatory/toolchain overclaims:

- Not linking `Vulkan::Vulkan` does **not** make system-installed Vulkan/platform headers unreachable. ADR-004 now distinguishes target dependency isolation from the mandatory banned-include CI check and narrows the exact compliance claim accordingly.
- `-fno-exceptions` violations are build/compile failures, not link failures. ADR-005 now specifies source checks and MSVC diagnostics-as-errors, and explicitly bans throwing `std::expected::value()` in governed code rather than relying on `bugprone-exception-escape` to catch it.

Validation: `git diff --check` passes, and both ADRs retain all required template sections. Linux, compliance, and security checks were green before this documentation-only follow-up; the Windows `cl.exe` PATH failure is pre-existing and unrelated to this PR.

Issue #21, opened 2026-07-26
S2 — Inventory files containing reproduced or paraphrased standard text


ambroise-leclerc:
Produce a reviewed inventory before deleting anything, so the decision in S3/S4 is made against a complete picture.

Known starting points:

| Path | Size | Note |
|---|---|---|
| `docs/MduX-IEC-62304-AI-Reference.md` | 2,950 L | Opens: *"a comprehensive markdown version of IEC 62304:2006, compiled from public source materials"* |
| `docs/MduX-ISO-13485-AI-Reference.md` | 2,604 L | Same pattern |
| `inputs/Documentation/IEC-DIS-62304-2.pdf` | 1.8 MB | The draft standard. **Tracked** — `git ls-files inputs/` confirms |
| `inputs/Documentation/IEC-62304-Complete.md` | 49 KB | Tracked |
| `inputs/Documentation/ISO-13485-Complete.md` | 118 KB | Tracked |
| `inputs/Documentation/Medical Device Cybersecurity for Engineers and Manufacturers.md` | 47 KB | A copyrighted book |

Note the `.gitignore` block for `inputs/Documentation/*.pdf` postdates the files, so ignoring them never untracked them.

Then sweep the rest of the tree — the four `docs/iec62304/0*.md` and `docs/iso13485/0*.md` modules, `docs/MduX_*-Framework.md`, and `risk-assessment-templates.md` (935 L) — and classify each passage as original prose, close paraphrase, or reproduction. Close paraphrase counts.

**Deliverable:** a table in the issue, one row per file, with a disposition (keep / rewrite / delete).

Part of #7.

ambroise-leclerc:
Done via #86 — see docs/governance/reproduced-text-inventory.md.

PR #70, opened 2026-07-26
Add ADR-006: no reproduction of normative standard text

ambroise-leclerc:
Closes #20.

Stacked on #69 (branched from it to avoid an index-table merge conflict in `docs/adr/README.md` — rebase onto `develop` once #69 merges, or merge #69 first).

Documents the specific violations already in the tree (both `MduX-*-AI-Reference.md` files, four tracked files under `inputs/Documentation/` including a 1.8 MB draft-standard PDF and a copyrighted book) and fixes the citation-key format and valid standard identifiers. Removal (#22) and any git-history purge (#23) are separate, deliberately not authorized by this ADR alone.

Issue #24, opened 2026-07-26
S5 — Delete inputs/ and libstd/


ambroise-leclerc:
Both directories are referenced by no CMake file and no `#include` anywhere in the project.

**`inputs/`** — a scratch drop that has outlived its purpose. Beyond the standards material handled in S3/S4:

- `webfrontCmakelists.txt` — the `CMakeLists.txt` of a **different project** (WebFront, a CEF/Chromium-Embedded web-UI library), kept as the template this build system was derived from. Likely the origin of the HTML/CSS-driven-UI concept that E7/S9 retires.
- `CompilerSettings.cmake`, `Linker.cmake` — earlier copies of files now in `cmake/`, since diverged (the repo versions gained `configure_medical_compliance` and `configure_vulkan_definitions`).
- `Logo.png`, an empty `CMake/` directory.

**`libstd/`** — ~34 umbrella headers forwarding to 30 files in `libstd/include/h/` (~3,185 lines). Its own `readme.md` describes it: *"A C++14 Standard Library implementation for ETL … since the avr 8-bit toolchain does not provide a standard C++ standard library."* It is vendored from an unrelated AVR project, is C++14 and freestanding-oriented, and directly contradicts this project's `import std;` + C++23 direction.

If `libstd/` is being kept as a future freestanding/certified-runtime option, that intent belongs in an ADR, not in an inert directory. Otherwise delete it.

Part of #7.

ambroise-leclerc:
Done via PR #71 (`regulatory/purge-reproduced-standard-text`), same commit as #22 — `inputs/` and `libstd/` no longer exist on develop. Not closed at merge time; closing now.

PR #71, opened 2026-07-26
Remove reproduced standard text, delete inputs/ and libstd/

### LINKED ISSUE #22: S3 — Remove reproduced standard text from the working tree
Act on the S2 inventory: delete or rewrite every file classified as reproduction or close paraphrase.

Land `docs/governance/citation-convention.md` in the same change, so the tree never sits in a state where the material is gone and nothing explains what replaces it. The replacement mechanism is E2's clause-accurate corpus; this issue only has to remove and point forward.

Note that deleting `docs/MduX-IEC-62304-AI-Reference.md` and `docs/MduX-ISO-13485-AI-Reference.md` removes the current AI-Reference indexes. E2/S7 rebuilds them in the correct form — one row per clause, one-sentence pointer, deep link into the module — so this is a temporary gap, and the sequencing should be stated in the commit message.

Part of #7.

ambroise-leclerc:
Closes #22, #24.

Related to #21, but does not close it: the required reviewed inventory table covering the broader regulatory corpus remains outstanding.

Stacked on #70. Merge order: #67, #68, #69, #70, then this PR.

Deleted:
- The two legacy MduX AI-reference transcriptions.
- The four reported standards/book files under inputs/Documentation.
- The remainder of the unreferenced inputs/ scratch tree, including the broken CMake gitlink.
- The unrelated vendored C++14 AVR libstd implementation.

Added docs/governance/citation-convention.md as the replacement citation policy.

Fixed dangling references in examples/README-VulkanSCTriangle.md and updated the repository regulatory-citation skill so it no longer claims the deleted AI-reference files are pending removal.

Known gap: docs/iec62304/ and docs/iso13485/ still require issue #21’s file-by-file inventory and issue #8’s clause-accurate rewrite. The citation convention no longer repeats suspected close-paraphrase wording while documenting that gap.


camille-martin-paris:
Review follow-up pushed in b1c5c9f.

- updated the regulatory-citation skill so it no longer directs agents to deleted AI-reference files
- corrected its trust-zone example to avoid the already-reviewed false claim that system headers are physically inaccessible
- removed the suspected close-paraphrase quotation from the new citation convention while retaining the known-gap warning
- made issue #21’s required file-by-file inventory explicitly remain open

I also updated the PR description from “Closes #21, #22, #24” to close only #22 and #24. Issue #21 requires a reviewed one-row-per-file inventory across the broader corpus; this PR records only a spot check, so auto-closing it would lose required work.

Validation: all deleted-path references were scanned, the broken inputs/CMake gitlink is gone, no tracked gitlinks remain, skill frontmatter/content assertions pass, and git diff --check passes.

Issue #25, opened 2026-07-26
S6 — mdux-docs-lint CI check


ambroise-leclerc:
The MduX equivalent of `cargo run -p trustsc-docs-lint -- check`, which TrustSC runs in CI after any change under `docs/` or `software_development_file/`.

**Checks:**

1. Every citation key matches `<Standard> §<clause> <Short clause title>` and uses one of the five valid standard identifiers.
2. Every cited clause exists in that standard's `AI-Reference.md` index.
3. Every `Justification` block validates against `justification.schema.json`; `justification_id` values are unique corpus-wide; every path in `evidence_refs[]` exists in the repository.
4. Heuristic reproduction detection — flag files that self-describe as a version/transcription of a standard, or that contain long runs of numbered normative-style clauses. This will produce false positives; make the rule list explicit and overridable per file with a documented marker.

Wire into `.github/workflows/ci.yml` as its own job so a docs-only PR does not wait on a full C++ matrix build.

Depends on S1 (the format to check against) and E2/S1 (the schema). Can be built against the format before the corpus is finished.

Part of #7.

ambroise-leclerc:
Done via PR #72 (`regulatory/docs-lint`) — `mdux-docs-lint` runs in the `docs-lint` CI job. Not closed at merge time; closing now.

PR #72, opened 2026-07-26
Add mdux-docs-lint and wire it into CI

ambroise-leclerc:
Part of #25; does not close it yet.

This PR introduces the first working mdux-docs-lint subset and a dedicated Documentation Lint CI job. It validates approved standard identifiers in recognized citation keys, Justification structure and uniqueness, repository-contained evidence paths, and explicit reproduced-text markers.

Review follow-up ce6db8f makes evidence_refs required and non-empty, rejects absolute and parent-traversal evidence paths, checks clause_ref/standard consistency, and adds committed unittest regression coverage that runs in CI without writing Python bytecode into the checkout.

Remaining #25 acceptance work depends on the unfinished regulatory corpus/schema:
- verify every cited clause against each standard AI-Reference index;
- validate against justification.schema.json;
- detect long numbered normative-style runs;
- broaden malformed citation candidate detection.

The current implemented checks and four regression tests pass against this branch.


camille-martin-paris:
Review follow-up pushed in ce6db8f.

The original linter treated evidence_refs as optional and accepted paths outside the repository, weakening traceability. The fix requires a non-empty evidence list, rejects absolute/parent-traversal paths, validates clause_ref format and standard consistency, and adds four committed regression tests wired into CI.

I also changed the PR description from Closes #25 to Part of #25. Clause-index lookup, JSON Schema validation, long normative-run detection, and broad malformed-citation detection remain unimplemented, so auto-closing the issue would discard explicit acceptance work.

Impact classification: not safety-relevant; this changes host-only documentation validation, not runtime behavior. Verification: unittest suite passes, the full current corpus lint passes, workflow YAML parses, and git diff --check passes.

Issue #48, opened 2026-07-26
S9 — Unblock the compilers (GCC 15 ICE, Clang CI leg)


ambroise-leclerc:
Part of #11. GCC 16 CI and the optimization-dependent memory-test workaround are complete; this issue now tracks the remaining compiler coverage gaps.

## Remaining work

1. Minimize and report the renderer/modules ICE upstream with the exact affected GCC version range.
2. Correct `examples/CMakeLists.txt`: the current `VERSION_GREATER_EQUAL 15.0` condition disables `VulkanSCTriangleExample` on GCC 16 as well as GCC 15 while the message says only GCC 15. Enable it on every verified unaffected compiler and scope any workaround narrowly.
3. Restore the Linux Clang 20+ CI leg through the `ninja-clang` preset, including build, full tests, evidence-labelled tests, and source-tree-cleanliness verification.
4. Update ADR-007 and active documentation only after Clang is genuinely green; until then the byte-identity claim is MSVC plus GCC 15/16.

The generated C-array emitter rule is implemented by #13's emitter child, where generated code exists; it is no longer acceptance work here.

## Acceptance

- MSVC, GCC 15, GCC 16, and Clang CI are green.
- Example exclusions name and match only reproduced compiler defects.
- Evidence tests run on Clang and no documentation claims coverage before it exists.

PR #73, opened 2026-07-26
Add CMake presets for GCC, Clang, and MSVC/GCC Debug

### LINKED ISSUE #45: S6 — CMake presets for Linux and Clang
`CMakePresets.json` (v3) defines exactly one preset — `ninja-msvc` — gated on `hostSystemName == Windows`.

There is no Linux preset and no macOS preset, so a Linux contributor has no supported path and must reconstruct the configure line from `.github/workflows/ci.yml`:

```bash
cmake -B build -S . -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DMDUX_BUILD_EXAMPLES=ON -DMDUX_BUILD_TESTS=ON -DMDUX_BUILD_DOCS=OFF
```

**Add:** `ninja-gcc`, `ninja-clang`, `ninja-msvc-debug`, `ninja-gcc-debug`, plus matching build and test presets.

Two things to get right while in here:

- **Ninja is mandatory.** The root `CMakeLists.txt` `FATAL_ERROR`s under the Visual Studio generator when `CXX_MODULE_STD` is unavailable. Presets are the right place to make that non-negotiable rather than a runtime surprise.
- **`cmake/SetupStdModule.cmake` exists but is never `include()`d.** It downloads libc++'s `std.cppm` for macOS/Clang. Since macOS is `FATAL_ERROR`-blocked anyway, either wire it into the Clang path where it is useful or delete it — an unreferenced helper that appears to solve the `import std` bootstrap is a trap for the next person.

CI should build from the presets rather than duplicating flags, so the two cannot drift.

Part of #11.

ambroise-leclerc:
Closes #45.

Adds `ninja-gcc`, `ninja-clang` (Linux-gated), `ninja-msvc-debug`, `ninja-gcc-debug` (via `inherits`) alongside the existing Windows-only `ninja-msvc`. Cache variables mirror `.github/workflows/ci.yml`'s actual configure lines for the active Linux job and the currently-disabled Clang job, so `ninja-clang` is ready the moment #48 re-enables that leg.

Verified `cmake --list-presets=all` (CMake 4.3.1) resolves all 5 build/test presets and their `inherits` chains cleanly. Could not run an actual configure — this session's environment is macOS/AppleClang, below the toolchain floor and outside the Windows/Linux `condition` gates by design (confirmed via `git stash` that the empty configure-preset listing on this machine is pre-existing behavior, not a regression). CI or a Linux/Windows machine is needed to exercise these end-to-end.

Follow-up noted, not done here: once #6/AGENTS.md merges, its §5 ("no preset exists on Linux") should be updated to reference these.


camille-martin-paris:
Review completed and fixes pushed in bc8f4c0.

Changes made:
- exercised `ninja-msvc` and `ninja-gcc` directly in the active CI configure/build/test steps
- removed `CMAKE_GENERATOR_PLATFORM`, which is incompatible with Ninja, and aligned the MSVC preset with the Release CI configuration
- ignored preset-specific `build-*` directories so the clean-tree CI assertion remains valid

Validation: `CMakePresets.json` and `.github/workflows/ci.yml` parse successfully, and `git diff --check` passes. A local configure could not be run because CMake is not installed in this workspace; the pushed GitHub checks will exercise both presets. The obsolete `cmake/SetupStdModule.cmake` mentioned by #45 is already absent.

PR #74, opened 2026-07-27
Fix GCC modules ICE in vulkansc_memory_tests, add GCC 16 CI leg

ambroise-leclerc:
Part of #48.

## Scope

This PR handles a second, optimization-dependent GCC modules ICE in `vulkansc_memory_tests`, distinct from the existing `std::array` renderer ICE. It applies a GNU-only `-O0` override to that test target and adds a GCC 16 CI leg.

The ICE was reproduced on GCC 15.2.0 and GCC 16; `-O0` succeeds while `-O1`, `-O2`, `-O3`, and `-Og` fail. Full Release builds and all tests passed on both versions, and the PR checks exercise Windows/MSVC, GCC 15, GCC 16, compliance docs, and security analysis.

## Remaining work in #48

- minimise and report the original renderer `std::array`/modules ICE upstream
- adopt the generated-code C-array emitter rule
- re-enable and validate the Clang CI leg with libc++


camille-martin-paris:
Review completed. The targeted `-O0` workaround is correctly scoped to `vulkansc_memory_tests` and GNU compilers, the GCC 16 job mirrors the active Linux build, downloads CMake outside the checkout, and all five PR checks pass.

No source modification was needed. I corrected the PR body from a contradictory `Closes #48 (partially)` claim to `Part of #48` and listed the remaining acceptance work: the original renderer ICE/upstream report, the C-array emitter rule, and Clang CI re-enablement.

PR #75, opened 2026-07-27
Add MduXTest framework, port all tests off hand-rolled TestRunner

### LINKED ISSUE #44: S5 — A real test framework (MduXTest), superseding ADR-002
`tests/CMakeLists.txt:3` says it plainly: `# Find or fetch a testing framework (placeholder for now)`. There is none.

**Today:** 21 tests across four executables, using **two different** hand-rolled runners that coexist — a bool-returning `TestRunner::runTest(const char*, bool)` in an anonymous namespace (added in `5228a4a` to fix an ODR violation), and a richer `function<void()>` runner in `tests/vulkansc/` that catches exceptions and times in µs.

Of the 21, roughly 14 assert that `constexpr` literals equal themselves — `mdux::Version::major == 0`, `Compliance::standards == "IEC 62304, IEC 62366"`. `MedicalUiRenderer`, `UiFileWatcher` and every UI path have **zero** tests.

**ADR-002 selected Catch2 v3 and was never implemented.** That decision now conflicts with the zero-SOUP doctrine this programme adopts: a library whose whole argument is "no unaudited third-party code in the qualified path" should not depend on a 40 kLOC test framework it must itself qualify.

**Proposal — supersede ADR-002 with an in-repo `MduXTest`:**

- `tests/framework/MduXTest.cppm`, ~300 lines
- `CHECK` / `REQUIRE` / a `SECTION`-lite
- one ctest registration per test case, so `ctest -R` selects individual cases
- test labels (`evidence`, `pixel`, `determinism`) — E6/S7, E7/S8 and E12/S4 all depend on label-based selection
- `--format=json` output for agent consumption (E13/S3)

Port all 21 existing tests; delete both old `TestRunner` classes.

**This blocks E6 and everything after it.** Every epic downstream defines its acceptance in terms of a test that does not currently have a framework to live in.

Part of #11.

ambroise-leclerc:
Closes #44.

Stacked on #74 (branched after the GCC-ICE fix to avoid duplicating that workaround across two PRs — merge #74 first, or rebase this one).

## What this replaces

Two different hand-rolled `TestRunner`/`ComplianceTestRunner` classes (`tests/TestMain.cpp`, `tests/ComplianceTestMain.cpp`, `tests/vulkansc/MemoryPoolManagerTests.cpp`, `tests/vulkansc/DeviceObjectManagerTests.cpp`) — one a bool-returning `runTest(name, bool)`, the other a richer `function<void()>`-based one with timing. `tests/CMakeLists.txt:3` said outright: *"Find or fetch a testing framework (placeholder for now)"*. Supersedes ADR-002 (Catch2, never implemented) per the rationale already in ADR-004/005: a zero-SOUP-direction library shouldn't depend on a 40kLOC framework it would itself need to qualify.

## What's new

- **`tests/framework/MduXTest.cppm`** (module `mdux.test`) — registry, check recording, runner, text/JSON reporters. Governed-adjacent in spirit (no Vulkan, no filesystem) though it lives under `tests/`, not `include/mdux/`, since it's test-only and never shipped.
- **`tests/framework/MduXTest.hpp`** — `TEST_CASE`/`CHECK`/`REQUIRE`/`CHECK_MESSAGE`/`REQUIRE_MESSAGE`/`MDUX_TEST_MAIN` macros. Modules can't export macros, so these live in a companion header alongside `import mdux.test;` — same split Catch2's own experimental-modules support uses.
- **`cmake/MduXTestDiscovery.cmake` + `MduXTestDiscoveryImpl.cmake`** — `mdux_discover_tests(<target>)`, a POST_BUILD step that runs the binary with `--list-tests` and generates one `add_test()` per case, so `ctest -R <name>` selects an individual test instead of only the whole executable.
- **`--format=json`** on every test binary, for agent consumption per the parity programme's agent-tooling direction.

All 21+ original test functions ported 1:1 (some split into more granular cases — e.g. the old combined `testRegulatoryCompliance()` is now 4 separate `TEST_CASE`s) to **27 discovered ctest entries**.

## Verified on real hardware

This session's usual environment (macOS/AppleClang) can't compile this project. Built and ran on the Ubuntu 26.04 box from #74/#48:

- Full clean build + `ctest --output-on-failure`: **27/27 passed**, on both **GCC 15.2.0 and GCC 16**.
- `ctest -R 'MemoryPoolCalculator Redundancy'` correctly selects and runs just that one case.
- `--format=json` output verified — correct structure, per-check expression/file/line captured.
- **Deliberately broke a test** (changed `major == 0` to `major == 99`) to confirm failure detection actually works: correct ❌, correct file:line, correct expression text, exit code 1. Then restored it.

## Two real bugs found and fixed along the way

1. **A plain `.cppm` listed as a regular executable source is not enough** — CMake 4.2.3 requires `target_sources(<target> PRIVATE FILE_SET CXX_MODULES FILES ...)` even for a module used only inside one executable, not just for installable library modules. Failed with `CMake Error: ... provides the `mdux.test` module but it is not found in a `FILE_SET` of type `CXX_MODULES`` until fixed.
2. **`add_test(NAME ... COMMAND ...)` mis-parses on this CTest 4.2.3** — reproduced with a two-line minimal repro outside the project entirely (`add_test(NAME simplename COMMAND /bin/echo hello)` on its own in an isolated `CTestTestfile.cmake`): CTest treats the literal word `NAME` as the test's display name and the actual name as the executable path (`Could not find executable simplename`). Old-style positional `add_test(<name> <command> [args...])` works correctly, including with spaces and colons in the name — confirmed with the same isolated repro before switching. Documented in code so nobody "modernizes" it back without re-checking.


tests/framework/MduXTest.hpp
```diff
@@ -0,0 +1,59 @@
...
+ *   MDUX_TEST_MAIN("MemoryPoolManager Tests")
+ */
+#pragma once
+
+#define MDUX_TEST_CONCAT_INNER(a, b) a##b
+#define MDUX_TEST_CONCAT(a, b) MDUX_TEST_CONCAT_INNER(a, b)
+
+/// Defines a test case function and registers it at static-init time. `labels...`
+/// is an optional trailing list of string labels (e.g. "evidence", "pixel") that
+/// later tooling can filter on; the registry stores them even though nothing
```
ambroise-leclerc: [P1] Issue #44 requires label-based selection because downstream evidence/pixel/determinism suites depend on it, but labels are currently dead metadata: `--list-tests` emits only names, discovery never sets CTest `LABELS`, and the runner has no label filter. On this head, `ctest -N -L regulatory` reports 0 tests even though four cases declare `"regulatory"`. Please carry labels through the discovery protocol and register them with CTest (and/or provide the promised runner filter) before closing #44.

tests/framework/MduXTest.hpp
```diff
@@ -0,0 +1,59 @@
...
+        (name), {__VA_ARGS__}, &MDUX_TEST_CONCAT(mduxTestFn_, __LINE__), __FILE__, __LINE__};     \
+    }                                                                                             \
+    static void MDUX_TEST_CONCAT(mduxTestFn_, __LINE__)()
+
+/// Records the outcome and continues the test case even if it fails.
+#define CHECK(expr) ::mdux::test::checkImpl(static_cast<bool>(expr), #expr, __FILE__, __LINE__, false)
+
+/// Records the outcome and, on failure, aborts the current test case (the
+/// remaining body of the TEST_CASE does not run) without aborting the run.
+#define REQUIRE(expr) ::mdux::test::checkImpl(static_cast<bool>(expr), #expr, __FILE__, __LINE__, true)
```
ambroise-leclerc, line 45: [P2] Issue #44 also explicitly scopes a `SECTION`-lite facility, but this public macro header exposes only TEST_CASE/CHECK/REQUIRE and message variants. Either implement the requested section mechanism and exercise it, or keep #44 open and track the omission explicitly rather than closing the issue with part of its proposed API absent.

camille-martin-paris:
Review completed and fixes pushed in 897a4c3.

The original patch ported the tests and provided discovery/JSON reporting, but two #44 acceptance items were not functional: labels were stored only in memory and SECTION-lite was absent. The follow-up now:
- emits test labels through the discovery protocol and assigns them as CTest `LABELS`, enabling `ctest -L ...`
- adds a documented single-pass `SECTION` macro
- makes JSON escaping valid for backspace, form-feed, and other control bytes

`git diff --check` passes. Local compilation was not attempted because this workspace lacks CMake and has GCC 13.3 while MduX requires GCC 15+; the pushed Windows, GCC 15, and GCC 16 checks will provide build/test evidence.

tests/framework/MduXTest.hpp
ambroise-leclerc: Fixed in 897a4c3: labels now flow name<TAB>comma-labels through --list-tests, MduXTestDiscoveryImpl.cmake parses them and calls set_tests_properties(... PROPERTIES LABELS ...), so `ctest -L regulatory` selects the 4 regulatory-labeled cases. CI (Linux GCC15/16, Windows) has rebuilt and passed on this commit.

tests/framework/MduXTest.hpp
ambroise-leclerc: Fixed in 773ae27: added a real exercise ("Version Test Sections" in tests/TestVersion.cpp) with two sibling SECTIONs and a shared counter proving both run in the same single-pass invocation, not implemented-but-unused as before.

PR #76, opened 2026-07-27
Add MduXCore governed target with mechanical trust-zone verification

### LINKED ISSUE #46: S7 — Split MduXCore out of MduX and add trust-zone verification

ambroise-leclerc:
Closes #46.

Stacked on #75 (branched after MduXTest so `core_tests` could use the discovery mechanism immediately — merge order: #74, #75, this).

Implements ADR-004 (already open in #69): a new `MduXCore` target that cannot reach Vulkan, checked mechanically rather than by convention.

## What's new

- **`MduXCore`** (alias `MduX::Core`) — the governed zone. Does **not** link `MduX_options` (which carries `Vulkan::Vulkan`); links only `MduX_warnings` (pure compiler-warning flags, confirmed Vulkan-free).
- **`include/mdux/core/Units.cppm`** (`mdux.core.units`) — `Px`, `Extent2D`, `Rect`, `ColorRgba8`, `LocaleId`. Simple value types, `std`-only.
- **`include/mdux/core/Result.cppm`** (`mdux.core.result`) — `Result<T,E>`/`ResultVoid<E>` aliases over `std::expected`, per ADR-005.
- **`cmake/MduXTrustZones.cmake`** — `mdux_declare_governed(<target>)` + `mdux_verify_trust_zones()`, walking the full transitive `LINK_LIBRARIES`/`INTERFACE_LINK_LIBRARIES` graph of every governed target and `FATAL_ERROR`ing on a Vulkan/GLFW match. Runs at the end of the top-level `CMakeLists.txt`, after every `add_subdirectory()`.
- **`include/mdux/core/.clang-tidy`** — inherits root config, adds the governed-zone profile from ADR-004 (re-enables `bugprone-exception-escape`, adds the `cppcoreguidelines-pro-*`/`no-malloc`/`owning-memory` set as errors).
- `MduX` now `PUBLIC`-links `MduX::Core` — existing `MduX::MduX` consumers get it transitively, unaffected.
- `tests/core/CoreTests.cpp` — 7 new cases, deliberately linking `MduX::Core` only (not `MduX::MduX`), which is itself a check that the governed modules are usable standalone.

## Verified on real hardware (same Ubuntu 26.04 box as #74/#75/#48)

- Full clean build + `ctest`: **34/34 passed** (27 previous + 7 new), on both GCC 15.2.0 and GCC 16.
- `mdux_verify_trust_zones()` prints `1 governed target(s) clean of Vulkan/windowing dependencies` on a normal configure.
- **Proved the check actually catches a violation, not just passes trivially**: added `target_link_libraries(MduXCore PRIVATE Vulkan::Vulkan)` by hand and reconfigured — got a clean `FATAL_ERROR` with the exact call stack (`MduXCore` → `Vulkan::Vulkan`), then reverted and confirmed a clean diff.

## Two more real CMake/GCC bugs found and fixed along the way

1. **A function parameter literally named `TARGET` collides with CMake's own `if(TARGET <name>)` keyword form.** `if(TARGET IN_LIST visited)` failed with `Unknown arguments specified` — CMake read the bare word `TARGET` as the start of its target-existence-check syntax instead of a variable reference, tried to check for a target literally named `IN_LIST`, and choked on the leftover `visited` token. Renamed the parameter to `TGT` throughout and documented why in a comment, so nobody reintroduces it.
2. **A third distinct GCC C++23-modules ICE**, this time a segfault: a `friend`-defaulted `operator==` declared inside a struct exported from a module crashes GCC 15/16 when the module is consumed (`internal compiler error: Segmentation fault` inside `Units.cppm:29`, pointing at the defaulted `operator==`). Switched all three affected types (`Extent2D`, `Rect`, `ColorRgba8`) from `friend constexpr bool operator==(...) = default` to a member `constexpr bool operator==(...) const = default` — same semantics, no crash. This is the third GCC-modules ICE this parity programme has hit (see #48 for the first two); worth keeping in mind for anything else that exports a struct with a defaulted comparison operator.


camille-martin-paris:
Review completed and enforcement fix pushed in 0ac4223.

The original verifier caught Vulkan/GLFW names but did not implement #46’s stronger rule that every target reachable from a governed target must itself be governed. That allowed an arbitrary adapter target to enter the Core graph undetected. The follow-up now:
- rejects every non-governed CMake target in the transitive link graph
- resolves aliases before governed-set membership checks
- explicitly governs the std-only/compiler-warning `MduX_warnings` interface target
- fails closed on generator-expression forms the configure-time verifier cannot prove clean

`git diff --check` passes. Local configure/build remains unavailable because CMake is absent and the installed GCC 13.3 is below the project’s GCC 15 floor; pushed CI will exercise configuration on Windows, GCC 15, and GCC 16.

Regulated-change assessment: this strengthens the governed/adapter architectural control and does not alter runtime safety behavior or make a new standards claim, so no risk-file or regulatory-citation update is required.

PR #77, opened 2026-07-27
Restore install/export with an end-to-end install-tree consumer test

### LINKED ISSUE #47: S8 — Restore install/export
`CMakeLists.txt:229-240` is entirely commented out, marked *"temporarily disabled for import std testing"*. There is no `cmake/MduXConfig.cmake.in`. **MduX cannot be installed or consumed via `find_package`.**

**This is not cleanup.** E9's `.medui` build integration requires that a consuming project can do:

```cmake
find_package(MduX REQUIRED)
mdux_add_medui_screen(my_app FILE screens/Vitals.medui SURFACE 1920 1080)
```

which means `MduX::meduic` must be exported as an `IMPORTED EXECUTABLE` alongside the library. Without install/export, `.medui` only works inside this repository — which defeats the point of shipping an SDK.

**Work:**

1. Un-comment and fix the install block; install `FILE_SET CXX_MODULES` correctly (the commented code already gestures at this).
2. Write the missing `cmake/MduXConfig.cmake.in`, with `find_dependency(Vulkan)` and `find_dependency(Threads)`.
3. Export `MduXCore`, `MduX`, `MduX_options`, `MduX_warnings`, and (as they land) the host tools as `IMPORTED EXECUTABLE`s.
4. **Add a ctest that configures and builds a tiny consumer project against the install tree.** An install rule nobody exercises is an install rule that is broken; this is the only way to know.

Expect friction from `CMAKE_EXPERIMENTAL_CXX_IMPORT_STD` — installed module interfaces plus `import std` in a consumer is exactly the young code path this project lives on. If it cannot be made to work, document precisely what fails, because E9/S9's cross-compilation design depends on knowing.

Part of #11.

ambroise-leclerc:
Closes #47.

Stacked on #76 (branched after MduXCore so MduXCore installs and exports alongside MduX from the start — merge order: #74, #75, #76, this).

## What was broken

`CMakeLists.txt:229-240` was entirely commented out, marked *"temporarily disabled for import std testing"*. `cmake/MduXConfig.cmake.in` already existed with real content (I didn't know this going in — found it while starting the work) but was unused, and was missing `find_dependency(Vulkan)` despite the library requiring it.

## What's fixed

- Un-commented and fixed the install block: `MduXCore` installs and exports alongside `MduX`, both via their own `FILE_SET CXX_MODULES`.
- Added `find_dependency(Vulkan REQUIRED)` to `cmake/MduXConfig.cmake.in` (it only had `Threads`).
- **`cmake/TestInstallConsumer.cmake`** + an `InstallTreeConsumer` ctest — installs to a scratch prefix, then configures/builds/**runs** a tiny external project (outside this repo's own CMake graph) against it via `find_package(MduX)`. The consumer program asserts real values (`mdux::Version::getString()`, `mdux::Compliance::isMedicalDeviceCompliant`) read through the installed module, so a pass means the installed `import std` + `FILE_SET CXX_MODULES` combination is genuinely usable from outside the source tree, not just present on disk. Old-style positional `add_test()`, consistent with the fix in #75 (the keyword form mis-parses on this project's CTest version).

## A real bug found by actually inspecting the installed tree

The original (pre-existing, already-authored) install rules combined a blanket `install(DIRECTORY include/ ... FILES_MATCHING PATTERN "*.cppm")` with the `FILE_SET CXX_MODULES` install — redundant, and produced every module file installed **twice**, once correctly at `include/mdux/mdux.cppm` and once incorrectly nested at `include/mdux/include/mdux/mdux.cppm`. Harmless functionally (the consumer test still found a valid copy and passed) but untidy. Fixed by adding explicit `BASE_DIRS` to both `FILE_SET CXX_MODULES` declarations and deleting the now-redundant `DIRECTORY` install — verified by re-inspecting the installed tree, which is now exactly 5 `.cppm` files at their correct paths.

## Verified on real hardware (same Ubuntu 26.04 box as #74/#75/#76/#48)

- Full clean build + `ctest`: **35/35 passed** (34 previous + the new `InstallTreeConsumer`), on both GCC 15.2.0 and GCC 16.
- Manually inspected the installed tree twice — once showing the duplication bug, once after the fix showing exactly the expected 5 module files plus `MduXConfig.cmake`/`MduXConfigVersion.cmake`/`MduXTargets.cmake`/`MduXTargets-release.cmake`/`libMduX.a`/`libMduXCore.a`.


camille-martin-paris:
Review completed and packaging fixes pushed in 0b0dcd9.

Findings addressed:
- `MduXCore` would have exported as `MduX::MduXCore`, inconsistent with the build-tree API `MduX::Core`; it now uses `EXPORT_NAME Core` and the consumer test asserts that target exists
- `MduX_warnings` was PUBLIC, which would impose MduX’s warning set and `-Werror`/`/WX` on downstream source files; warning policy is now PRIVATE to library compilation
- the install consumer hard-coded version `0.1.0`; it now receives and validates `${PROJECT_VERSION}`

`git diff --check` passes. Local install-tree execution remains unavailable because this workspace has no CMake and GCC 13.3 is below the required GCC 15 floor; pushed CI will rerun the external consumer on supported toolchains.

PR #78, opened 2026-07-27
Fix Windows CI: replace hardcoded vcvars64.bat path with ilammy/msvc-dev-cmd

ambroise-leclerc:
Not stacked on the parity-programme PRs — this fixes a repo-wide CI break affecting every open PR identically (#6, #67, #68, and by extension #69-#77 once they rebase), so it should merge independently and as soon as possible.

## Root cause

`windows-build` called
`"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"`
directly, three times. GitHub's `windows-latest` runner image has since moved to a newer Visual Studio release at a different install path — a recent failing run shows `cl.exe` resolving under `Visual Studio\18\Enterprise\...`, not `2022`. The hardcoded `call` silently failed to find the batch file, and the resulting error was actively misleading: cmd's `call` on a missing file left a garbled command line that then printed `'#' is not recognized as an internal or external command` — masking the real cause behind what looks like an unrelated syntax error.

## Fix

Replaced the three inline `vcvars64.bat` calls with a single `ilammy/msvc-dev-cmd@v1` step. It locates whatever MSVC version is actually present via `vswhere` and exports the dev environment into `GITHUB_ENV` once, for every later step in the job — no hardcoded version or edition, so this survives future runner image upgrades without maintenance.

## Verification

- Workflow YAML parses.
- Watching this PR's own CI run below to confirm Windows actually goes green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,11 +26,20 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v5
       
-    - name: Setup CMake (latest)
+    # Pinned, not 'latest': CMakeLists.txt hardcodes CMAKE_EXPERIMENTAL_CXX_IMPORT_STD
+    # to a specific UUID that only matches CMake's `import std` experimental-feature
+    # ABI as of 4.1.1 - CMake changes that UUID as the feature evolves release to
+    # release, so newer CMake versions reject it ("CMAKE_EXPERIMENTAL_CXX_IMPORT_STD
+    # is set to incorrect value") and import std silently stops working, cascading
+    # into hundreds of "could not find module 'std'" errors at compile time. Every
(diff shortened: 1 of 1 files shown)


camille-martin-paris:
Review follow-up pushed in `e9f9f4f`.

The Windows CI repair is functionally sound and all four jobs were green. I hardened the newly introduced third-party workflow dependency by replacing mutable `ilammy/msvc-dev-cmd@v1` with the exact commit for v1.13.0 (`0b201ec74fa43914dc39ae48a89fd1d8cb592756`), retaining the version comment for readability. This prevents the executed action code from changing without a repository diff.

Validation: GitHub API verification that the SHA resolves to v1.13.0, workflow YAML parsing, and `git diff --check`.

Issue #55, opened 2026-07-26
S7 — CI: ctest -L evidence on both toolchain legs


ambroise-leclerc:
Wire the verification into `.github/workflows/ci.yml` so a drifted artifact fails the PR that caused it.

**In every leg:**

```bash
ctest --test-dir build -L evidence --output-on-failure
git status --porcelain    # must be empty
```

The second line catches a build that wrote into the source tree — forbidden by E6/S5's design — and an artifact that should have been committed but was swallowed by `.gitignore` (E5/S4).

**Why both legs matter, and why this must not be "optimised" to one:**

Running the evidence tests on **Windows/MSVC and Linux/GCC in the same PR** proves the bakers produce byte-identical output across two independent toolchains, two standard libraries, and two floating-point code generators. TrustSC's equivalent check runs against a single rustc; this is strictly stronger, and it is free.

It is also load-bearing for E12: `ml.determinism.crossToolchain` is the same check applied to a baked `ModelPackage`, and it is the evidence behind the claim that golden vectors computed on the host reproduce bit-for-bit on the device.

Once E5/S9 re-enables the Clang leg, add it here too — three toolchains is better than two, and Clang's floating-point code generation differs from both.

Failure output should name the first differing byte offset, not just report a mismatch.

Part of #12.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

Issue #54, opened 2026-07-26
S6 — tools/common/ (TOML subset and CLI)


ambroise-leclerc:
Shared host-only infrastructure for the six bakers. **Host tools zone** — never linked into `MduXCore` or `MduX`, so it may use exceptions freely (E5/S2).

**`tools/common/Toml.cppm`** — a deliberately minimal TOML reader for recipe files: tables, strings, integers, booleans, and arrays of those. **No** dotted keys, datetimes, inline tables, or multi-line strings.

The narrow scope is the point: recipe files are authored by this project, TOML's full grammar is large, and every line here is code a manufacturer may have to qualify. If a recipe needs a feature outside this subset, the recipe format is wrong, not the parser.

**`tools/common/Cli.cppm`** — argument parsing shared by every tool, so they present one interface:

```
mdux-<kind>bake bake   <recipe> <output-dir>
mdux-<kind>bake verify <recipe> <package.json> <report.json>
                       [--format=json|text]
```

`bake` and `verify` running the *same* code path with different output handling is what makes the CI check meaningful — verify must not be a reimplementation that can drift.

`--format=json` here is the first instance of E13/S3's stable diagnostic envelope; define it once and have every tool inherit it.

Part of #12.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

Issue #53, opened 2026-07-26
S5 — cmake/MduXBake.cmake


ambroise-leclerc:
The CMake surface every baker registers through, so adding a new asset kind is one function call rather than a new pile of custom commands.

```cmake
mdux_bake_artifact(
    KIND    font
    ID      roboto-ui
    TOOL    MduX::fontbake
    RECIPE  recipes/font/roboto-ui.toml
    SOURCES assets/fonts/Roboto-Regular.ttf
    OUTPUTS package.json report.json atlas.bin)
```

Each call registers three things:

1. A `bake-font-roboto-ui` custom target writing to `${CMAKE_BINARY_DIR}/mdux_bake/font/roboto-ui/`.
2. A ctest `evidence.font.roboto-ui` with `LABELS evidence`, running `cmake -E compare_files` for every output against `generated/font/roboto-ui/`.
3. Membership in the aggregate targets `mdux-bake-all` and `mdux-bake-update`.

**The source-tree rule, enforced by construction:** a normal build **never** writes into the source tree. It bakes into the build directory and compares. `cmake --build build --target mdux-bake-update` is the only path that copies build-dir artifacts over `generated/`. An author runs it deliberately and commits the diff; a reviewer reads that diff.

This mirrors Cargo's `OUT_DIR` discipline, which MduX otherwise loses by not having `build.rs`.

**Host-tool resolution** goes through `MduX::<tool>` uniformly (see E9/S9's `cmake/MduXHostTools.cmake`), so a cross-compiling build substitutes an imported executable without any call site changing.

Part of #12.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

Issue #52, opened 2026-07-26
S4 — mdux.evidence.report


ambroise-leclerc:
`include/mdux/evidence/Report.cppm` + `src/evidence/Report.cpp`. **Governed zone.** The shared shape of every `report.json`, so all six bakers produce audit records an auditor can read the same way.

```json
{
  "schemaVersion": 1,
  "tool": "mdux-fontbake",
  "toolVersion": "0.2.0",
  "toolGitSha": "a1b2c3d…",
  "recipe":  { "path": "recipes/font/roboto-ui.toml", "sha256": "…" },
  "inputs":  [ { "path": "assets/fonts/Roboto-Regular.ttf", "sha256": "…" } ],
  "options": { "…fully resolved, defaults expanded…" },
  "outputs": [ { "path": "package.json", "sha256": "…" },
               { "path": "atlas.bin",    "sha256": "…" } ]
}
```

**`options` must be the fully resolved set with defaults expanded**, not the recipe's literal contents. Otherwise changing a default silently changes every output while every report still looks unchanged — which is exactly the failure a byte-verified pipeline exists to prevent.

**`toolGitSha`** comes from a configure-time `git rev-parse` baked in as a compile definition. When git is unavailable it is `"unknown"` — and CI must **fail** on `"unknown"` appearing in a *committed* report, since that means someone baked outside a working tree.

Types: `PackageHeader` (schemaVersion, id, kind) shared by every `package.json`, and `BakeReport` with the fields above. Both serialize through E6/S3's canonical writer.

Part of #12.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

PR #79, opened 2026-07-27
S1 — ADR-007: evidence pipeline doctrine

ambroise-leclerc:
## Summary
- Adds ADR-007, recording the evidence-pipeline doctrine every baker in the upcoming kernel (issues #50-#55) follows: `recipe+source -> host baker -> committed generated/<kind>/<id>/{package.json,report.json,payload.bin}`, verified in CI by re-baking and asserting byte-identity.
- Documents two deliberate deviations from TrustSC's equivalent decision: sidecar `.bin` payloads instead of base64-in-JSON (keeps git history usable at font-atlas/ML-weight sizes), and floats encoded as `u32` bit patterns rather than decimal text (`printf`/`std::format` are not byte-identical across MSVC/glibc/libc++, and this pipeline crosses all three in CI).
- Records the two-toolchain byte-identity claim (stronger than TrustSC's single-rustc guarantee) and the source-tree rule (`mdux-bake-update` is the only path that writes into `generated/`).

## Context
First of a stack of PRs implementing Epic #12 (Evidence kernel), the sole gate on Track C (#13) and Track D (#18) per the Wave 2 roadmap. This PR is documentation only — no code.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings (citation format, justification schema, reproduced-text heuristics all pass)
- [x] `docs/adr/README.md` index updated

Stacked on top of this PR: #50 (digest), #51 (canonical JSON), #52 (bake report), #53 (CMake bake infra), #54 (host-tools TOML/CLI), #55 (CI wiring).

### DIFF
--- a/docs/adr/ADR-007-evidence-pipeline-doctrine.md
+++ b/docs/adr/ADR-007-evidence-pipeline-doctrine.md
@@ -1,0 +1,235 @@
+# ADR-007: Evidence pipeline doctrine
+
+## Status
+Accepted (2026-07-27)
+
+## Context
+MduX currently has no evidence pipeline at all. The committed SPIR-V under
+`examples/shaders/compiled/` has no consumer, no recorded provenance, and no check that it
+corresponds to the GLSL beside it — and `.gitignore` ignored `*.spv` until issue #43 fixed the
+negations. Nothing in the repository can answer "which tool, at which version, from which input,
+produced this binary artifact".
+
+Every asset kind still to be built needs that answer: shaders (issue #13), fonts and text packages
+(issue #14), compiled `.medui` screens (issue #15), images and traces (issue #17), and ML model
+packages (issue #18). Six bakers, six chances to invent a different provenance format. Deciding the
+shape once — before the first baker exists — is cheaper than reconciling six of them later, and it
+is the difference between artifacts an auditor can read uniformly and six bespoke conventions.
+
+The constraint that shapes the whole design is that authoring dependencies must not reach the device
+build. A font parser, a safetensors reader and a SPIR-V walker are exactly the kind of code a
+manufacturer would otherwise have to qualify as SOUP. If baking happens at build time on the host
+and the *result* is committed, the runtime build never links any of it.
+
+## Medical Device Considerations
+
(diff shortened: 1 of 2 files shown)


camille-martin-paris:
Review finding (blocking): Decision 5 is incompatible with Decision 3's byte-identity check as written.

If a baker embeds configure-time `git rev-parse HEAD` in `report.json`, the artifact is baked at H0 and then committed as H1. CI checks out H1 (or GitHub's synthetic merge commit), re-bakes with a different SHA, and must report a byte mismatch. There is no commit whose own SHA can be embedded in a file in that same commit.

Please define provenance using a stable input to the bake (for example a digest/identity of the baker sources and inputs, or another explicitly supplied immutable revision) before accepting the doctrine. This is potentially safety-relevant because it affects whether the claimed traceability and verification control can operate at all.

ambroise-leclerc:
Confirmed — this is correct and blocking, thank you. Baking at H0, committing the report, then re-baking at H1 (or the merge SHA) can never reproduce H0's embedded value; there's no fixed point, since a commit's tree can't correctly name its own hash while that hash is still being computed from it.

Fixed by removing `toolGitSha` from `BakeReport` entirely rather than finding a workaround — the field was redundant with `git log --follow` on the artifact path anyway. `toolVersion` (deliberately, manually bumped) remains as the report's only tooling-identity field, and it has no self-reference problem since a version bump is a separate human action, not derived from "current HEAD."

Rewrote decision 5 in this ADR (05594ae) to explain the structural impossibility and record the fix, and added a rejected-alternatives entry so this isn't re-proposed unread. The same fix is threaded through the whole stack: `cmake/MduXBuildInfo.cmake` (#82) no longer defines `MDUX_GIT_SHA`, `BakeReport`/`Report.cppm` (#82) has no `toolGitSha` field or `requireKnownGitSha()`, and the now-obsolete CI check on #85 is removed. Added a regression test in #82 (`ReportTests.cpp`) asserting two reports built from identical inputs are byte-identical regardless of "when" each was built — the property a reintroduced commit-SHA field would break.

Issue #50, opened 2026-07-26
S2 — mdux.evidence.digest (SHA-256)


ambroise-leclerc:
`include/mdux/evidence/Digest.cppm` + `src/evidence/Digest.cpp`. **Governed zone** — `std` only, `noexcept`, no allocation.

A from-scratch SHA-256. Every `report.json` digest, every `package.json` sidecar reference, and `Classifier1D::create()`'s weight-blob verification (E12/S7) depend on it.

Hand-writing it rather than taking a dependency is the zero-SOUP position, and SHA-256 is one of the few algorithms where that is uncontroversial: ~200 lines, a fixed specification, and an official test-vector suite.

**API:**

```cpp
using Digest = std::array<std::uint8_t, 32>;
Digest sha256(std::span<const std::byte>) noexcept;

class Sha256 {                       // streaming, for large files
public:
    void update(std::span<const std::byte>) noexcept;
    [[nodiscard]] Digest finish() noexcept;
};
std::array<char, 64> toHex(const Digest&) noexcept;
```

**Tests:** the NIST FIPS 180-4 vectors, the empty input, inputs straddling the 64-byte block boundary, and a multi-chunk streaming case that must equal the one-shot result. Add a case over a multi-megabyte buffer so the streaming path is exercised at the size E12's weight blobs will hit.

Part of #12.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

PR #80, opened 2026-07-27
S2 — mdux.evidence.digest (SHA-256)

ambroise-leclerc:
## Summary
- Adds `mdux.evidence.digest`: governed-zone (ADR-004), `noexcept`, allocation-free SHA-256 — `Digest = std::array<uint8_t,32>`, `sha256(span<const byte>)`, and a streaming `Sha256` class for inputs too large for one span.
- Hand-written rather than a dependency (the zero-SOUP position ADR-007 requires of anything the device build could link), which is uncontroversial for SHA-256 specifically: a fixed FIPS 180-4 specification, ~200 lines, an official test-vector suite.
- Wires `Digest.cppm`/`Digest.cpp` into `MduXCore`'s `FILE_SET`, and adds a new `evidence_tests` executable (own `MDUX_TEST_MAIN`, links `MduX::Core` only — same rationale as `core_tests`: a governed module must be usable without pulling in Vulkan).

## Test plan
12/12 pass, checked against **independently-known** digests (published FIPS 180-4 vectors, and Python's `hashlib.sha256` for the rest) — never against this implementation's own output:
- [x] Empty string, `"abc"`, the 448-bit and 896-bit standard vectors, one million `'a'`
- [x] Both sides of the 64-byte block boundary at nine lengths (55/56/57, 63/64/65, 119/120, 127/128)
- [x] Streaming in awkward chunk sizes (1, 7, 63, 64, 65, 128, 299 bytes) against the one-shot digest
- [x] A 4 MiB buffer — the order of magnitude issue #18's ML weight blobs will reach
- [x] `finish()` non-destructive / repeatable, `reset()`, `toHex()` lowercase/64-char

Verified by compiling directly with Clang 21 + libc++'s `std` module (this repo's CMake `import std` UUID is pinned to CMake 4.0.3 and this sandbox has 4.3.1 — pre-existing, unrelated to this change, confirmed it also blocks building master's own `Result.cppm`). Real CI is the first place the actual CMake/Vulkan build graph runs this.

Stacked on #79 (ADR-007). Part of the #12 Evidence kernel stack (issue #50).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,9 @@ target_sources(MduXCore
         FILES
             include/mdux/core/Units.cppm
             include/mdux/core/Result.cppm
+            include/mdux/evidence/Digest.cppm
+    PRIVATE
+        src/evidence/Digest.cpp
 )
 target_compile_features(MduXCore PUBLIC cxx_std_23)
 target_include_directories(MduXCore PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
--- a/include/mdux/evidence/Digest.cppm
+++ b/include/mdux/evidence/Digest.cppm
@@ -1,0 +1,71 @@
+/**
+ * @file Digest.cppm
+ * @brief Governed-zone SHA-256: the digest every evidence artifact is identified by.
+ *
+ * @compliance ADR-004 Trust zones in C++ (governed zone: std only, no Vulkan, no windowing)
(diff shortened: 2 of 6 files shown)


camille-martin-paris:
Reviewed the isolated digest delta and its cumulative use in the stack. I found no additional actionable defect in the SHA-256 API/implementation or test coverage. The independent vectors, boundary cases, streaming equivalence, repeatable `finish()`, and reset coverage are proportionate.

Local C++ build execution was unavailable here because the environment has no CMake and only GCC 13.3 (the repository requires CMake 4+ and GCC 15+).

ambroise-leclerc:

Issue #51, opened 2026-07-26
S3 — mdux.evidence.json (canonical writer and strict reader)


ambroise-leclerc:
`include/mdux/evidence/Json.cppm` + `src/evidence/Json.cpp`. **Governed zone.** The component that makes byte-identity possible — or, if it gets one rule wrong, impossible.

**Canonical form:**

- keys sorted, lexicographic by UTF-8 code unit
- 2-space indent, LF line endings, UTF-8 without BOM, trailing newline
- **no timestamps, no absolute paths, no environment-dependent values** anywhere
- **floats encoded as `u32` bit patterns** — `{"bits": 1065353216}` — never decimal text

That last rule is the whole ballgame. `printf("%.9g")` and `std::format("{}", f)` are not guaranteed byte-identical across MSVC, glibc and libc++. Baked font metrics, layout bounds and ML golden vectors are all floats, and all three toolchains build this project.

**Add the lint that enforces it:** a CI grep over `src/evidence/` and `tools/` banning `%f`, `%g`, `%e` and float `std::format` specifiers. One stray specifier in a baker silently breaks byte-identity in a way that only reproduces on someone else's OS.

**Reader:** strict — reject duplicate keys, trailing commas, comments, NaN/Infinity literals, and any number that is not exactly representable in the declared type. A permissive reader would let a hand-edited `generated/` file pass verification.

**Tests:** round-trip every package type; assert key ordering is stable regardless of insertion order; assert that every `f32` bit pattern including denormals, ±0 and NaN survives a round trip bit-exactly.

Part of #12.

ambroise-leclerc:
Implemented on develop and verified by the v0.3.0 release. The final release commit f7756aa1a7637bad48eef5e1372150a38f2694d0 passed all seven jobs in CI run 30717524381. Closing during the Wave 3 roadmap reconciliation; any remaining follow-up is now represented by the updated active epics.

PR #81, opened 2026-07-27
S3 — mdux.evidence.json (canonical writer + strict reader) and mdux-evidence-lint

ambroise-leclerc:
## Summary
- Adds `mdux.evidence.json`: canonical JSON writer + strict reader, the component byte-identity across MSVC/glibc/libc++ depends on. Canonical form — keys sorted by UTF-8 code unit, 2-space indent, LF, no BOM, trailing newline, floats as `{"bits": N}` never decimal text.
- The reader is deliberately stricter than general JSON: rejects duplicate keys, trailing commas, comments, `NaN`/`Infinity`, BOM, invalid UTF-8, trailing content, excessive nesting — and **any number with a fraction or exponent**, since canonical form never emits one.
- Adds `tools/evidence-lint/` banning printf/`std::format` float conversions inside string literals under `src/evidence/` and `tools/` — the CI-enforced half of the same rule.

## Test plan
- [x] 23 new JSON module tests (35/35 cumulative in `evidence_tests`) — canonical form asserted against **literal expected byte strings**, not re-serialization; every notable `f32` bit pattern (±0, denormals, NaN, ∞) round-trips bit-exactly; 20+ rejection cases each pinned to the specific expected error code
- [x] 22 lint self-tests, including two regression cases from my own false positives caught during development: `Mode mode{Mode::Bake}` and `int r = a %factor` both looked like format specifiers to an earlier whole-line version of this lint — fixed by scanning string literals only, which is the only fix that doesn't require impossible rewording of ordinary C++
- [x] `python3 tools/evidence-lint/mdux_evidence_lint.py` — 0 findings

Same sandbox caveat as #80 (no local `import std` due to the CMake UUID pin) — verified with Clang 21 + libc++'s `std` module directly.

Stacked on #80 (digest). Part of the #12 Evidence kernel stack (issue #51).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,8 @@ CLAUDE.md
 .roo/
 GEMINI.md
 .github/copilot-instructions.md
+
+# Python bytecode from the host-only lint tools (tools/docs-lint, tools/evidence-lint).
+# CI sets PYTHONDONTWRITEBYTECODE=1, but a local run of the self-tests still leaves these behind.
+__pycache__/
+*.pyc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,8 +117,10 @@ target_sources(MduXCore
             include/mdux/core/Units.cppm
             include/mdux/core/Result.cppm
             include/mdux/evidence/Digest.cppm
+            include/mdux/evidence/Json.cppm
     PRIVATE
         src/evidence/Digest.cpp
+        src/evidence/Json.cpp
 )
 target_compile_features(MduXCore PUBLIC cxx_std_23)
 target_include_directories(MduXCore PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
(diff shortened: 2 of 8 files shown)


camille-martin-paris:
Reviewed the canonical JSON implementation, strict-reader tests, and evidence-lint tool. The Python evidence-lint suite passes (22 tests), and the lint reports 0 findings across 10 files. I found no additional actionable defect in this PR.

I could not execute the C++ tests locally because CMake is absent and the available GCC is 13.3, below the required GCC 15 floor.

ambroise-leclerc:

ambroise-leclerc:
Retargeted to `develop` — the whole stack was mistakenly opened against `master` instead of `develop` (this repo's actual default branch). #79 and #80 had already merged into `master` directly before this was caught; since `develop` was a strict ancestor of `master` with no divergent commits, I fast-forwarded `develop` to `master`'s tip (`9aa30f9..16a66f3`) rather than a real merge, so both branches are identical again and nothing from #79/#80 is missing from `develop` going forward.

`master` now has branch protection requiring all changes go through a pull request (enforced for admins too — verified with a direct-push test, rejected with `GH006: Changes must be made through a pull request`). From here, `master` should only move via a deliberate release PR from `develop`; this stack (and anything else in flight) targets `develop`.

Diff against the new base is clean and unchanged in content — see the stat above.

Issue #13, opened 2026-07-26
Epic: Shader pipeline and renderer vertical slice


ambroise-leclerc:
Draw the first library pixel through a fixed-budget Vulkan renderer using shaders produced by the completed evidence kernel (#12). This is the next active vertical slice after Wave 3 reconciliation.

## Children and merge order

- [x] #119 — Shader schema and baker
- [x] #120 — Authored and committed UI shaders (after S1)
- [x] #121 — Module/header emitter using C arrays (after S2)
- [x] #123 — Fixed-budget governed draw types (may start after S1)
- [x] #124 — Vulkan renderer (after S3 and S5)
- [x] #125 — Headless/offscreen readback (after S6)
- [x] #126 — Lavapipe pixel test (after S7)
- [x] #122 — Triangle example migration (after S3 and #48)
- [x] #127 — Retire HTML/CSS path (last, after replacement and pixel test)

## Integration constraints from Wave 2

- The schema is canonical; bakers, emitters, and runtime import it rather than restating types.
- Dependent PRs do not merge as an unreconciled stack. After each predecessor merges, rebase the successor on current develop, re-review shared CMake/module wiring, and wait for green post-merge CI.
- Generated code uses C arrays with span views, not std::array.
- All tools use the shared diagnostic envelope and the implemented BakeReport with no commit SHA.

**Blocked by:** the individual dependencies above. #12 is complete.
**Blocks:** #14 and #16.

## Done when

The library records a fixed-budget draw, a headless pixel test passes on lavapipe, the example consumes embedded shaders, and the legacy HTML/CSS runtime path is removed without breaking install/export.

ambroise-leclerc:
All nine children (#119–#127) are merged and closed.

Delivered: the shader schema and baker, committed and byte-verified UI shaders, the module/header emitter, the governed fixed-budget draw types, the Vulkan renderer, offscreen readback, the project's first pixel test under lavapipe in CI, the triangle example migrated onto the baked package, and the HTML/CSS path retired.

Verified on `develop` from a clean build: 360/360 tests pass under GCC 16.1, including `ctest -L pixel`.

With this closed, **#14 (font and text) is unblocked** — its two blockers were #12 and #13, both now done.

PR #82, opened 2026-07-27
S4 — mdux.evidence.report and cmake/MduXBuildInfo.cmake

ambroise-leclerc:
## Summary
- Adds `mdux.evidence.report`: `PackageHeader` and `BakeReport` — the shared audit-record shape every baker (issues #13-#18) will emit, serialized through #81's canonical JSON writer.
- `validate()` enforces two invariants that would otherwise silently break byte-identity or auditability: paths are repository-relative with `/` separators (rejects absolute paths and backslashes — either embeds the producing machine/OS in a committed artifact), and a report must list at least one output (empty outputs would make verification trivially pass).
- `toolGitSha` gets two checks with different strictness: `validate()` permits `"unknown"` (a source-tarball build with no git metadata must still work locally), `requireKnownGitSha()` — the check CI applies to a *committed* report — does not, since `"unknown"` there means the artifact's provenance is unverifiable.
- Adds `cmake/MduXBuildInfo.cmake`: `mdux_define_build_info()` creates `MduX::BuildInfo` carrying `MDUX_TOOL_VERSION`/`MDUX_GIT_SHA` (configure-time `git rev-parse HEAD`, `CMAKE_CONFIGURE_DEPENDS` on `.git/HEAD` so a later commit invalidates the cached SHA).

## Test plan
- [x] 59/59 evidence-module tests (cumulative) — canonical serialization asserted literally; round-trip through write/parse reproduces identical bytes; every `validate()` rejection path (empty tool fields, absolute/backslash/`..` paths in recipe *and* inputs *and* outputs, no-outputs, duplicate-output-path, unsupported schema version) pinned to its specific error code; `validate()` vs `requireKnownGitSha()` divergence on `"unknown"` explicitly tested
- [x] `cmake/MduXBuildInfo.cmake` configured against **this actual repository** (not a synthetic stub) — picked up the real `git rev-parse HEAD` for this branch's own commit; separately verified the git-detection and `.git/HEAD` ref-following logic in isolation with `cmake -P`

Stacked on #81 (JSON). Part of the #12 Evidence kernel stack (issue #52).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,11 @@ include(StaticAnalyzers)
 include(Doxygen)
 include(DependencySummaryTable)
 include(MduXTrustZones)
+include(MduXBuildInfo)
+
+# Artifact provenance for host tools (ADR-007): defines MduX::BuildInfo carrying
+# MDUX_TOOL_VERSION and MDUX_GIT_SHA. Runs before any target is created so bakers can link it.
+mdux_define_build_info()
 
 # Build options
 option(MDUX_BUILD_EXAMPLES "Build MduX examples" ON)
@@ -118,9 +123,11 @@ target_sources(MduXCore
             include/mdux/core/Result.cppm
             include/mdux/evidence/Digest.cppm
             include/mdux/evidence/Json.cppm
+            include/mdux/evidence/Report.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
+        src/evidence/Report.cpp
 )
 target_compile_features(MduXCore PUBLIC cxx_std_23)
 target_include_directories(MduXCore PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
--- a/cmake/MduXBuildInfo.cmake
+++ b/cmake/MduXBuildInfo.cmake
@@ -1,0 +1,45 @@
+# MduXBuildInfo.cmake
+#
+# Provenance a baked artifact's report.json is allowed to carry: which tool *version* produced it
+# (ADR-007, decision 5).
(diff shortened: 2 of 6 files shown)


camille-martin-paris:
Review finding (blocking): `MDUX_GIT_SHA` is derived from configure-time `git rev-parse HEAD` and serialized into the byte-compared `report.json`.

That makes committed output impossible to reproduce: bake at H0, commit the generated report as H1, then re-bake at H1 (or the Actions merge SHA) and the report differs solely in `toolGitSha`. This is a self-reference problem, not a flaky-toolchain edge case.

Please change the provenance model to an identity stable across committing and CI re-baking, and add an end-to-end test that commits/changes HEAD before verification.

ambroise-leclerc:
Confirmed and fixed (1fb93e8) — `toolGitSha` is removed from `BakeReport` entirely, and `cmake/MduXBuildInfo.cmake` no longer defines `MDUX_GIT_SHA` at all. `toolVersion` (the project's manually-bumped semver) is the report's only remaining "which tooling produced this" identity, and it has no self-reference problem since a version bump is a deliberate separate action, not derived from HEAD at bake time.

Also removed: `kUnknownGitSha`, `requireKnownGitSha()`, and the `EmptyGitSha`/`UnknownGitSha` error codes — none of that machinery has a purpose once there's no commit-dependent field to gate.

Added a regression test (`ReportTests.cpp`): two `BakeReport`s built from identical inputs, standing in for "baked before commit" vs. "re-baked after commit," must produce byte-identical JSON. That's the property a reintroduced commit-SHA field would break, so it's what would catch this class of bug returning.

Full writeup of the structural reasoning is in ADR-007 decision 5 (#79, 05594ae) and as a rejected alternative there. 58/58 evidence-module tests pass (was 59; net one test removed with the dead machinery, one added for the regression guard).

PR #83, opened 2026-07-27
S5 — cmake/MduXBake.cmake

ambroise-leclerc:
## Summary
- Adds `cmake/MduXBake.cmake`: `mdux_bake_artifact(KIND ID TOOL RECIPE SOURCES OUTPUTS)` registers a `bake-<kind>-<id>` custom target (writes only under `${CMAKE_BINARY_DIR}/mdux_bake/`, never the source tree), an `evidence.<kind>.<id>` ctest (`LABELS evidence`) byte-comparing against `generated/<kind>/<id>/`, and membership in `mdux-bake-all`/`mdux-bake-update`.
- The source-tree rule from ADR-007 is enforced **by construction**: every custom-command output lives under the build directory, so a baker cannot write into `generated/` without someone editing this file. `cmake --build <dir> --target mdux-bake-update` is the only path that copies over it, via the new `MduXUpdateArtifacts.cmake`.
- `MduXCompareArtifacts.cmake` reports the **first differing byte offset** on mismatch, not just "files differ" — using windowed `file(READ ... OFFSET ... LIMIT ... HEX)` rather than reading the whole file into one hex string and walking it with `string(SUBSTRING)`. The naive version is quadratic: measured **8.3s** on a 4 MiB payload with the difference near the end; the windowed version measures **2.3s** on the same case.
- `mdux_declare_bake_targets()` is deliberately independent of `add_subdirectory(tools)` — this PR and the host-tools PR (#54) can land in either order.

## Test plan
End-to-end against these exact files, in an isolated CMake project with a stand-in baker (not the real fonts/shaders bakers, which don't exist yet — those land with issues #13-#18):
- [x] A normal build never writes into the source tree (`generated/` absent after `cmake --build`)
- [x] `ctest -L evidence` **fails** before `mdux-bake-update` runs, **passes** after
- [x] `mdux-bake-update` is idempotent (second run reports "already up to date")
- [x] Perturbing the baker's output is caught, with the byte offset verified correct at offset 0, mid-file, and 4194204 in a 4 MiB payload, plus the "one file is a prefix of the other" case
- [x] Changing the recipe re-triggers the bake (`DEPENDS` tracking works); a missing recipe is a **configure-time** `FATAL_ERROR`, not a build-time surprise
- [x] This branch's `CMakeLists.txt` state configured standalone against this repository (with a stub Vulkan package) — `mdux-bake-all`/`mdux-bake-update` exist and are buildable even with `tools/` not yet added (confirms the decoupling from #54 above)

Stacked on #82 (report + BuildInfo). Part of the #12 Evidence kernel stack (issue #53).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,10 +80,16 @@ include(Doxygen)
 include(DependencySummaryTable)
 include(MduXTrustZones)
 include(MduXBuildInfo)
+include(MduXBake)
 
 # Artifact provenance for host tools (ADR-007): defines MduX::BuildInfo carrying
 # MDUX_TOOL_VERSION and MDUX_GIT_SHA. Runs before any target is created so bakers can link it.
 mdux_define_build_info()
+
+# Bake aggregate targets, declared here rather than next to add_subdirectory(tools) below so
+# this has no dependency on tools/ existing. They exist even with no artifact registered yet, so
+# `--target mdux-bake-all` succeeds trivially rather than failing with "no such target".
+mdux_declare_bake_targets()
 
 # Build options
 option(MDUX_BUILD_EXAMPLES "Build MduX examples" ON)
--- a/cmake/MduXBake.cmake
+++ b/cmake/MduXBake.cmake
@@ -92,6 +92,14 @@ function(mdux_bake_artifact)
             "imported executable; declare it before registering this artifact.")
     endif()
 
+    foreach(component KIND ID)
+        if(NOT ARG_${component} MATCHES "^[a-z0-9][a-z0-9-]*$")
+            message(FATAL_ERROR
+                "mdux_bake_artifact: ${component} must be a lowercase slug containing only "
+                "letters, digits and '-' (got '${ARG_${component}}')")
+        endif()
+    endforeach()
+
     set(label "${ARG_KIND}/${ARG_ID}")
     set(target_name "bake-${ARG_KIND}-${ARG_ID}")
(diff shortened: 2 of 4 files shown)


camille-martin-paris:
I found that `KIND`, `ID`, and `OUTPUTS` were used in target/filesystem paths without confinement checks, allowing separators or `..` to escape the intended artifact directory and contradict the source-tree rule.

I pushed commit `a5d2cd7` to this PR's head branch: KIND/ID are now lowercase slugs, output entries must be direct file names, and duplicate outputs are rejected. Static diff checking passes for the fix; CMake execution was unavailable locally because this environment has no CMake.

The stack's separate `toolGitSha` reproducibility issue remains tracked on #79/#82/#85.

ambroise-leclerc:
Thank you for pushing the fix directly rather than just flagging it — the KIND/ID slug constraint and the OUTPUTS confinement (no `/`/`\`, no duplicates) are both correct and needed. Rebased this branch onto the corrected #82 (the `toolGitSha` fix) with your commit (a5d2cd7) preserved on top; re-verified the full end-to-end bake flow — including the confinement checks — against the rebased tree. Re-tested the specific attack shapes your fix closes (`KIND=../../etc`, an `OUTPUTS` entry containing `/`, a duplicate output path) and confirmed each is now a configure-time `FATAL_ERROR` rather than a path that could escape `generated/<kind>/<id>/`.

The `toolGitSha` issue you flagged as tracked separately on #79/#82/#85 is now fixed on all three — see my replies there.

Issue #19, opened 2026-07-26
Epic: Agent and LLM tooling parity


ambroise-leclerc:
Make repository tooling first-class for coding agents and LLMs through vendor-neutral instructions, open-format skills, stable diagnostics, and machine-readable contracts.

## Current state — assessed 6 September 2026

All six original children are closed and their implementation is present on `develop`. The old S4–S6 unchecked entries were stale: #263, #264 and #265 landed through #300, #301 and #302 respectively.

The compiler publishes `docs/medui/grammar.json` and supports `--explain`; all six baked kinds have resolved-option schemas under `docs/recipes/`; `mdux-meduic --dump-ir` emits resolved layout, colour and text-budget information; and `docs/tools/manifest.json` is generated from host-tool build registration. These features need no duplicate implementation issues.

The assessment found a guidance-maintenance gap, tracked in #304. PR #306 delivered the corrections and the regression check for empty documented test selections; #304 and this epic are now closed. All post-merge workflows passed at e68e04b9f41277775dd4a13b692790aa8c13ad86.

## Children

- [x] #65 — S1 Align AGENTS.md with the implemented architecture
- [x] #66 — S2 Repository skills for regulatory citations, evidence, medui and SDF documents
- [x] #118 — S3 Stable JSON diagnostic envelope across tools
- [x] #263 — S4 Machine-readable `.medui` grammar and `--explain`
- [x] #264 — S5 JSON Schemas for every recipe kind
- [x] #265 — S6 `--dump-ir` JSON and generated tool manifest
- [x] #304 — S7 Refresh agent instructions and reject empty documented test selections (assessment follow-up)

**Status:** S1–S7 delivered; this epic is closed. Phase 2 application/authoring work is tracked separately in #307–#311.
**Blocked by:** nothing; no original epic child remains open.

## Assessment evidence

Reviewed repository state at `b49a3480f86bd3dcdb67f58dc4bd464c448f134b` and live issue states.

- Tool-manifest freshness check passed (11 tools).
- Schema-drift check passed (six recipe schemas against eight committed reports, plus governance bindings).
- Manifest, schema-drift and diagnostic-envelope Python suites: 80 tests passed.
- Existing GCC build: all 149 `medui_tools_spec` CTest cases passed, including grammar/registry/IR coverage.
- The obsolete suite selectors in the build skill selected zero tests in that same populated build, substantiating #304.

[...]

PR #84, opened 2026-07-27
S6 — tools/common/ (TOML subset and CLI)

ambroise-leclerc:
## Summary
- Adds `tools/common/`: shared host-tools-zone (ADR-004) infrastructure every future baker (issues #13-#18) builds on.
- `mdux.tools.toml` — a deliberately minimal TOML reader: tables, bare keys, basic strings, integers, booleans, arrays. **No** dotted keys, inline tables, multi-line strings, floats, datetimes, digit separators. The narrow scope is the point — every line is code a manufacturer may have to qualify. Floats are rejected with a message pointing at ADR-007's bit-pattern rule.
- `mdux.tools.cli` — parses the `bake <recipe> <output-dir>` / `verify <recipe> <package.json> <report.json> [--format=json|text]` interface every baker presents, and the stable diagnostic envelope (`{file, line, code, severity, message, fixHint}`) — the first instance of the JSON output every MduX tool will share (issue #19).
- Adds `MduXToolsCommon` (links `MduX::Core`, since bakers build artifacts through the governed evidence modules) and wires `add_subdirectory(tools)` into the top-level build, ahead of `tests/`.

## Test plan
- [x] 32/32 host-tools tests, including a full battery of "outside this TOML subset" rejections (floats, dotted keys/tables, inline tables, literal strings, hex/octal/binary integers, digit separators, datetimes) each pinned to a message naming the *specific* unsupported feature — not just "syntax error"
- [x] The diagnostic envelope's exact JSON shape asserted **literally** — it's a contract future agent tooling (#19) keys off, not just human-readable text
- [x] Configured against this repository with a stub Vulkan package: `MduXToolsCommon`, `tools_tests`, `mdux-bake-all` all present and buildable together

Stacked on #83 (Bake CMake infra) — decoupled from it deliberately (see #83's description), so either could have landed first.
Part of the #12 Evidence kernel stack (issue #54).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,6 +274,12 @@ target_link_libraries(MduX PRIVATE MduX_warnings)
 target_link_libraries(MduXCore PRIVATE MduX_warnings)
 
 # Subdirectories
+#
+# tools/ comes before tests/ so the test targets can link MduX::ToolsCommon. It is not gated on
+# an option: the bakers under tools/ are what produce every committed artifact, so a build that
+# skipped them could not run `ctest -L evidence` at all (ADR-007).
+add_subdirectory(tools)
+
 if(MDUX_BUILD_EXAMPLES)
     add_subdirectory(examples)
 endif()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -68,6 +68,36 @@ endif()
 
 mdux_discover_tests(evidence_tests)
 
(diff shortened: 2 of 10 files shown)


camille-martin-paris:
Reviewed the host-tools boundary, CLI diagnostics, TOML subset parser, and tests. I found no additional actionable defect in this isolated PR. The host-only target separation is consistent with ADR-004/ADR-005.

C++ tests could not run locally because CMake is absent and GCC 13.3 is below the required floor.

ambroise-leclerc:

PR #85, opened 2026-07-27
S7 — CI: ctest -L evidence on all toolchain legs, evidence-lint job

ambroise-leclerc:
## Summary
- Adds a "Verify evidence artifacts" step (`ctest ... -L evidence`) to the Windows/MSVC, Linux/GCC-15, and Linux/GCC-16 jobs — each its own step, so a drifted artifact is unmistakable in the job log rather than one line among hundreds.
- Documents the two-toolchain byte-identity claim inline (MSVC + GCC in the same PR, per ADR-007) so a future "optimise CI to one leg" change is a deliberate reversal, not an unnoticed regression. Also notes where the Clang leg's step goes once that job is re-enabled — issue #48 added GCC 16 but did not turn Clang back on, so that half of the story is still open.
- New `evidence-lint` job: runs `mdux-evidence-lint` (#81) and its self-tests (no C++ toolchain needed, so a docs/tooling-only PR gets fast feedback), plus a check that every committed `report.json` carries a real git SHA (a committed `"unknown"` means the artifact's provenance can't be verified — ADR-007).

## Test plan
- [x] `.github/workflows/ci.yml` diffed only in the expected places against the parent branch (8 new step/job insertions, no unrelated changes)
- [x] The `toolGitSha` shell check extracted verbatim from this exact file and run against constructed fixtures: no `generated/` (no-op), `generated/` present but no reports (no-op), one good SHA (pass), one `"unknown"` mixed with one good (fails, names the offending file) — all four match expected behavior
- [x] This step is a no-op today (no `generated/` directory exists — the first bakers land with issues #13-#18); it's in place ahead of them so nothing has to remember to add it later

Stacked on #84 (host tools). **Last PR in the #12 Evidence kernel stack** (issue #55) — merging this completes Epic #12, the sole gate on Track C (#13, .medui renderer slice) and Track D (#18, zero-SOUP ML) per the Wave 2 roadmap.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,14 @@ jobs:
       run: ctest --preset ninja-msvc
       shell: cmd
 
+    # Byte-identity of every baked artifact, on this toolchain (ADR-007). Run as its own step
+    # rather than relying on the full suite above, so a drifted artifact is unmistakable in the
+    # job log instead of one line among hundreds. The corresponding Linux/GCC step is what makes
+    # the claim two-toolchain - see the comment there before removing either.
+    - name: Verify evidence artifacts (MSVC)
+      run: ctest --preset ninja-msvc -L evidence
+      shell: cmd
+
     - name: Verify no source-tree writes
       shell: pwsh
       run: |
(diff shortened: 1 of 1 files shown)


camille-martin-paris:
Review finding (blocking): the new `ctest -L evidence` steps will fail as soon as the first committed report exists, because #82 embeds configure-time `HEAD` in the bytes being compared.

A report baked before its commit contains H0; CI on H1 or the PR merge SHA emits a different report. The grep job only rejects `"unknown"`; it does not—and cannot—repair that byte mismatch or prove that a non-unknown SHA identifies the relevant tool sources.

Please resolve the stable provenance identity first, then add an end-to-end CI fixture that includes at least one committed report so this job is not a no-op.

ambroise-leclerc:
Confirmed and fixed (63a87f4). Rather than build the fixture you asked for, I removed the check itself — it was verifying a field (`toolGitSha` == `"unknown"`) that no longer exists, because `BakeReport` dropped it entirely (see #82 and the ADR-007 decision-5 rewrite on #79). There's no lesser-strict version of that check that would have been correct to keep: the field it inspected was the self-referential one, not a separate, fixable validation.

The `evidence-lint` job now only runs `mdux-evidence-lint` and its self-tests (unaffected by this — those were never about `toolGitSha`). The three `Verify evidence artifacts` steps on the toolchain legs are unaffected by this fix and remain as originally reviewed.

Issue #23, opened 2026-07-26
S4 — Purge reproduced standards from git history


ambroise-leclerc:
Deleting a tracked 1.8 MB copy of a draft IEC standard from `HEAD` does not remove it from the repository. Anyone cloning still gets it, and GitHub still serves it at the old blob URL.

**Preferred:** `git filter-repo` over the paths identified in S2, then force-push all branches and tags.

**Coordination required before doing this** — a history rewrite invalidates every existing clone and open branch:

- Local branches `develop`, `gcc16`, `5-configure-an-agentsmd-for-ai-support`
- ~20 stale `origin/*` branches (`#17`, `35tuple`, `45uniqueptr`, `BigPR`, `BiggerPR`, `metautilsCecile`, `=r`, and others) — most look abandoned and could be pruned first, which shrinks the rewrite considerably
- Any open PRs

**If the rewrite is refused** — a legitimate decision, given the disruption — then this issue instead delivers a written record of that decision and its residual risk, referenced from the S1 ADR. What is not acceptable is leaving it undecided.

Part of #7.

ambroise-leclerc:
Done. Ran `git filter-repo` on a fresh mirror clone (never touched the working repo directly), then force-pushed the rewritten history.

**Purged (15 paths, not just the 6 originally scoped — see below):**
- `docs/MduX-IEC-62304-AI-Reference.md`, `docs/MduX-ISO-13485-AI-Reference.md`
- `inputs/Documentation/{IEC-DIS-62304-2.pdf, IEC-62304-Complete.md, ISO-13485-Complete.md, Medical Device Cybersecurity for Engineers and Manufacturers.md}`
- **Found during verification, not in the original inventory** (all history-only — already removed from HEAD by an earlier, never-completed cleanup commit `da2f4b7`/`740ba14`, "Remove unwanted documentation", 2025-10-06):
  - `2022_Book_MedicalDeviceGuidelinesAndRegu.{pdf,epub}` — a full commercial book
  - `Developing_an_Iso_13485-certified_Quality_Management_System...pdf` — a full commercial book
  - `Medical Device Cybersecurity for Engineers and Manufacturers.pdf` — a second copy (separate from the .md above)
  - `Medical Device Cybersecurity for Engineers and Manufacturers/` — 295 page-by-page JPG scans of the same book
  - `NF62304-2006.pdf` — the French AFNOR adoption of IEC 62304
[...]

PR #86, opened 2026-07-27
S2 — Inventory files containing reproduced or paraphrased standard text

ambroise-leclerc:
## Summary
- Adds `docs/governance/reproduced-text-inventory.md`: the full-history audit trail issue #21 asks for, and the scope document issue #23 (purge) has to work from.
- The epic body for #7 described the problem as one PDF and two markdown transcriptions. Walking full git history found the real scope is much larger: **482 paths, 889 unique blob versions, ~105 MB** — including 303 JPG images that are a complete page-by-page scan of a copyrighted book, and 87 PNG screenshots of ISO 13485 and a second book.
- Surfaces a blocking problem for #23: **32 other remote branches still carry `inputs/`/`libstd/` live at their current tip**, not just in old history. A purge of `develop`/`master` alone would remove nothing in practice while these exist. 18 of the 32 are fully merged already (safe to delete); 14 carry unique commits and are flagged as not this document's call to discard.
- States plainly what a git history purge cannot guarantee (existing clones, GitHub-side caches, forks) so #23 isn't later described as more complete than it is.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py docs/governance/reproduced-text-inventory.md` — 0 findings
- [x] Every count in the document was recomputed from `git log --all --diff-filter=A` and cross-checked with an exact, non-overlapping partition (all 482 paths bucketed with zero left over) rather than taken from the epic body's original description
- [x] The two AI-Reference line counts (2,950 / 2,604) and the self-describing quote were verified directly against the historical blob, not assumed
- [x] The merged-vs-unmerged branch split (18 / 14) was computed via `git rev-list --count origin/develop..<branch>` for all 32 branches

Part of #7. Issue #23 (the actual purge) needs a decision on the 14 branches with unmerged commits before it can proceed — raising that separately.


camille-martin-paris:
Review complete. I found no blocking content issue. I fixed the article in “An 88th `.png`” and
pushed commit `dfca6df` to this branch.

Verification:

- `python3 tools/docs-lint/mdux_docs_lint.py docs/governance/reproduced-text-inventory.md` — 0 findings
- `git diff --check` — pass

Impact classification: not safety-relevant; this is an audit/governance documentation correction
only. Fresh CI is running after the push.

Issue #26, opened 2026-07-26
S1 — docs/governance/citation-convention.md and the Justification schema


ambroise-leclerc:
The document every other file in the corpus links back to, and the schema that makes a compliance claim machine-checkable.

**`docs/governance/citation-convention.md`** states the rule from E1/S1 and the exact key format, and explains *why* no normative text is reproduced — so a reader who wonders where the standard's own words went finds the answer in one hop.

**`docs/iec62304/schemas/justification.schema.json`** — the shared object, referenced by all five standards:

```json
{
  "justification_id": "JUS-014",
  "standard": "IEC 62304:2006",
  "clause_ref": "IEC 62304:2006 §5.3.3 Identify segregation necessary for risk control",
  "rationale": "Governed modules are compiled without Vulkan or platform include directories, making the trust-zone boundary a compile error rather than a review comment.",
  "requirement_id": "REQ-ARCH-004",
  "evidence_refs": ["cmake/MduXTrustZones.cmake", "docs/adr/ADR-005-trust-zones.md"]
}
```

`justification_id` is unique across the whole corpus, not per standard. `evidence_refs[]` must contain real repository paths — E1/S6's lint verifies they exist.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

PR #87, opened 2026-07-27
S1 — docs/governance/citation-convention.md and the shared Justification schema

ambroise-leclerc:
## Summary
- Completes `docs/governance/citation-convention.md`, which shipped with #22 as an explicitly-deferred "minimal first version".
- Adds `docs/governance/schemas/justification.schema.json` (Draft 2020-12, `additionalProperties: false`) and replaces the illustrative (unvalidated `jsonc`) preview block with two real worked examples that validate against it and against `mdux-docs-lint`'s existing justification checks.
- The schema cross-checks that `clause_ref` actually starts with the same standard named in `standard`, via a per-standard `if`/`then` pair — a plain regex on `clause_ref` alone can't reference a sibling property.

## Test plan
- [x] Schema validated as well-formed Draft 2020-12 (`jsonschema.Draft202012Validator.check_schema`)
- [x] 10 cases tested with `jsonschema`: the valid example, the same example with its optional field omitted, and 8 rejection cases (empty `evidence_refs`, missing `rationale`, unapproved standard, malformed `justification_id`, unknown extra field, duplicate `evidence_refs`, malformed `clause_ref`, and — the one that actually caught a real bug in my first draft — **`clause_ref` naming a different standard than `standard`**, which a regex-only version silently accepted
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings on both new files
- [x] Both worked examples' `evidence_refs` paths verified to exist in the repository before committing

Part of #8. Issue #26 complete.

### DIFF
--- a/docs/governance/citation-convention.md
+++ b/docs/governance/citation-convention.md
@@ -6,10 +6,13 @@ why. This is the replacement: how to cite a standard's clause without reproducin
 
 ## Status
 
-This is a minimal first version, landed alongside the removal it explains (issue #22) so the tree
-is never in a state where the reproduced material is gone and nothing describes what replaces it.
-The full convention — including the `Justification` object's JSON Schema and worked examples — is
-tracked as issue #8, S1, and will extend this document rather than replace it.
+Landed in two parts. The first (issue #22) shipped alongside the reproduced-text removal so the
+tree was never in a state where the material was gone and nothing described what replaces it. This
+version (issue #8, S1) adds the formal `Justification` schema and worked examples, completing it.
+
+`docs/iec62304/` and `docs/iso13485/` do not follow this convention yet — see "Known gap" below.
+Issues #27/#28 are the clause-accurate rewrite that fixes this; this document defines the target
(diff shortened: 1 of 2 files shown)


camille-martin-paris:
Review complete. I checked the citation-convention changes, the Draft 2020-12 schema structure,
the standard/clause cross-check, the worked examples, and every cited evidence path. I found no
actionable issue and made no code changes.

Verification:

- `python3 tools/docs-lint/mdux_docs_lint.py` — 33 files checked, 0 findings
- JSON syntax and schema constraints were inspected; the optional local `jsonschema` package is not
  installed in this review environment, so I did not independently rerun `check_schema`
- Documentation Lint and Compliance & Documentation pass in CI

Impact classification: not safety-relevant at runtime; this defines governance and traceability
metadata. The current GCC 15 and MSVC jobs are failing while the documentation-specific checks pass,
so those build failures should remain a merge-gate concern even though they do not appear caused by
this documentation-only diff.

PR #90, opened 2026-07-27
S3 — New docs/iso14971/

ambroise-leclerc:
## Summary
- Adds `docs/iso14971/` — new content, not a rewrite. Seven files matching the standard's real structure (§1–3, §4, §5, §6–7, §8–9, §10), plus a README.
- ISO 14971 describes a risk management process run against a *specific device's* intended use. MduX has no device and no device-level analysis to run this against — most clauses here say so plainly rather than stretching a citation to fit.
- The two genuine MduX-level analogs are narrow: the trust-zone architecture (ADR-004) forecloses a hazardous-situation category by construction, and the exception-disabled governed zone (ADR-005) addresses a second, narrower one. Both cited as inherent-safety-by-design control measures, verified mechanically.
- Converts three `docs/iso14971/` plain-text placeholders in the existing corpora to real links, now that the directory exists.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (41 files)
- [x] All markdown links resolve across all three regulatory corpora (checked programmatically)
- [x] Both `Justification` objects validate against `docs/governance/schemas/justification.schema.json`
- [x] All `evidence_refs` paths verified to exist

Stacked on #89 (issue #28). Part of #8, issue #29.

Issue #30, opened 2026-07-26
S5 — New docs/iec62366/


ambroise-leclerc:
IEC 62366-1:2015 (usability engineering for medical devices) is **entirely absent** from MduX today — and it is the standard most directly relevant to a project whose whole purpose is medical device *user interfaces*.

**Target**, mirroring TrustSC's `docs/iec62366/`:

- `01-scope-and-general-requirements.md`
- `02-use-specification-and-ui-design.md`
- `03-formative-and-summative-evaluation.md`
- `AI-Reference.md`
- `schemas/usability-engineering-record.schema.json`

This is the corpus that gives `.medui`'s safety features their regulatory footing: `@safety_critical` annotations, the `requirement:` binding on critical components, text-budget validation against every approved locale, and the rendered-truth verification in E10 are all usability risk controls before they are anything else. Write the modules so those mechanisms are what the clauses point at.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

Issue #29, opened 2026-07-26
S4 — New docs/iso14971/


ambroise-leclerc:
ISO 14971:2019 currently exists in MduX only as the monolithic `docs/MduX_ISO-14971-Risk-Management-Framework.md` (713 L) plus `risk-assessment-templates.md` (935 L) at the repository root.

**Target**, mirroring TrustSC's `docs/iso14971/`:

- `01-scope-and-general-requirements.md`
- `02-risk-analysis-and-evaluation.md`
- `03-risk-control-and-residual-risk.md`
- `04-production-and-post-production.md`
- `AI-Reference.md`
- `schemas/risk-record.schema.json`

The risk schema must be field-aligned with the `Hazard` type from E3/S1 — specifically the rule that a `Hazard` carries a non-empty `controlled_by` linking to a requirement, which is the machine-checked integration point between ISO 14971 and IEC 62304 §4.2.

`risk-assessment-templates.md` should be reviewed against E1/S2's inventory and then either folded into `software_development_file/templates/` (E3/S4) or deleted.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

Issue #27, opened 2026-07-26
S2 — Rewrite docs/iec62304/ against the real clause structure


ambroise-leclerc:
Today `docs/iec62304/README.md` maps its six modules to "sections 1-16" — a flat renumbering that does not exist in IEC 62304. Nothing in that corpus can be cited by a real clause number, which blocks the `Justification` object, the lint, and every cross-reference from an SDF document.

**Target structure**, matching the standard's actual clauses:

| Module | Clauses | Focus |
|---|---|---|
| `01-scope-and-general-requirements.md` | §1-§4 | Scope, definitions, QMS/risk integration, safety classification, legacy software |
| `02-development-planning-and-requirements.md` | §5.1-§5.2 | Development planning, requirements analysis |
| `03-development-design.md` | §5.3-§5.4 | Architectural and detailed design, segregation, SOUP identification |
| `04-development-implementation-and-testing.md` | §5.5-§5.8 | Unit implementation/verification, integration, system testing, release |
| `05-maintenance-process.md` | §6 | Maintenance plan, problem/modification analysis |
| `06-risk-management-process.md` | §7 | Software hazard analysis, risk control, verification |
| `07-configuration-management-process.md` | §8 | Configuration identification, change control, status accounting, SOUP |
| `08-problem-resolution-process.md` | §9 | Problem report handling, evaluation, traceability |

Original prose throughout, per E1/S1. Where a clause has a real MduX mechanism behind it, point at it — a type, an ADR, a CI step — rather than restating the requirement abstractly.

**MduX-specific divergence from TrustSC:** Class A stays in scope here. TrustSC models Class B/C only and says so explicitly; MduX's README claims A/B/C. Whichever is true, state it once in `01-*.md` §4.3 and make the code agree.

The existing `ai-automation-schemas/` and `code-examples/` under `docs/iec62304/` should be folded into the new structure (`schemas/`), not duplicated alongside it.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

PR #91, opened 2026-07-27
S4 — New docs/iec62366/ (IEC 62366-1) [confidence flag - see PR body]

ambroise-leclerc:
## Summary
- Adds `docs/iec62366/` — new content covering IEC 62366-1's usability engineering process (§1–3 scope/terms, §4 general requirements, §5 the process itself, application specification through summative evaluation).
- **Confidence flag, stated prominently in the README**: sub-clause numbering here (`§5.1`–`§5.9`) is written from general professional familiarity with the process, not verified against the standard text (which this project doesn't hold a copy of). Confidence in the *named steps* is higher than in their *exact numeric position*. This is meaningfully lower confidence than #27–#29, and is flagged as needing follow-up verification rather than presented with false precision.
- No `Justification` objects: the process needs actual users and a specific device's UI, which MduX doesn't have. Every §5 sub-clause states plainly that no MduX mechanism exists yet.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (45 files)

Stacked on #90 (issue #29). Part of #8, issue #30.

**Flagging for reviewer attention:** this PR's clause numbering should be checked against the actual IEC 62366-1 text before being relied on for anything compliance-facing — see the confidence note in the README for specifics.

### DIFF
--- a/docs/iec62366/01-scope-and-terms.md
+++ b/docs/iec62366/01-scope-and-terms.md
@@ -1,0 +1,34 @@
+# IEC 62366-1:2015 §1–§3 — Scope, normative references, terms and definitions
+
+## §1 Scope
+
+IEC 62366-1 specifies a process for a manufacturer to analyze, specify, design, verify, and
+validate the usability of a medical device's user interface, so far as usability relates to safety.
+It is explicitly scoped to *use-related* risk — errors and hazards arising from how a user
+interacts with the device — rather than the device's broader functional risk, which ISO 14971
+covers.
+
+MduX renders content described by a device's own UI design; it does not itself define what that
+device's users see or how they interact with it in a specific clinical context. This corpus
+documents where MduX's rendering and compilation mechanisms bear on use-related risk regardless of
+the device (chiefly, once built: a compile-time-enforced content budget that keeps critical
+information from being silently truncated or occluded) and states plainly where the process this
+standard describes remains entirely the integrating device's responsibility.
+
+## §2 Normative references
+
+IEC 62366-1 is written to be read alongside ISO 14971, for the risk-management framework its
+use-related risk analysis sits inside. See [`../iso14971/`](../iso14971/) for this project's
+treatment of that standard.
+
+## §3 Terms and definitions
+
+As elsewhere in this corpus, definitions are not restated. The terms below recur across this
+directory.
+
+| Term | Where MduX makes it concrete |
+|---|---|
+| User interface | For MduX, the concrete artifact is a compiled `.medui` screen — a `CompiledScreenPackage` (issue #15) — not the source `.medui` file, since the runtime only ever sees the compiled form. |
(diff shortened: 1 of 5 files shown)


camille-martin-paris:
Wave 2 review: `mdux-docs-lint` passes (45 files, 0 findings), and the confidence warning is
honest, but that warning also establishes that this PR cannot yet satisfy issue #30's
clause-accurate deliverable.

Blocking scope gaps:

- The §5.1–§5.9 numbering is explicitly unverified against IEC 62366-1. The regulatory-citation
  protocol requires confirming a clause number before citing it.
- #30 requires `schemas/usability-engineering-record.schema.json`; none is added, and #94 later
  supplies only a clause-index schema.
- The roadmap calls for the corpus to connect `.medui` safety annotations, requirement binding,
  locale text budgets, and rendered-truth verification to usability controls. The PR mostly records
  that these mechanisms do not exist yet; that is honest status, but it is not completed Wave 2
  coverage.

Impact classification: regulatory-documentation relevant. Maintainer/domain-expert verification
against the actual standard is required before this can be called clause-accurate. No GitHub checks
are currently reported for this stacked branch.

ambroise-leclerc:
All three points addressed (commit `bf8e04a`).

**The unverified clause numbers are removed, not caveated.** You were right that a confidence warning cannot substitute for confirming a clause number before citing it, and that a directory calling itself "clause-accurate" while saying its numbering is unchecked states two incompatible things. `03-usability-engineering-process.md` no longer numbers its sections: a warning label does not stop a number being copied into a design history file, and a reader who copies it has cited something nobody checked. The named process steps are what the corpus asserts, stated as headings, and the README title now says what is true — top-level clauses only, §1 through §5.

**This part remains genuinely blocked, and is flagged rather than absorbed.** Restoring the sub-clause numbering needs a maintainer with access to IEC 62366-1:2015 checking each heading against §5. I cannot do that without the standard, and guessing is what produced the problem. It is bounded work and stays on issue #30.

**`schemas/usability-engineering-record.schema.json` is added.** Its `process_step` is named rather than numbered, for the same reason as above — unverified numbers in machine-readable form are worse than in prose. `control_status` is required with no default: an `implemented` record must carry evidence and a `planned` one must name the issue that would build it, because a usability engineering file citing a control that does not exist is the exact failure the standard's evaluation steps exist to catch.

**The `.medui` mechanisms are now named concretely.** The §5 module tabulates `@safety_critical`, `requirement:` binding, locale text budgets (#15) and rendered-truth verification (#16), with what each would and would not establish — and none dressed as an existing control.

Verification: `mdux-docs-lint` 44 files, 0 findings; schema parses as JSON.

Impact classification: regulatory-documentation relevant. Domain-expert verification of the §5 numbering is still required before this corpus is used compliance-facing.

Issue #31, opened 2026-07-26
S6 — New docs/iec81001/


ambroise-leclerc:
IEC 81001-5-1:2021 (health software cybersecurity) — absent from MduX. The only related material is `inputs/Documentation/Medical Device Cybersecurity for Engineers and Manufacturers.md`, a copyrighted book being removed in E1/S3.

**Target**, mirroring TrustSC's `docs/iec81001/`:

- `01-scope-and-relationship-to-iec62304.md`
- `02-security-risk-management.md`
- `03-secure-design-and-implementation.md`
- `04-security-verification-and-update-management.md`
- `AI-Reference.md`
- `schemas/security-risk-record.schema.json`

**Carry TrustSC's caveat:** its `docs/iec81001/README.md` states that the clause numbering in its corpus is *provisional* and warns readers before relying on those clause references. MduX should carry the same caveat rather than implying more precision than the authors have.

The natural MduX hooks: the SOUP register (E3/S3), the byte-verified evidence chain (E6) as a supply-chain integrity control, and `Classifier1D`'s digest check (E12/S7) as an integrity control on a shipped artifact.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

PR #92, opened 2026-07-27
S5 — New docs/iec81001/ (IEC 81001-5-1) [lowest confidence - see PR body]

ambroise-leclerc:
## Summary
- Adds `docs/iec81001/` — the fifth and final standard in the citation convention's closed set. **Lowest-confidence corpus of the five, stated prominently (twice) in the README.**
- IEC 81001-5-1:2021 is newer and more specialized than the other four, and this project holds no copy of its text. What's asserted with confidence: the universal IEC front-matter skeleton through clause 4 (a convention already relied on for IEC 62304 and IEC 62366-1), and the well-documented security-lifecycle practice categories the standard is understood to require (closely aligned with IEC 62443-4-1). What's *not* asserted: specific sub-clause numbers for those practices.
- One `Justification`, cited conservatively at §4 (general requirements) rather than a specific practice sub-clause: the trust-zone architecture as a secure-by-design property.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (48 files)
- [x] The one `Justification` validates against the schema

Stacked on #91 (issue #30). Part of #8, issue #31.

**Flagging for reviewer attention, same as #91 but more strongly:** this corpus should be checked against the actual IEC 81001-5-1 text before anything below clause 4 is treated as load-bearing.


camille-martin-paris:
Wave 2 review: `mdux-docs-lint` passes (48 files, 0 findings), but this PR does **not** complete
issue #31.

Blocking scope gaps:

- The README labels the directory a “clause-accurate reference” while also saying nothing below
  §4 has been checked against the standard. Those claims are incompatible; this remains a
  provisional practice-category overview.
- #31 specifies separate security-risk, secure-design/implementation, and
  verification/update-management modules. This PR collapses them into one unnumbered practice file.
- #31 requires `schemas/security-risk-record.schema.json`. #94 explicitly excludes IEC 81001 from
  schema output, so the required deliverable is absent from the full stack.

The confidence caveat should remain, but it cannot substitute for the planned verification and
schema. Impact classification: regulatory-documentation relevant; domain-expert review is still
required. No GitHub checks are currently reported for this stacked branch.

ambroise-leclerc:
All three blockers addressed (commit `c24a77b`).

**The two incompatible claims are resolved.** The README called this a "clause-accurate reference" while stating nothing below §4 had been checked. The title was the wrong one — a reference is not clause-accurate at a depth it does not cite — and now reads "practice reference, top-level clauses only". The caveat itself stays, per issue #31's instruction to carry it.

**The collapsed practice file is split** into the three modules #31 names: `02-security-risk-management.md`, `03-secure-design-and-implementation.md`, `04-security-verification-and-update-management.md`. `01-scope-and-terms.md` is renamed to `01-scope-and-relationship-to-iec62304.md` as the issue names it, and given the activity-by-activity mapping to IEC 62304's life cycle that the filename promises — the practical point being that a manufacturer runs one process, not two.

**`schemas/security-risk-record.schema.json` is added**, and it is an extension of the ISO 14971 risk record rather than a replacement. That is the substantive decision: a security failure that can lead to harm is a hazard, and this standard adds the analysis that finds it (threat, asset, weakness, attack surface), not a second risk file to record it in. `id`, `description` and `controlled_by` stay field-aligned with `mdux::governance::Hazard`, checked by the drift tool added on #94, so a security hazard cannot escape the review every safety hazard gets.

The modules also name the hooks #31 identifies — the SOUP register, the byte-verified evidence chain as a supply-chain integrity control, and `Classifier1D`'s digest check — each marked present or planned rather than implied. What no mechanism covers is stated too: no fuzzing of the parsing surfaces, no coordinated disclosure route, no security triage, no downstream notification. Those are gaps, not scope exclusions.

Verification: `mdux-docs-lint` 49 files, 0 findings; schema parses as JSON.

Impact classification: regulatory-documentation relevant. Domain-expert review of the clause structure below §4 is still required, and is tracked on #31.

Issue #32, opened 2026-07-26
S7 — AI-Reference.md per standard


ambroise-leclerc:
One compact index per standard: **one row per clause**, a one-sentence pointer, and a deep link into the module that covers it. Used to locate the right module fast, then open it for the actual prose.

This replaces the two monolithic files deleted in E1/S3 — `docs/MduX-IEC-62304-AI-Reference.md` (2,950 L) and `docs/MduX-ISO-13485-AI-Reference.md` (2,604 L) — which were transcriptions, not indexes.

**Shape**, following TrustSC:

```markdown
## §5.1-§5.2 — Planning and requirements ([detail](02-development-planning-and-requirements.md))

- **§5.2.1 Define requirements** — `Requirement { id, title, source_clause, verification_intent }`. [→](02-development-planning-and-requirements.md#521-defining-requirements)
- **§5.2.4 Verify requirements** — every `Requirement` needs ≥1 `VerificationCase` (`ComplianceProgram::validate()`). [→](02-development-planning-and-requirements.md#524-verify-software-requirements)
```

**Two invariants the lint should enforce:**

1. No clause is stubbed or condensed to a placeholder — if a row exists, its module has real content behind the link.
2. Every clause referenced by any module appears here, and every row here links to a real anchor.

The pointer should name a real MduX mechanism wherever one exists, not paraphrase the clause.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

PR #93, opened 2026-07-27
S6 — AI-Reference.md per standard

ambroise-leclerc:
## Summary
- Adds `tools/docs-lint/generate_ai_reference.py` and runs it against all five regulatory corpora, producing a one-row-per-clause `AI-Reference.md` in each (clause, title, source file, `Justification` IDs).
- Deliberately mechanical: the script reads what's already in each corpus (headings + `Justification` blocks) rather than deciding what a clause covers. It can't drift from the prose it indexes without the source changing too.
- **A real bug caught before committing, not after:** the first version matched Justifications to clause headings by string comparison and silently dropped IEC 81001-5-1's one Justification (cited at §4, which doesn't match any practice-category heading in that file, deliberately, per that corpus's confidence limits). Fixed by attributing by *position* (which section a `json` block textually falls under) instead — added as a regression test.
- Updates each README to link the real file instead of the issue #32 placeholder; the two lower-confidence corpora note that the generated index inherits their existing caveats rather than implying new confidence.

## Test plan
- [x] 12 new tests for the generator, all passing (16 total in `tools/docs-lint/`)
- [x] Cross-checked against source of truth before committing: 11 unique `JUS-003`–`JUS-013` across all five generated indexes, matching a direct grep of every `justification_id` field exactly
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (53 files)
- [x] Every markdown link resolves

Stacked on #92 (issue #31). Part of #8, issue #32.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -482,6 +482,14 @@ jobs:
         python3 -m unittest discover -s tools/docs-lint -p "test_*.py"
         python3 tools/docs-lint/mdux_docs_lint.py
 
+    # Every AI-Reference.md is generated, so an edited module with a stale index is a defect the
+    # build should catch rather than a reader. --check writes nothing; it fails when regenerating
+    # would change a byte, and when a clause is cited in prose that no index row covers.
+    - name: Check AI-Reference indexes are current
+      env:
+        PYTHONDONTWRITEBYTECODE: "1"
+      run: python3 tools/docs-lint/generate_ai_reference.py --check
+
   evidence-lint:
     name: Evidence Pipeline Lint
     runs-on: ubuntu-latest
--- a/docs/iec62304/01-scope-and-terms.md
+++ b/docs/iec62304/01-scope-and-terms.md
@@ -29,6 +29,7 @@ convention names — see [`docs/governance/citation-convention.md`](../governanc
 for the full closed set of approved identifiers.
 
 ## §3 Terms and definitions
+<!-- pointer: Names the terms that recur across this corpus and points at where MduX gives each a checkable meaning, rather than restating a glossary that is itself normative text. -->
 
 Rather than restate the standard's glossary — which is exactly the kind of clause where a
 "definition" is normative text, easy to reproduce nearly word-for-word without meaning to — this
--- a/docs/iec62304/02-general-requirements.md
+++ b/docs/iec62304/02-general-requirements.md
@@ -1,6 +1,7 @@
 # IEC 62304:2006 §4 — General requirements
 
 ## §4.1 Quality management system context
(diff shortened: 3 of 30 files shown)


camille-martin-paris:
Wave 2 review: the generator is deterministic and its 16 tests pass, and `mdux-docs-lint` reports
0 findings. However, the generated files do **not** meet issue #32's required index shape or
invariants.

Blocking findings:

- Rows link only to the source file, not to the clause heading's deep anchor.
- Rows repeat the heading title; they do not provide the required one-sentence pointer naming a
  real MduX mechanism where one exists.
- IEC 62304 §9 is emitted as clause `—` because the generator ignores the numbered H1 when the
  file contains an unnumbered H2.
- The tests do not enforce either issue invariant: every referenced clause indexed with a real
  anchor, and every index row backed by substantive clause content.

This needs a generator/index-shape redesign rather than a small review patch. Impact classification:
traceability-documentation relevant, no runtime behavior changed. No GitHub checks are currently
reported for this stacked branch.

ambroise-leclerc:
Redesigned rather than patched, as you asked (commit `17f3152`). All four findings addressed.

**Deep anchors.** Every row now targets the heading's GitHub anchor, with the same de-duplication suffix GitHub applies when two headings in one file slugify identically. A test resolves all 101 links against the real corpora.

**Pointers, not restated titles.** Resolved in three steps: an authored `<!-- pointer: ... -->` comment, then the section's first `Justification` rationale — which names a real mechanism by construction, since `mdux-docs-lint` checks its evidence paths exist — then the first usable sentence naming MduX. A section yielding none of the three is a **generator error**, not a padded row, which is what stops the old behaviour returning. Table rows and sentences opening with a bare pronoun are rejected, since both read as nonsense once lifted into a table cell. 24 pointers are authored where no derivation was good enough.

**§9 is indexed as §9.** A section with no clause number of its own inherits its file's. Inheritance is restricted to files naming a *single* clause: a narrative section inside a `§1–§3` file is not all three of those clauses, and indexing it as such would assert a mapping nobody made.

**Both invariants are enforced, against the real corpora.** A row must have substantive content behind its link, and every clause cited in prose must have a row. The second found real gaps while being written — clauses discussed in prose with no section of their own — and the IEC 81001 relationship table now uses full citation keys so another standard's clause numbers are not read as this corpus's own.

Also adds `--check`, wired into the Documentation Lint job, so an edited module with a stale index is a build failure rather than something a reader notices later.

Verification: 37 generator tests pass (was 16); `mdux-docs-lint` 54 files, 0 findings; `--check` clean on all five corpora.

Impact classification: traceability-documentation relevant; no runtime behaviour changed.

Issue #33, opened 2026-07-26
S8 — JSON Schemas per standard


ambroise-leclerc:
`docs/<standard>/schemas/*.json`, field-aligned with the governance types from E3/S1 — the alignment is the point, not the files.

| Schema | Standard | Aligns with |
|---|---|---|
| `requirement.schema.json` | IEC 62304 | `mdux::governance::Requirement` |
| `hazard.schema.json` | IEC 62304 | `Hazard` (non-empty `controlled_by`) |
| `verification-case.schema.json` | IEC 62304 | `VerificationCase` |
| `safety-classification.schema.json` | IEC 62304 | `SafetyClass` |
| `justification.schema.json` | shared | E2/S1 |
| `quality-management-system.schema.json` | ISO 13485 | — |
| `risk-record.schema.json` | ISO 14971 | `Hazard` + risk evaluation |
| `usability-engineering-record.schema.json` | IEC 62366-1 | — |
| `security-risk-record.schema.json` | IEC 81001-5-1 | — |

Fold in the existing `docs/iec62304/ai-automation-schemas/*.json` and `docs/iso13485/ai-automation-schemas/*.json` rather than duplicating them — three of those five files already cover this ground and should be revised into place.

Add a ctest or lint step asserting the schemas and the C++ types have not drifted. A schema that silently diverges from the type it documents is worse than no schema.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

PR #89, opened 2026-07-27
S3 — Rewrite docs/iso13485/ to the same clause-accurate shape

ambroise-leclerc:
## Summary
- Replaces `docs/iso13485/` entirely. The previous version built each module around fabricated per-file metadata (invented "AI decision matrices," "automation schemas," "code examples" counts) — the same pattern #27 already fixed for `docs/iec62304/`.
- New structure: six files matching ISO 13485:2016's real top-level clauses (§1–3, §4, §5, §6, §7, §8), plus a README.
- Unlike IEC 62304, most of this standard describes a manufacturer's own organizational/manufacturing processes MduX doesn't have (no employees, no facility, no certified QMS). Most of §5 and §6, and several sub-clauses of §7–8, say plainly that no MduX-specific mechanism applies rather than stretching a citation to fit — that's a deliberate choice, not a gap I missed.
- Four `Justification` objects, concentrated where this standard actually overlaps a software repository: §4.1 (evidence pipeline as control over an "outsourced process"), §7.3 design and development (trust-zone architecture + the digest module's externally-verified tests), §8.2 monitoring and measurement (the CI matrix).
- Updates `citation-convention.md`: both regulatory directories are now compliant, closing the gap tracked since #22.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (34 files)
- [x] All markdown links resolve (checked programmatically)
- [x] All 4 `Justification` objects validate against `docs/governance/schemas/justification.schema.json`
- [x] Every `evidence_refs` path verified to exist before committing
- [x] 278 lines total vs. four files well over 100KB combined in the version replaced

Stacked on #88 (issue #27). Part of #8, issue #28.

PR #88, opened 2026-07-27
S2 — Rewrite docs/iec62304/ against the real clause structure

ambroise-leclerc:
## Summary
- Replaces `docs/iec62304/` entirely. The previous version used a flat "sections 1-16" numbering invented for this repository — not IEC 62304:2006's actual structure — and, per the audit already in `citation-convention.md`'s "Known gap" section, contained sentences reading as close paraphrase of the standard's §1.1 wording.
- New structure: seven files, one per real clause (§1–3, §4, §5.1–5.8, §6, §7, §8, §9), plus a README. Every clause's prose is original — explains the requirement's intent, then either cites a real MduX mechanism via a `Justification` object or says plainly none exists yet (most of the actual product hasn't been built, so most clauses honestly have nothing to cite).
- Six `Justification` objects link real mechanisms: ADR-004's trust-zone architecture, ADR-005's exception-disabled governed zone, ADR-007's evidence pipeline, and the FIPS-180-4-verified digest module.
- Deletes `ai-automation-schemas/` and `code-examples/` — the former's README listed 16 schema files, of which 2 ever existed. That's the aspirational-documentation pattern issue #8 exists to fix, not one worth carrying into the rewrite. JSON Schemas for these mechanisms are issue #33, to be built for real.
- Updates `citation-convention.md`'s "Known gap": `iec62304/` is compliant now; `iso13485/` isn't yet (issue #28).

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (32 files)
- [x] Every markdown link in the new corpus resolves to a real file or directory (checked programmatically, not by eye)
- [x] All 6 `Justification` objects validate against `docs/governance/schemas/justification.schema.json` (`jsonschema.Draft202012Validator`)
- [x] Every `evidence_refs` path in every `Justification` verified to exist before committing
- [x] 390 lines total across the new corpus vs. 5,435 in the six files replaced — the old size was itself padding (emoji headers, repeated scaffolding, fictional automation claims), not substance

Stacked on #87 (issue #26 — the `Justification` schema this corpus cites). Part of #8, issue #27.

PR #94, opened 2026-07-28
S7 — JSON Schemas per standard [last story in #8]

ambroise-leclerc:
## Summary
- Adds `docs/<standard>/schemas/clause-index.schema.json` for four of the five corpora, and extends `generate_ai_reference.py` to emit a companion `AI-Reference.json` validated against it.
- Scoped deliberately narrow: formalizes exactly the `AI-Reference.md` row shape (clause, title, file, Justification IDs) that #32 already built, rather than inventing new schema content — the old deleted per-standard schemas described 16 automation workflows that mostly never existed, and this issue exists specifically to not repeat that.
- **IEC 81001-5-1 is deliberately excluded** (both JSON export and schema), encoded as an explicit `NO_JSON_EXPORT` set: that corpus's README already states lower confidence than the other four, and a formal schema would overstate it.

## Test plan
- [x] All four schemas verified to actually reject bad data, not just pass everything: 7 rejection cases each (malformed clause, empty title, wrong filename pattern, malformed Justification ID, duplicate IDs, unknown field, missing required field) against `jsonschema.Draft202012Validator`
- [x] All four generated `AI-Reference.json` files validate against their own schema
- [x] Generation is idempotent (running twice produces byte-identical output) — now a test, not just a manual check
- [x] The `NO_JSON_EXPORT` exclusion is tested directly (iec81001 gets no `.json`/schema output)
- [x] 20 tests total in `tools/docs-lint/`, all passing
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings, full repo (53 files)

Stacked on #93 (issue #32). **Part of #8, issue #33 — the last story in this epic.**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -490,6 +490,14 @@ jobs:
         PYTHONDONTWRITEBYTECODE: "1"
       run: python3 tools/docs-lint/generate_ai_reference.py --check
 
+    # Issue #33: the alignment between docs/<standard>/schemas/*.json and the governance types is
+    # the point of those schemas. A schema that has silently diverged from the type it documents
+    # still looks authoritative, which is what makes the drift worth failing a build over.
+    - name: Check schemas have not drifted from the governance types
+      env:
+        PYTHONDONTWRITEBYTECODE: "1"
+      run: python3 tools/docs-lint/check_schema_type_drift.py
+
   evidence-lint:
     name: Evidence Pipeline Lint
     runs-on: ubuntu-latest
--- a/docs/governance/schemas/clause-index.schema.json
+++ b/docs/governance/schemas/clause-index.schema.json
@@ -1,0 +1,73 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://mdux.dev/schemas/governance/clause-index.schema.json",
+  "title": "Clause index export",
+  "description": "Validates the `AI-Reference.json` that tools/docs-lint/generate_ai_reference.py writes beside each corpus's `AI-Reference.md` - the same rows, in a form a tool can read without parsing a markdown table. One schema covers all five standards: the row shape does not vary by standard, and five near-identical copies would be five places for a change to be applied four times. Regenerate rather than hand-edit if validation fails; the markdown and the JSON come from one pass over the corpus and cannot disagree.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["standard", "corpus", "rows"],
+  "properties": {
+    "standard": {
+      "type": "string",
+      "description": "The standard this index covers, spelled as in docs/governance/citation-convention.md.",
+      "enum": [
+        "IEC 62304:2006",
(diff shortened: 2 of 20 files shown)


camille-martin-paris:
Wave 2 review: the 20 generator tests and full docs lint pass, but this PR implements a different
deliverable from issue #33 and cannot be treated as the last story in epic #8.

Issue #33 requires governance/domain schemas:

- IEC 62304: requirement, hazard, verification-case, safety-classification, and shared justification
- ISO 13485: quality-management-system
- ISO 14971: risk-record
- IEC 62366-1: usability-engineering-record
- IEC 81001-5-1: security-risk-record
- plus a CTest/lint drift check against the corresponding C++ types

This PR instead adds four copies of a clause-index row schema and explicitly excludes IEC 81001.
That is useful export metadata, but it satisfies none of the domain-schema/type-alignment work
that gives #33 its purpose. The issue must either be implemented as written or explicitly
re-scoped; the current PR should not close #33/#8.

Impact classification: traceability/governance relevant, no runtime behavior changed. No GitHub
checks are currently reported for this stacked branch.

ambroise-leclerc:
Implemented as issue #33 is written, rather than re-scoped (commit `50eb5ff`).

**The four IEC 62304 domain schemas** — requirement, hazard, verification-case, safety-classification — each field-aligned with the matching type in `include/mdux/governance/Governance.cppm`. With the QMS, risk-record, usability and security-risk schemas added on #89, #90, #91 and #92, and the shared justification schema from #87, every row of #33's table now exists.

**`tools/docs-lint/check_schema_type_drift.py`** is the part you identified as giving the schemas their purpose. It checks member names in both directions, and that a schema `enum` matches the C++ `k*WireValues` array *in order* — the check that catches a new `VerificationMethod` enumerator whose wire spelling nobody added to the schema. It is explicit about what it does not check: types, patterns and semantics need a compiler and a specification, not a regex. It parses rather than compiles so it runs on the docs-only CI job, which is exactly the PR that introduces this drift.

Verified by mutation against the real governance module, not only by passing: a renamed C++ field and an added enumerator each fail it. The check reports and exits 0 when the module is absent, because it lands on the other stack (#95); I merged both stacks locally and confirmed it goes live and passes. 21 tests cover its failure paths regardless of branch.

**The clause-index export stays, corrected.** You called it useful metadata and it is, so `AI-Reference.json` is now written by the same pass as the markdown — the two cannot disagree — and carries the pointer and anchor the index gained on #93. The four near-identical schemas become one at `docs/governance/schemas/clause-index.schema.json`; the row shape does not vary by standard, and four copies are four places to apply a change three times.

Also corrects `docs/iec62304/README.md`, which still described MduX's own components as Class A "software that cannot contribute to a hazardous situation" — the same overclaim you corrected on #88, surviving in the one file that was not updated.

Verification: 65 docs-lint-directory tests pass; `mdux-docs-lint` 54 files, 0 findings; generator `--check` clean; drift check clean against the real module.

Impact classification: traceability/governance relevant; no runtime behaviour changed.

Issue #35, opened 2026-07-26
S2 — Traceability matrix export


ambroise-leclerc:
Requirement → verification → evidence, exported in the canonical JSON form from E6/S3 so it is diffable, committable, and byte-verifiable like every other evidence artifact.

Two exports:

- **`traceabilityMatrix()`** — every `Requirement`, the `VerificationCase`s that discharge it, and the `evidence_refs[]` those cases point at. Gaps are rows, not omissions: an unverified requirement must appear with an empty verification list, because an auditor's question is "what is *not* covered".
- **`releaseEvidenceSummary()`** — the IEC 62304 §5.8 release view: validation status, open `ProblemReport`s, the digest of every committed `generated/` artifact.

Both are pure functions over a `ComplianceProgram`. No I/O in the governed module — the caller writes the file.

Add a ctest that builds a small program with a deliberate gap (a requirement with no verification case, a hazard with an empty `controlledBy`) and asserts both that `validate()` fails and that the matrix names the specific gap.

Part of #9.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The governance/traceability API, SOUP register, SDF templates, regulatory records, schemas, and tests are present; post-merge CI run 30716625565 passed all seven jobs.

Issue #34, opened 2026-07-26
S1 — mdux.governance module


ambroise-leclerc:
`include/mdux/governance/Governance.cppm` + `src/governance/Governance.cpp`. **Governed zone** — no Vulkan, `std::expected` error handling per E5/S2.

The MduX equivalent of `crates/trustsc-governance`. Types, with the validation rules that make them worth having:

| Type | Key fields | Validation |
|---|---|---|
| `Requirement` | `id`, `title`, `sourceClause`, `verificationIntent` | needs ≥1 `VerificationCase` |
| `Hazard` | `id`, `description`, `controlledBy[]` | `controlledBy` must be non-empty — the machine-checked IEC 62304 §4.2 integration point |
| `VerificationCase` | `id`, `requirementId`, `method`, `evidenceRefs[]` | `requirementId` must resolve |
| `ProblemReport` | `id`, `description`, `closed`, `affectsRisk` | — |
| `AuditEvent` | `category` (Lifecycle / Verification / Change), `timestamp`, `subject` | — |
| `ComplianceProgram` | the aggregate | `validate()` enforces all of the above |

`ComplianceProgram::validate()` is what a release gate calls. It should return a structured failure list, not a bool — an unverified requirement and a hazard with no control are different problems and the caller needs to know which.

Field names and semantics must match the JSON Schemas in E2/S8; the two are checked against each other.

**Note for the README rewrite (E4/S4):** this is the first code in the repository that makes any of the compliance claims in `README.md`'s Implementation Status table true. Everything that table currently marks "Completed" for ISO 14971 / ISO 13485 / IEC 62304 does not exist.

Part of #9.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The governance/traceability API, SOUP register, SDF templates, regulatory records, schemas, and tests are present; post-merge CI run 30716625565 passed all seven jobs.

PR #95, opened 2026-07-28
S1 — mdux.governance module (Justification type)

ambroise-leclerc:
## Summary
- Adds `mdux.governance`: a governed-zone C++ type mirroring `docs/governance/schemas/justification.schema.json` field-for-field — `Justification` with `validate()`/`toJson()`/`write()`/`parse()`, following the exact pattern `mdux.evidence.report` (`BakeReport`) already established.
- This is the type issue #8's regulatory corpus already emits by hand in fenced `json` blocks (11 real objects across five standards). #35 (traceability matrix export) is what reads these back into this type to build a cross-corpus matrix; this PR is only the type and its validation.

## A real bug caught before it shipped
An early version of the clause-number parser scanned for "any digit or dot" rather than matching `[0-9]+(\.[0-9]+)*` group by group, so `§.3 leading dot` validated when it shouldn't have. Fixed to parse digit-groups explicitly; the failing case plus two adjacent ones (trailing dot, empty group between two dots) are now regression tests.

## Cross-validated against reality, not just synthetic cases
All 11 real `Justification` objects already committed across the five regulatory corpora (from the still-open #8 PR stack, #87–94) were extracted and fed through `Justification::parse()` with a throwaway driver — all 11 parse and validate cleanly, confirming the C++ type agrees with the already-shipped JSON Schema on every real object in the corpus, not just invented examples.

## Test plan
- [x] 74 tests total (11 new for governance; shares `evidence_tests`/`EvidenceTestMain` with the existing evidence-kernel suite, since governance is governed-zone for the same reason)
- [x] `python3 tools/evidence-lint/mdux_evidence_lint.py` — 0 findings
- [x] Full CMake configure (stub Vulkan package) succeeds, `evidence_tests` target present
- [x] All 11 real committed `Justification` objects parse and validate via a standalone driver program

**Independent of the #8 regulatory-docs stack** — branched from `develop` directly, matching the epic's "runs alongside #8" note. Does not depend on PRs #87–94 merging first.

Part of #9, issue #34.


camille-martin-paris:
Wave 2 review found both implementation defects and a story-level scope gap.

I fixed and pushed commit `79d6c01` for three JSON-Schema parity bugs:

- reject empty strings in `evidence_refs`
- require the exact single-space separator before `§`
- reject unknown JSON members to match `additionalProperties: false`

Regression tests were added. `mdux-evidence-lint` passes (10 files, 0 findings) and
`git diff --check` passes. This environment cannot compile the module because it has no CMake,
only GCC 13.3 (the project requires GCC 15+), and no Vulkan headers; fresh CI is running.

The PR still does **not** complete issue #34. That issue requires `Requirement`, `Hazard`,
`VerificationCase`, `ProblemReport`, `AuditEvent`, and aggregate `ComplianceProgram` types, with
structured validation of requirement-to-verification resolution and non-empty hazard controls.
This PR supplies only `Justification`, which is not the governance aggregate described by #34.

Impact classification: potentially safety-relevant because validation behavior and regulatory
traceability metadata changed; no rendering/device runtime behavior changed. The validation fixes
are ready for CI, but the missing governance model remains a review blocker for claiming Wave 2
coverage.

camille-martin-paris:
CI follow-up after `79d6c01`:

- GCC 16 passes.
- The new governance module and regression tests compile on MSVC; all governance tests shown in
  the log pass.
- MSVC remains red on pre-existing Unicode failures in `JsonTests.cpp`/`TomlTests.cpp`.
- GCC 15 remains red while compiling the pre-existing `src/evidence/Json.cpp`, where GCC 15.3
  raises `-Werror=array-bounds` in `std::char_traits`; `Justification.cppm` and
  `Justification.cpp` compiled before that failure.
- Documentation, compliance, evidence lint, and security checks pass.

The pushed validation fix is therefore verified on GCC 16 and by the MSVC governance tests; the
two remaining red jobs are outside this PR's changed code, but they still remain repository merge
gates.

ambroise-leclerc:
The story-level gap is closed (commits `f104f84`, `5fdc8e8`). Your three validation fixes are kept as-is; the checks they added are now the shared helpers every record type uses, so `Requirement::validate()` and `Justification::validate()` cannot drift on what an id or a citation key means.

**The governance model #34 describes now exists**: `Requirement`, `Hazard`, `VerificationCase`, `ProblemReport`, `AuditEvent`, and the `ComplianceProgram` aggregate.

`ComplianceProgram::validate()` returns the whole list of `ValidationFailure`s rather than a bool, each naming the record it is about — an unverified requirement and a hazard with no control are different problems, as #34 puts it. Two rules in it are the reason the type exists, and neither is checkable by any per-record `validate()`:

- every `Requirement` is covered by at least one `VerificationCase` (§5.2.4, reduced to something a build can fail on);
- every `Hazard` names at least one control **and** every control resolves to a real `Requirement` — the ISO 14971 §7 / IEC 62304 §4.2 join. A dangling control id is the failure this exists to catch and is invisible to a per-object check.

A record whose own shape is wrong is excluded from the cross-record rules, so one mistake produces one failure rather than a cascade.

`SafetyClass` is recorded as a *declared input* to a program, never inferred — consistent with the correction you made on #88. `docs/iec62304/schemas/safety-classification.schema.json` on #94 enforces the same discipline by requiring an `assigned_by`.

Layout follows #34: `Governance.cppm` is the single module interface (renamed from `Justification.cppm`), with `Governance.cpp` holding the vocabulary and shape checks shared by every record. Field names match the JSON Schemas on #94, and `tools/docs-lint/check_schema_type_drift.py` there now asserts that alignment mechanically.

**On CI.** Your follow-up noted GCC 16 green with GCC 15 and MSVC red on pre-existing code. That was still true, and the first push here also hit a **GCC 16 ICE** — a segfault in `module_state::mangle` on a defaulted hidden-friend `operator==` inside an exported struct. `5fdc8e8` declares it as a member instead; GCC 16 is green again.

The two remaining red legs are fixed in **#101**, which is green on all seven checks. Both were repository merge gates exactly as you described: a false `-Warray-bounds` in `src/evidence/Json.cpp` at `-O3`, and a missing `/utf-8` on MSVC — without it MSVC encodes a universal-character-name into the host ANSI code page rather than UTF-8, which is what failed the Unicode assertions. The fix there is the build, not the tests: `mdux.evidence.json` emits UTF-8 unconditionally per ADR-007.

Verification: 96/96 `evidence_tests` pass (22 new), 164/164 ctest bar the local-only `InstallTreeConsumer` CMake-version failure, docs-lint and evidence-lint both 0 findings.

Impact classification: potentially safety-relevant — validation behaviour and traceability metadata changed; no rendering or device runtime behaviour changed.

PR #96, opened 2026-07-28
S2 — Traceability matrix export

ambroise-leclerc:
## Summary
- Adds `mdux.governance.compliance`: `Requirement`, `VerificationCase`, `Hazard`, `ProblemReport`, aggregated into a `ComplianceProgram`, plus two pure exports over it - `traceabilityMatrix()` (requirement → verification → evidence_refs, one row per requirement) and `releaseEvidenceSummary()` (the IEC 62304 §5.8 release view: validation status, open `ProblemReport`s, digests of every committed `generated/` artifact the caller supplies).
- A separate dotted module rather than an addition to `mdux.governance` (#34's `Justification` type): follows the same one-module-per-concern split `mdux.evidence` already uses (`digest`, `json`, `report` as three modules, not one). `ComplianceProgram` has no dependency on `Justification`.
- No I/O anywhere in the module, per ADR-007's existing split between assembling evidence and exporting it - the caller walks `generated/`, hashes files, and populates `ComplianceProgram.generatedArtifacts`.

## The central design point
`validate()` and the two export functions read the same `ComplianceProgram` and deliberately disagree about what to do with a gap:
- `validate()` fails on an unverified `Requirement` or a `Hazard` with an empty `controlledBy` - those are release blockers.
- `traceabilityMatrix()` does **not** gate on `validate()` and does not skip the gap's row - it emits the requirement with an empty `verification_cases` array, because the question this document exists to answer is "what is *not* covered", and an omitted row cannot answer that.

The ctest (`traceabilityMatrix() names a coverage gap as an empty row, not an omission`) asserts both halves on the same deliberately-gapped program: `validate()` fails, and the matrix still names the specific gap - exactly the acceptance criterion in #35's issue body.

Both exports also sort their output (by requirement/case/report/artifact id) so the JSON is byte-stable regardless of the order the caller assembled the program in - the canonical-JSON writer already sorts object keys, but array order is the caller's, and this is meant to be a diffable, committable evidence artifact.

## Test plan
- [x] 16 new tests; 30 total in `evidence_tests` cumulative with the existing suite (governance + evidence-kernel)
- [x] `python3 tools/evidence-lint/mdux_evidence_lint.py` - 0 findings across 10 files
- [x] Full CMake configure (stub Vulkan package) succeeds, `evidence_tests` and `MduXCore` targets both present

**Stacked on `governance/mdux-governance-module` (#95, still open)** rather than `develop` directly, since both branches touch the same `CMakeLists.txt` `FILE_SET` block. Not otherwise dependent on `Justification` - `ComplianceProgram` and the `Justification` type do not reference each other.

Part of #9, issue #35.


camille-martin-paris:
Wave 2 review found and fixed a release-gate defect.

`releaseEvidenceSummary()` previously derived `validation_passed` only from requirement coverage
and `VerificationCase.passed`. A program with an uncontrolled hazard, dangling reference, or
duplicate ID could therefore be reported as a passing release even though
`ComplianceProgram::validate()` rejected it. I pushed `98bcc91`, reusing the complete program
validation result and adding an uncontrolled-hazard regression test. The branch was also
synchronized with #95's schema-parity fix; current head is `0cc1e56`.

Verification:

- `python3 tools/evidence-lint/mdux_evidence_lint.py` — 0 findings
- `git diff --check` — pass
- Local compilation is unavailable: no CMake, GCC 13.3 versus required GCC 15+, and no Vulkan
  headers. No GitHub checks are currently reported for this stacked branch.

Coverage note: #96 substantially implements issue #35, but it does not cure #34's remaining model
gaps: the required `AuditEvent`, structured failure list, `Requirement` source/verification fields,
and `ProblemReport.affectsRisk` are still absent across #95/#96. Impact classification:
potentially safety-relevant because this changes a release-validation result; maintainer/domain
review remains required.

Issue #36, opened 2026-07-26
S3 — docs/governance/soup-register.toml


ambroise-leclerc:
The structured third-party dependency register required by IEC 62304 §5.3.4 and §8. TrustSC's lives at `docs/governance/soup-register.toml` and every crate reachable from `tools/` or `adapters/` must have an entry.

**Per entry:** `component_id`, supplier, version, license, usage path, architectural confinement (which trust zone reaches it), risk controls, and known-anomaly tracking (§8.3.2).

**MduX's actual SOUP today** — the list is short, which is the point:

| Component | Zone | Note |
|---|---|---|
| Vulkan SDK / loader | adapter | `find_package(Vulkan REQUIRED)`; the only runtime native dependency |
| GLFW 3.4 | examples only | via CPM `glfw/glfw@3.4`; never linked into `MduX` itself |
| CMake 4.0+ | build | tool, not shipped |
| MSVC 19.40+ / GCC 15+ / Clang 20+ | build | tool qualification is the manufacturer's, not MduX's |

The trust-zone split from E5/S7 is what keeps this list short and is worth recording as the risk control: `MduXCore` reaches none of these.

As bakers land (E6-E12), each new host-tool dependency needs an entry — and the zero-SOUP scoping decisions (hand-parsed TrueType, hand-parsed safetensors, hand-walked SPIR-V, QOI instead of PNG) exist precisely to keep this file from growing. Say so in the register's header.

Part of #9.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The governance/traceability API, SOUP register, SDF templates, regulatory records, schemas, and tests are present; post-merge CI run 30716625565 passed all seven jobs.

PR #97, opened 2026-07-28
S3 — docs/governance/soup-register.toml

ambroise-leclerc:
## Summary
- Adds `docs/governance/soup-register.toml`, the structured third-party dependency register IEC 62304 §5.3.4 and §8 require, matching the shape TrustSC's own `docs/governance/soup-register.toml` already uses.
- Per-entry fields match #36's issue body: `component_id`, `supplier`, `version`, `license`, `usage`/`integration_path`, architectural confinement (`trust_zone` + `boundary_rationale`), `risk_controls`, and `known_anomaly_tracking` (§8.3.2) — the last one points at `mdux.governance.compliance`'s `ProblemReport` (#35) as where a defect against that component would actually be recorded, rather than inventing a second tracking mechanism.
- Four entries, exactly the table in #36's issue body: Vulkan SDK/loader (adapter zone, the only `runtime_deployment = true` entry), GLFW 3.4 (examples zone only — never linked into MduX itself), CMake 4.0+, and the MSVC/GCC/Clang C++23 toolchain floor (both build zone — tool qualification is the manufacturer's per IEC 62304 §8.1.2, not this register's job).
- The header's `scoping_policy` field records why this list stays short, per the issue's closing note: the trust-zone split (ADR-004, mechanically enforced by `mdux_verify_trust_zones()`) confines every entry to adapter/examples/build and `MduXCore` reaches none of them, and the zero-SOUP scoping decisions already made (hand-written SHA-256 and canonical JSON, issue #12) or planned (hand-parsed TrueType/safetensors/SPIR-V, QOI instead of PNG — issues #13/#14/#17/#18) are why the file is expected to stay short as bakers land in E6-E12, not grow with each one.

## Test plan
- [x] Parses cleanly with Python's `tomllib`
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings across 33 files

Branched from `develop` directly — docs-only, no CMake or C++ file overlap with the still-open `governance/mdux-governance-module` (#95) or `governance/traceability-matrix` (#96) branches.

Part of #9, issue #36.

### DIFF
--- a/docs/governance/soup-register.toml
+++ b/docs/governance/soup-register.toml
@@ -41,7 +41,7 @@ runtime_deployment = false
 support_model = "Community-maintained library; no medical-device qualification or vendor support agreement."
 boundary_rationale = "cmake/MduXTrustZones.cmake's banned-dependency list matches ^glfw$ and ^glfw3$ by name against every governed target; GLFW is linked only into the example executables (MedicalUiExample, VulkanSCTriangleExample)."
 risk_controls = [
-    "Trust-zone enforcement: mdux_verify_trust_zones() rejects glfw/glfw3 as a link dependency of MduXCore or MduX.",
+    "Trust-zone enforcement: mdux_verify_trust_zones() rejects glfw/glfw3 as a link dependency of MduXCore. MduX is the adapter target and is not governed by that check; its no-windowing policy is a separately reviewed target-definition constraint, not a property this check proves.",
     "GIT_TAG 3.4 pins an exact tagged release rather than a branch, so the fallback CPM path cannot silently pick up an unreviewed newer version.",
 ]
 known_anomaly_tracking = "A defect found in this dependency is recorded as a ProblemReport (mdux.governance.compliance, issue #35) with this component_id in its description; none open as of this entry."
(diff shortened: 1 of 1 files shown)


camille-martin-paris:
Wave 2 review found one overstatement in the SOUP register and fixed it in `fae8335`.

The original GLFW risk-control entry said `mdux_verify_trust_zones()` rejects GLFW on both
`MduXCore` and `MduX`. The CMake function checks only targets declared governed; `MduXCore` is
governed, while `MduX` is the adapter target. The register now states that exact boundary and
does not imply the checker proves the adapter's separate no-windowing policy.

Verification:

- TOML parses with Python `tomllib`
- `mdux-docs-lint` — 33 files, 0 findings
- `git diff --check` — pass
- GCC 16 and all documentation/compliance/lint/security checks pass
- GCC 15 and MSVC remain red on the repository's existing evidence JSON/compiler/Unicode failures

With that correction, this PR covers issue #36's requested register shape and four scoped entries.
Impact classification: regulatory/audit documentation relevant; no runtime behavior changed.

PR #98, opened 2026-07-28
S4 — software_development_file/templates/

ambroise-leclerc:
## Summary
- Adds the blank, manufacturer-fillable `software_development_file/templates/` tree issue #37 describes: `IEC_62304/{SAD,SDD,SOUP}.md`, `IEC_62366/Usability_Engineering_File.md`, `IEC_81001/Cybersecurity_SAD.md`, `ISO_13485/README.md`, `ISO_14971/Risk_Management_File.md`, plus a top-level `README.md` explaining the `templates/`/`regulatory/` split.
- Every section header and citation blockquote is checked against the real clause structure already committed on the still-open `regulatory/*` branches (issues #26-33), not guessed or copied from TrustSC's Rust-specific equivalent. In particular, `Cybersecurity_SAD.md` cites IEC 81001-5-1 practice categories by **name**, not an invented numbered sub-clause — matching this project's own confidence caveat that clause-numbering is verified only through §4 for that standard.
- **The rule repeated in every document, per the issue's closing note:** a filled document summarizes, and never duplicates, a machine register. A SOUP document points at `soup-register.toml` (#36); a risk file points at the `Hazard`/`Requirement`/`VerificationCase`/`ProblemReport` records a `ComplianceProgram` (#35) assembles. Restating either as a hand-copied table is how a technical file goes stale.

## Two things that needed care
- **Cross-references to not-yet-merged sibling files** (`docs/iec62304/*`, `docs/iso14971/*`, `docs/iec62366/*`, `docs/iec81001/*`, `docs/governance/soup-register.toml`, `docs/governance/schemas/justification.schema.json`) are written as plain backticked text, not markdown links — a link to a path that doesn't exist on `develop` yet would be dead. `docs/governance/citation-convention.md`, which already exists on `develop`, keeps its real link. Verified programmatically: every relative link under `software_development_file/` resolves to a real file.
- **Every example `Justification` is fenced as `jsonc`, not `json`** — `mdux-docs-lint` only schema-checks a fenced `json` block containing `justification_id`, and a placeholder `"JUS-NNN"` would fail that check as if it were real. `docs/governance/citation-convention.md` already established this convention; these templates are its second user.

## Also in this PR
[...]

Issue #37, opened 2026-07-26
S4 — software_development_file/templates/


ambroise-leclerc:
Blank, manufacturer-fillable documents — the tree TrustSC gives a manufacturer building on it. MduX has nothing equivalent.

```
software_development_file/
  README.md
  templates/
    IEC_62304/{SAD.md, SDD.md, SOUP.md}
    IEC_62366/Usability_Engineering_File.md
    IEC_81001/Cybersecurity_SAD.md
    ISO_13485/README.md
    ISO_14971/Risk_Management_File.md
```

Each template is structured against the E2 corpus and cites into it with the E2/S1 citation keys, so a manufacturer filling one in is guided to the clause it discharges.

**The rule that keeps these documents honest:** filled documents **summarize, never duplicate**, machine registers. A SOUP document points at `docs/governance/soup-register.toml`; it does not restate its rows. A risk file points at the `Hazard` records; it does not copy them. Duplication is how a technical file goes stale.

`risk-assessment-templates.md` (935 L, repository root) should be reviewed here — either folded in or deleted, per E2/S4.

Part of #9.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The governance/traceability API, SOUP register, SDF templates, regulatory records, schemas, and tests are present; post-merge CI run 30716625565 passed all seven jobs.

PR #99, opened 2026-07-28
S5 — software_development_file/regulatory/

ambroise-leclerc:
## Summary
- Fills in the `software_development_file/templates/` (#37) tree for MduX itself, mirroring TrustSC's `templates/`/`regulatory/` split: real architecture description, a real SOUP summary derived from `docs/governance/soup-register.toml` (#36), and real citations into the `docs/<standard>/` corpus and the seven `Accepted` ADRs.
- Every claim was checked against the actual tree, not assumed from TrustSC's Rust equivalent or from this project's own `README.md`.

## The honesty checks that mattered most
- **The trust-zone segregation is stated precisely as what it is.** `mdux_verify_trust_zones()` is a link-graph name check (`Vulkan::`/`glfw`/`glfw3`), not a memory-safety guarantee — C++ has no `#![forbid(unsafe_code)]` equivalent, and `SAD.md` §4 / `Cybersecurity_SAD.md` §3 say so explicitly rather than implying more rigor than exists.
- **No `ComplianceProgram` (#35) has actually been populated for MduX itself.** Only the types and their 30 unit tests (synthetic fixture IDs, not real data) exist. `SDD.md`, `Risk_Management_File.md`, and `ISO_13485/README.md` say this directly rather than fabricating a worked example the way TrustSC's own `examples/class_c_monitor` citation does — there is no MduX equivalent to cite.
- **Root `README.md`'s "Implementation Status" table overclaims are named, not repeated.** It marks several of these items "Completed" and references a `mdux::MedicalDeviceContext` / `mdux/compliance/MedicalDevice.cppm` API — verified absent from the tree (`grep` across `include/`/`src/` finds nothing). `ISO_13485/README.md` calls this out directly under its own heading, consistent with `AGENTS.md` § 2's existing acknowledgment.
- MduX's real UI building block is `MedicalUiRenderer` consuming an HTML/CSS file, not a DSL — `Usability_Engineering_File.md` describes that, and clarifies that `MedicalUiContent.validationErrors` checks structural completeness (verified by reading the actual `push_back` call sites in `src/mdux.cpp`), not hazard-related use-scenario coverage, since the two share only an English word.

## A scoping decision worth flagging
Five of the seven embedded `Justification` records are fenced as `jsonc`, not `json`, even though their content is real: their `evidence_refs` name `include/mdux/governance/Compliance.cppm` or `docs/governance/soup-register.toml`, which exist on the still-open `governance/traceability-matrix` (#96) and `governance/soup-register` (#97) branches but not yet on `develop` — `mdux-docs-lint`'s `evidence_refs` existence check would otherwise fail here. `software_development_file/README.md` names exactly which blocks and why, so they can be re-fenced as `json` once those PRs merge. I verified this isn't a shortcut around a real problem: temporarily merging both branches locally, confirming all seven blocks validate cleanly against the real schema and real paths, then reverting so this PR's diff stays scoped to its own story rather than re-showing #96/#97's changes.

## Test plan
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 0 findings across 48 files
- [x] Every relative markdown link under `software_development_file/` resolves to a real file on this branch (checked programmatically)
- [x] All seven `Justification` blocks verified to validate against the real schema/paths once #96 and #97 are present (tested via a temporary local merge, then reverted)

Stacked on `software-development-file/templates` (#98), matching the SDF README's own description of `regulatory/` as the immediately-following PR.

Part of #9, issue #38.


camille-martin-paris:
Wave 2 review found a substantive claim mismatch and corrected it.

The risk file originally said that requiring any `VerificationCase` meant a risk control had been
verified. `ComplianceProgram::validate()` checks only that a case exists; it does not require
`VerificationCase.passed`, and it does not require non-empty `evidenceRefs`. Commit `d88644f`
now distinguishes a defined verification activity from its result and points to
`releaseEvidenceSummary()` for the passed flag. Commit `7efa905` updates the documented governance
test count to 31 after #96's new regression test.

Verification:

- `mdux-docs-lint` — 48 files, 0 findings
- `git diff --check` — pass
- No GitHub checks are currently reported for this stacked branch.

Coverage blocker: five real Justifications remain fenced as `jsonc` because #96/#97 are not in this
branch ancestry. They are consequently not schema/path checked as real records. This PR should be
restacked on the governance and SOUP branches—or updated after those merge—and the blocks changed
to `json` before issue #38 is treated as complete. It also honestly records that no real
`ComplianceProgram` is populated for MduX; this is a filled gap analysis, not a completed risk
file. Impact classification: regulatory/traceability documentation relevant.

Issue #28, opened 2026-07-26
S3 — Rewrite docs/iso13485/ to the same shape


ambroise-leclerc:
Same treatment as E2/S2, against ISO 13485:2016's real clause structure. The four existing modules (`01-foundations-and-qms.md` through `04-measurement-analysis-improvement.md`, ~3,350 lines total) are restructured and rewritten as original prose with real clause keys.

Fold in `docs/iso13485/ai-automation-schemas/*.json` and `docs/iso13485/code-examples/**/*.py` rather than leaving them as a parallel tree.

Retire the monolithic `docs/MduX_ISO-13485-Quality-Management-Framework.md` (1,423 L) in favour of the modular corpus, per E4/S6's archiving decision.

**Scope note worth stating up front:** ISO 13485 is a QMS standard. MduX cannot provide an operating QMS — no management review, no CAPA, no supplier qualification. This corpus explains the clauses and points at the few places MduX genuinely contributes (design controls, configuration management, records). E3/S6's scope-limits document says the rest.

Part of #8.

ambroise-leclerc:
Completed in Wave 2 and verified on develop at 52735c80e45a6bfd8af4d47b4b5b589a8d341be4. The regulatory corpus, schemas/indexes, citation governance, scope limits, and documentation lint evidence are present; post-merge CI run 30716625565 passed all seven jobs.

PR #104, opened 2026-07-31
fix: unbreak develop after the Wave 2 merges

ambroise-leclerc:
`develop` does not configure, and CI has been red on every push since #97.

### The build wiring lost three files

#96 branched before #95's review rewrite renamed `Justification.cppm` → `Governance.cppm` and split the implementation across `Governance.cpp` and `Program.cpp`. Merging develop into #96 (`d4626c0`) resolved the `CMakeLists.txt` conflict by taking #96's side wholesale:

```
-            include/mdux/governance/Governance.cppm
-        src/governance/Governance.cpp
-        src/governance/Program.cpp
+            include/mdux/governance/Justification.cppm     ← deleted file
```

`tests/CMakeLists.txt` lost `ComplianceProgramTests.cpp` the same way. The sources were never deleted — only their wiring — so restoring the four entries is the whole fix.

### The IEC 62304 index was stale, and stale in a way that mattered

Its §4.3 row still read **"MduX's own components are declared Class A"** — the exact overclaim the review on #88 corrected — while the module it indexes now says the opposite. Regenerating replaces it. The `--check` step added on #93 was correctly failing on this.

### Dangling references to documents Wave 2 deleted

`README.md`'s quickstart told a reader to `cat` two files that no longer exist; `AGENTS.md` and the `mdux-regulated-change` skill pointed at the same two. Deleting a document and leaving pointers to it is not a finished deletion. They now point at the corpus that replaced them, and `AGENTS.md` gains a pointer to `superseded-documents.md` so "where did it go" has an answer.

Also corrects `AGENTS.md`'s claim that only one Windows-only CMake preset exists — there have been five since issue #45.

### Deliberately not fixed here

`docs/MduX_IEC-62304-Software-Lifecycle-Framework.md` (3,197 L) is still present and still superseded by `docs/iec62304/`. Issue #27 did not ask for its retirement the way #28 and #29 did for their monoliths, and epic #10's "archive the point-in-time docs" is where that decision belongs.

**The duplicate governance model is also not fixed here** — see the separate write-up. `mdux.governance` (#34) and `mdux.governance.compliance` (#35) both define `Requirement`, `Hazard`, `VerificationCase`, `ProblemReport` and `ComplianceProgram` in `namespace mdux::governance`, with different field names. They compile today only because no translation unit imports both. That is a design defect needing its own change, not a build fix.

### Verification

- 178/179 ctest pass. The one failure is `InstallTreeConsumer`, which fails on this machine for an unrelated CMake-version reason and passes in CI.
- 65 docs-lint tests pass; `mdux-docs-lint` 71 files, 0 findings.
- `generate_ai_reference.py --check` clean on all five corpora.
- `check_schema_type_drift.py` clean.

### Impact classification

Not safety-relevant at runtime. Restores a build that does not configure, and removes a superseded safety-classification claim from a generated index.


camille-martin-paris:
Review completed; I pushed the fixes directly to this branch.

Findings addressed:

- Removed the accidental whole-file LF normalization in `CMakeLists.txt` and `README.md`, reducing the effective PR diff from roughly 1,500 noisy changed lines to the intended documentation edits.
- Merged current `develop` and resolved the CMake conflict. The build-wiring repair is already present on `develop`, so the remaining PR diff is focused on the stale/dangling documentation references.
- Preserved #103's removal of tracked `.vscode/settings.json`.

Verification performed on the updated head:

- 65 `tools/docs-lint` unit tests pass.
- `mdux-docs-lint`: 71 files, 0 findings.
- All five generated AI reference corpora are current.
- Schema/type drift check passes.
- `git diff --check origin/develop...HEAD` passes.

Impact classification: not safety-relevant at runtime; the remaining changes correct contributor and regulatory-document navigation only. I could not run the C++ build locally because this environment has no CMake or Ninja and provides GCC 13 rather than the required GCC 15+, so GitHub CI is the build verification source.

PR #105, opened 2026-07-31
refactor(governance): reconcile the compliance exports onto the #34 model

ambroise-leclerc:
Stacked on #104, which unbreaks the build. Closes the last Wave 2 gap.

`mdux.governance.compliance` (#35) was written against a version of `mdux.governance` that had only `Justification`, and was never revisited after #34's review landed the full model. Both then declared `Requirement`, `Hazard`, `VerificationCase`, `ProblemReport` and `ComplianceProgram` in `namespace mdux::governance`:

| | #34 `mdux.governance` | #35 `mdux.governance.compliance` |
|---|---|---|
| Requirement | `id`, `title`, `sourceClause`, `verificationIntent` | `requirementId`, `description` |
| VerificationCase | `id`, `requirementId`, `method`, `evidenceRefs` | `caseId`, `requirementId`, `evidenceRefs`, `passed` |
| ProblemReport | `id`, `description`, `closed`, `affectsRisk` | `reportId`, `description`, `open` |
| `validate()` | structured failure list | single error |

They compiled only because no translation unit imported both.

Issue #35 says the exports are "pure functions over a `ComplianceProgram`" — meaning #34's. They now are: this module imports the model instead of restating it, and its duplicate types and `ComplianceError` are deleted.

### The matrix can now trace to a clause

The local `Requirement` carried only an id and a description, so `traceabilityMatrix()` could not name the clause a requirement came from — the one thing a *regulatory* traceability matrix exists to show. Every row now carries `source_clause` and `verification_intent`.

### The export agrees with the schemas again

It emitted `case_id` and `report_id` while `verification-case.schema.json` declares `id`, so a row and the record it came from could not be read side by side. Member names now match the fields they come from, and `check_schema_type_drift.py` covers them.

### Two #35 capabilities preserved, not dropped

- **`VerificationCase::passed`** is added to the canonical type and to `verification-case.schema.json` — required on both sides with no default. A case that has not stated an outcome has not passed, and a release view that assumed otherwise would report an unrun verification as evidence.
- **`generatedArtifacts`** becomes a parameter of `releaseEvidenceSummary()` rather than a field. A digest of a built file is evidence *about a build*, not an authored governance record; keeping it out preserves `ComplianceProgram::parse()`'s strict member set and matches ADR-007's rule that assembling evidence is the caller's job.

The summary also gains `validation_failures`, the structured list #34's `validate()` returns — a release review told only `validation_passed: false` has to re-run the gate to learn anything.

### Tests

`ComplianceTests.cpp` is rewritten against the shared model and now covers issue #35's stated acceptance criterion directly: a program with an unverified requirement *and* an uncontrolled hazard fails `validate()` while the matrix still names both gaps as rows. Also added: assembly order cannot change the exported bytes, and the matrix's member names match the schemas.

### Verification

- 176/177 ctest pass; 109/109 `evidence_tests`. The one failure is `InstallTreeConsumer`, which fails on this machine for an unrelated CMake-version reason and passes in CI.
- `check_schema_type_drift.py` clean — including the new `passed` binding.
- `mdux-docs-lint` 71 files, 0 findings; 65 docs-lint tests pass.

### Impact classification

Potentially safety-relevant: this changes what a release gate's evidence export contains and what `validate()` is called on. No rendering or device runtime behaviour changes. The export is strictly more informative than before — it gains the clause, the failure list, and the `affects_risk` flag on open problem reports.


camille-martin-paris:
Review completed; I pushed a release-evidence fix and synchronized this stacked branch with the updated #104 base.

Finding addressed:

- `validation_passed` became false when a `VerificationCase::passed` value was false, but `validation_failures` remained empty because case outcomes are intentionally outside `ComplianceProgram::validate()`'s structural checks. The resulting summary claimed to carry every gate failure while omitting the only reason it failed. The export now includes a deterministic `failed_verification_cases` array, and the compliance tests cover both the clean and failed-case forms.

Verification performed on the updated head:

- 65 `tools/docs-lint` unit tests pass.
- `mdux-docs-lint`: 71 files, 0 findings.
- All five generated AI reference corpora are current.
- Schema/type drift check passes.
- `git diff --check` against the updated #104 base passes.

Impact classification: safety-relevant because this changes the release-evidence API and makes a failed verification outcome explicitly visible to release review. The affected interface is `mdux.governance.compliance`; verification evidence is in `tests/governance/ComplianceTests.cpp`. Maintainer/domain-expert review remains required before merge. I could not run the C++ build locally because this environment has no CMake or Ninja and provides GCC 13 rather than the required GCC 15+, so GitHub CI is the build verification source.

ambroise-leclerc:
Completed and incorporated into develop. The exact reviewed head 8ae6fe672ab904c791d3782644c06d0e3ad4aa68 was merged in develop commit 52735c80e45a6bfd8af4d47b4b5b589a8d341be4 after #104; post-merge CI run 30716625565 passed all seven jobs. Closing this stacked PR because its changes are already present.

Issue #124, opened 2026-08-01
S6 — mdux.render.vulkan fixed-budget renderer


ambroise-leclerc:
Part of #13. Depends on the emitter and `mdux.draw` children.

Implement `mdux.render.vulkan` as the adapter-zone fixed-budget renderer. Allocate vertex/index resources once at initialization, construct the real graphics pipeline from caller-owned device/render-pass context, and record solid-rectangle draw commands without growth or runtime shader I/O.

Acceptance: invalid contexts and budget overflow fail explicitly; resource ownership/destruction is deterministic; command-recording tests and active platform builds pass.

Issue #121, opened 2026-08-01
S3 — Shader C++ emitter with module and header outputs


ambroise-leclerc:
Part of #13. Depends on the committed shader package child.

Add the generated-code emitter backend for module and header consumers. Emit namespace-scope C arrays with `std::span` views, not `std::array`, to avoid the documented GCC modules ICE. Generated C++ stays in the build tree and is mechanically derived from committed canonical artifacts; reviewers inspect JSON and binary digests, not generated source.

Acceptance: module and header outputs describe identical bytes/contracts, both compile in CI, and no normal build changes tracked files.

Issue #120, opened 2026-08-01
S2 — Author and bake the MduX UI shaders


ambroise-leclerc:
Part of #13. Depends on the shader schema/baker child.

Author the minimal MduX UI vertex/fragment shaders and a recipe, then commit their evidence-pipeline outputs. Cover solid, R8 coverage, and sampled-RGBA modes with a single fixed pipeline contract. Record fully resolved options and input/output digests; no runtime shader file loading.

Acceptance: `mdux-bake-update` produces the reviewed artifacts, `ctest -L evidence` is byte-identical on active toolchains, and shader contract tests reject incompatible resources.

Issue #118, opened 2026-08-01
S3 — Stable JSON diagnostic envelope across all tools


ambroise-leclerc:
Part of #19. Establish the diagnostic contract before shader, `.medui`, and ML tools multiply it.

## Interface

Use one JSON envelope everywhere: `{tool, file, line, column, code, severity, message, fixHint}`. `line` and `column` are 1-based; `0` means the diagnostic applies to the whole file or has no precise position.

## Work

- Add `column` to `mdux::tools::cli::Diagnostic` without changing the meaning of existing fields.
- Publish the envelope as a versioned JSON Schema.
- Align `MduXTest`, `mdux-docs-lint`, and `mdux-evidence-lint` output or adapters with the same field names and severity vocabulary.
- Pin literal JSON-output tests and rejection tests for unknown schema fields where strict consumers are used.
- Require all later bakers and `mdux-medui-check` to use the shared renderer rather than inventing envelopes.

## Acceptance

- Current tools emit schema-valid JSON.
- Existing text output remains human-readable and exit-status behavior is unchanged.

PR #129, opened 2026-08-02
S1 — mdux.shader.schema and mdux-shaderbake

### LINKED ISSUE #119: S1 — mdux.shader.schema and mdux-shaderbake
Part of #13. Depends on the completed evidence kernel (#12).

Implement the governed shader package schema and host-only `mdux-shaderbake`. The baker hand-walks SPIR-V and must not link glslang. It validates module structure, entry points, descriptor/push-constant contracts, and resolved options, then emits canonical `package.json`, `report.json`, and a binary sidecar through the existing evidence APIs and shared CLI diagnostics.

Acceptance: malformed/truncated/unsupported SPIR-V fixtures fail with stable codes; bake/verify share one code path; MSVC and GCC evidence checks reproduce byte-identical artifacts; normal builds do not write the source tree.

ambroise-leclerc:
Closes #119. First story of #13, and the repository's first baker.

`mdux_bake_artifact()` has had no call sites since #53 landed it, so `ctest -L evidence` currently selects zero tests. This PR builds the tool; #120 commits the first artifact and makes that CI gate start doing work.

## `mdux.shader.schema` — governed, canonical

`include/mdux/shader/Schema.cppm` joins `MduXCore`, with `validate`/`toJson`/`write`/`parse` mirroring `BakeReport`'s shape. Canonical in the sense epic #13 asks for: the baker, the #121 emitter and the #124 renderer import it rather than restating it.

**It names no Vulkan type, deliberately.** `DescriptorKind` and `Stage` are MduX's own vocabulary with their own wire spellings, translated to `VkDescriptorType`/`VkShaderStageFlagBits` in the adapter zone. A governed module naming a `Vk` type could not link into `MduXCore` at all — `mdux_verify_trust_zones()` fails the configure step — and the package format would become unreadable to any tool without the Vulkan headers.

**One binary sidecar**, addressed by `byteOffset`/`byteLength`, rather than a file per module. A package then has exactly one payload to digest in one read, and #121 emits one C array instead of one per stage. Per-module digests are still recorded, so a corrupted range stays attributable.

`validate()` rejects overlapping ranges, ranges past the sidecar, and lengths that are not a multiple of 4. A SPIR-V module is a sequence of 32-bit words, so an unaligned length cannot be one — and the same mistake reaching `vkCreateShaderModule` surfaces as a device loss rather than a diagnostic.

## `mdux-shaderbake` — host-only

**Hand-written SPIR-V reflection; no glslang, no SPIRV-Reflect.** Same reasoning as the TOML subset: a dependency here would be SOUP in the tool that produces every committed artifact, and its version would become part of the byte-identity contract. Two passes over the instruction stream (the first indexes definitions, because SPIR-V permits forward references), every read bounds-checked against the module length, and anything unrecognised refused with a specific code rather than guessed.

Three decisions worth a reviewer's attention:

- **A byte-swapped module is rejected, not swapped.** Accepting it would mean the same source could bake to two different committed artifacts depending on the endianness of the machine that ran the bake — precisely what the pipeline exists to make impossible.
- **Descriptors are sorted by `(set, binding)`** before reaching the package, so committed bytes do not depend on id allocation order inside whatever compiler produced the SPIR-V.
- **A binding two stages declare *differently* is an error**, not a silent choice. Taking either would produce a pipeline layout matching one shader and not the other. A binding they declare identically merges into one with both stage bits.

`bake` and `verify` share one code path, as ADR-007 requires: `run()` produces every output byte in memory and the caller either writes or compares them. Verify also compares the sidecar rather than trusting the digest the package records for it — a package whose recorded digest matches a file nobody compared proves nothing.

Push-constant block sizes come from the members' `Offset` decorations rather than by summing member sizes, so whatever padding the shader compiler inserted comes out right instead of being re-derived badly.

## A GCC modules finding, recorded for #121 and #124

`BakeOutputs` deliberately does not carry the `ShaderPackage` it rendered — callers parse `packageJson`, which is better for a test anyway since it exercises the reader on the writer's own output.

There is also a compiler reason, documented in the header. **GCC 15 rejects a struct exported from a host-tools module that embeds a governed-module type**, reporting `use of deleted function ~ShaderPackage()` in any plain TU that imports it, because the implicit destructor's exception specification is computed inconsistently across the module boundary. `ShaderPackage` itself is fine in a plain TU importing `mdux.shader.schema` directly — `SchemaTests.cpp` does exactly that and compiles. It is specifically the re-export that breaks. Worth knowing before #121 and #124 design their own interfaces.

## Tests

74 cases across three files, weighted towards rejection — a parser whose failure paths are untested is a parser with untested failure paths.

Fixtures are assembled word by word by `tests/shader/SpirvFixtures.hpp` rather than loaded from committed `.spv` files. That buys three things: a malformed case can be built exactly (a truncated instruction is one word count away from a valid one), the tests need no shader compiler, and no binary enters the tree whose provenance a reviewer must take on trust.

Cases that earn their keep: an instruction claiming a word count of zero (without the check the parser would spin forever), adjacent-but-not-overlapping module ranges (the off-by-one on the overlap test), an entry point name whose length is a multiple of four (the NUL needs a word of its own), and `verify()` failing on a changed byte, a changed length, a missing file, and a corrupted sidecar.

`shader_tests` links `MduX::ShaderBakeLib` rather than spawning the executable, so a failing assertion says which artifact differed at which byte instead of reporting that an exit status was 1.

## Two stale claims corrected

The `evidence-pipeline` skill said no baker and no `cmake/MduXBake.cmake` existed — the latter landed in #53 — and pointed readers at `examples/shaders/compiled/`, which no longer exists. Rewritten to say what is actually in the tree, including that no `generated/` directory exists yet. `AGENTS.md`'s routing row is updated to match. The wider AGENTS.md pass is #65's.

## Verification

```
python3 tools/evidence-lint/mdux_evidence_lint.py     # OK (15 files checked, 0 findings)
python3 tools/docs-lint/mdux_docs_lint.py             # OK (71 files checked, 0 findings)
python3 tools/docs-lint/check_schema_type_drift.py    # OK
python3 -m unittest discover -s tools/docs-lint       # OK
```

C++ compile-checked locally with GCC 15.2: `MduXCore`, `MduXShaderBakeLib`, `mdux-shaderbake` and every `shader_tests` TU build with **zero errors and zero warnings** under the project's `-Werror` profile. **Test execution is CI-only** — linking any executable on macOS fails on `Undefined symbols: "initializer for module std"`, a CMake/GCC-on-macOS gap, not a project one. So the 74 assertions passing is this PR's CI to demonstrate, not mine.

---

- **Predecessor**: none. Branches from `develop` at f7756aa, independently of #128 — the TOML recipe parser reports lines only, so nothing here needs #118's `column`, and there was no reason to create a stack.
- **Shared registries edited**: `CMakeLists.txt` (one module + one source added to the `MduXCore` file set), `tools/CMakeLists.txt` (two new targets, appended), `tests/CMakeLists.txt` (one new target, appended). No existing entry reordered or removed.
- **Local validation**: as above.
- **Post-merge gate**: `develop` CI must be green before #120 opens, since #120 adds the first `mdux_bake_artifact()` call site against this tool.


ambroise-leclerc:
**Correction to the "GCC modules finding" section above.** My characterisation there is wrong, and the follow-up PR corrects it in the source.

The trigger is **not** re-exporting a governed type from a host-tools module. It is **`std::optional<T>` where `T` is, or contains, a class imported from another named module**. `std::expected<T, E>` is unaffected — which is why `ShaderPackage::parse()` returns one and has always worked from plain translation units.

I found this while writing the contract tests for #120: a plain `std::optional<ShaderPackage>` in a TU that imports `mdux.shader.schema` directly fails identically, with no host-tools module involved at all. That rules out my original explanation.

What actually happens in this PR is unchanged and still correct — `BakeOutputs` not carrying a `ShaderPackage` is the right design regardless, since callers parse `packageJson` and thereby exercise the reader on the writer's own output. Only the *reason* recorded in the header comment was wrong.

The practical rule, for #121 and #124: **prefer `mdux::core::Result` over `std::optional` for a type that crosses a module boundary.**

No action needed on this PR; flagging it here so the description does not mislead a reviewer reading it after the fact.

tests/shader/SchemaTests.cpp
```diff
@@ -0,0 +1,411 @@
...
+    return result.error();
+}
+
+}  // namespace
+
+// ---------------------------------------------------------------------------
+// Wire encoding
+// ---------------------------------------------------------------------------
+
+TEST_CASE("Stage and DescriptorKind wire spellings are stable", "evidence-unit") {
```
camille-martin-paris, line 59: Please convert the new shader suites to Catch2 BDD style: `SCENARIO` plus meaningful `GIVEN`/`WHEN`/`THEN` sections, not `TEST_CASE`. This applies to `SchemaTests.cpp`, `ShaderBakeTests.cpp`, and `SpirvTests.cpp`; ADR-002 selected BDD syntax specifically for traceability.

tools/shader/ShaderBakeMain.cpp
```diff
@@ -0,0 +1,99 @@
+/**
```
camille-martin-paris, line 1: Please rename this source file to UpperCamelCase (for example `ShaderBakeMain.cpp`) and update the CMake source list. `CONTRIBUTING.md` requires UpperCamelCase for `.cpp` and `.hpp` filenames.

include/mdux/shader/Schema.cppm
```diff
@@ -0,0 +1,214 @@
+/**
+ * @file Schema.cppm
```
camille-martin-paris: Project documentation rules say file-level Doxygen uses `@brief` only—no `@file` or `@author` tags. Please remove `@file` from the new C++ files in this PR.

include/mdux/shader/Schema.cppm
```diff
@@ -0,0 +1,214 @@
...
+import std;
+import mdux.core.result;
+import mdux.evidence.digest;
+import mdux.evidence.json;
+import mdux.evidence.report;
+
+export namespace mdux::shader {
+
+/// The `<kind>` component of `generated/<kind>/<id>/`, and the value of a package's `kind` member.
+inline constexpr std::string_view kKind = "shader";
```
camille-martin-paris: Please remove the `k` prefixes from constants throughout this PR. `CONTRIBUTING.md` explicitly applies lowerCamelCase to `const` and `constexpr` variables, so this should be `kind`, with names such as `stageWireValues`, `vertexBit`, `toolName`, `recipeUnparsed`, and `magic` updated consistently in the schema, baker, SPIR-V parser, and fixtures.

ambroise-leclerc:

## On the BDD request

tools/shader/ShaderBakeMain.cpp
ambroise-leclerc: Renamed. `tools/shader/` is now `Spirv.cpp`, `ShaderBake.cpp`, `ShaderBakeMain.cpp`, `Emit.cpp`, `ShaderEmitMain.cpp`, with `tools/CMakeLists.txt` updated to match.

include/mdux/shader/Schema.cppm
ambroise-leclerc: Removed. No `@file` or `@author` tag remains in any C++ file this stack adds or edits.

include/mdux/shader/Schema.cppm
ambroise-leclerc: Done throughout. Three could not be a plain prefix strip: `packageKind` (bare `kind` raises MSVC C4459 against the `DescriptorKind kind` parameters), `spirvMagic` (a local `magic` in `reflect()`), and `floatTypeId`/`structTypeId` (`float` and `struct` are keywords). One `k` name survives and is deliberate: `evidence::kSchemaVersion` is pre-existing wave-2 code at `include/mdux/evidence/Report.cppm:65`, referenced but not introduced here — renaming it belongs to its own PR.

ambroise-leclerc (commented):

Already handled
- `@file`/\`@author\` removed, `k` prefixes dropped (with three documented non-trivial renames: `spirvMagic\`, `floatTypeId\`/\`structTypeId\`, `maxIndexableVertices\`), file rename to `ShaderBakeMain.cpp\`. BDD request declined — agreed (Catch2 not in repo, ADR-002 `Proposed (never implemented)\`).

Four Copilot items to address

1. `tools/shader/Spirv.cpp:206\` and \`:328\` — `typeSize()\` computes \`width/8\` with no check that the bit width is non-zero and byte-aligned. A width of 0 or 1 yields a 0-byte size and confusing downstream layout maths rather than rejection. Reject the type as unsupported before the division.

2. `include/mdux/shader/Schema.cppm:174\` — the \`ShaderPackage\` header comment is duplicated verbatim. Looks accidental; remove one.

3. `tools/shader/ShaderBake.cppm:54\` — comment says `One [[module]] entry\` but the recipe format is a \`[modules]\` table with parallel arrays, not TOML array-of-tables. Misleads readers of the public interface. Rewrite to match what the reader actually parses.

4. The \`readFile()\` \`tellg()\` pattern recurs here and on #130/#131. Not blocking on its own but worth a shared helper that guards \`tellg()\` returning -1 before the \`static_cast<size_t>\`, so the three call sites don't each carry the same latent allocation bomb.

tests/shader/SchemaTests.cpp
ambroise-leclerc: Leaving this open rather than actioning it, because the framework it asks for is not available here and its replacement is blocked upstream.

 `tests/shader/SchemaTests.cpp` and the other new shader suites are in that first batch.

tools/shader/Spirv.cpp
```diff
@@ -0,0 +1,611 @@
...
+    const std::span<const std::uint32_t> operands = type.operands;
+
+    switch (type.opcode) {
+    case OpTypeInt:
+    case OpTypeFloat:
+        // operands: [resultId][width][signedness?]
+        if (operands.size() < 2) {
+            return err(ParseError::UnsupportedType);
+        }
+        return operands[1] / 8;
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `fde8e17`. Widths that are zero or not a multiple of 8 are now `UnsupportedType`, with a test for each. Worth adding that this is reachable from a legal module rather than only a malformed one — SPIR-V permits 1-bit integers, and wider odd widths under capabilities this reflector does not implement.

On "also appears on line 328": there is only one `/ 8` in the file, so I have read that as a diff anchor rather than a second site. If you meant something else there, point me at it and I will take another look.

tools/shader/ShaderBake.cppm
```diff
@@ -0,0 +1,137 @@
...
+import mdux.evidence.report;
+import mdux.shader.schema;
+import mdux.tools.cli;
+
+export namespace mdux::tools::shaderbake {
+
+/// The tool name that appears in every diagnostic and in `report.json`.
+inline constexpr std::string_view toolName = "mdux-shaderbake";
+
+/// One `[[module]]` entry, in recipe order.
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `fde8e17`. It is a `[modules]` table with parallel `ids`/`sources` arrays, and the corrected comment now also says *why* — `tools/common/Toml.cppm:13` lists arrays of tables as deliberately unsupported, which is the reason the format looks unusual in the first place.

Issue #123, opened 2026-08-01
S5 — mdux.draw fixed-budget draw types


ambroise-leclerc:
Part of #13. May begin after the canonical shader schema lands, but must rebase before integration if shared CMake/module registries changed.

Add governed `mdux.draw` types: the 24-byte `UiVertex`, bounded `DrawList`, `DrawCommand`, and compiler-computed `DrawBudget`. Public types use spans/PODs, contain no Vulkan handles, allocate no memory while building a frame, and fail explicitly when a budget is exceeded.

Acceptance: layout/static assertions, deterministic ordering tests, exact-budget and overflow tests, and governed-zone checks pass.

Issue #117, opened 2026-08-01
S11 — Stack-safe PR integration and post-merge policy


ambroise-leclerc:
Part of #11 and the Wave 2 RETEX.

Wave 2 stacked PRs lost CMake/module/test wiring during conflict resolution and allowed two incompatible governance models to coexist until integration. Encode a stack-safe delivery policy in repository guidance and the PR template.

## Required policy

- A PR declares its predecessor/base dependency and every shared registry it edits (`CMakeLists.txt`, module file sets, test source lists, schemas, generated indexes).
- A stacked successor cannot merge until its predecessor is merged, the successor is rebased onto current `develop`, conflicts are resolved as an explicit union, and its final diff against `develop` is reviewed again.
- Canonical shared types/schemas land before exports and consumers; a successor imports them rather than restating them.
- Required PR CI must be green, then the resulting `develop` workflow must be green before the next dependent PR merges.
- Emergency local/API-bypass merges require an issue/PR comment naming the exact incorporated head SHA and post-merge CI run.

## Repository changes

- Update `CONTRIBUTING.md` and `AGENTS.md` with the policy.
- Add a pull-request template checklist containing dependency, shared-registry, rebase SHA, local validation, and post-merge-gate fields.

## Acceptance

- The template is usable for independent and stacked PRs without ambiguous fields.
- Documentation lint passes.
- A maintainer review confirms the Wave 2 #104/#105 failure mode is explicitly covered.

Issue #17, opened 2026-07-26
Epic: Content components


ambroise-leclerc:
Fill out the `.medui` component dictionary beyond the solid-rect and text primitives.

## Current status (2026-08-28)

**Unblocked and open.** #15 closed on 23 August at 12/12. #242 has since drawn a `Label` and #244 has
bound a screen to the text packages it was compiled against, so the text path these components build
on is complete rather than assumed.

## Children

- [ ] #256 — S1 `mdux-imagebake` (QOI) and the `Image` component
- [ ] #257 — S2 `SignalTrace`: ring-buffer binding and CPU polyline expansion
- [ ] #258 — S3 `NumericDisplay` and `Clock` live binding — **blocked by #219**
- [ ] #259 — S4 `StatusIndicator` with a closed `states` list
- [ ] #260 — S5 `TextInput`, display and caret only
- [ ] #261 — S6 `Button` and `CriticalButton` with `requirement:` binding — touches #219

Unlike #16's, these are largely independent of one another. **#256 is the one to start**: it depends
on nothing else here.

## #219 orders two of these

#219 closes `ClockFormat` and `SystemEvent`, and it is a prerequisite rather than a nicety:

- An open format name **cannot be measured**, only looked up, so #258's `Clock` would be built
  against the product-supplied table #195 needs today and then have it removed.
- A screen that can name any system event can name one nothing implements — worst discovered on the
  press of a critical button, which is #261.

Land #219 before either.

## Two deliberate scope cuts

**QOI, not PNG, in v1.** QOI is ~200 lines, lossless, and public domain — trivially qualifiable as a from-scratch decoder. PNG needs a hand-written DEFLATE inflater: ~400 lines of much subtler code, for a format designers can export out of. Document the PNG→QOI export step instead.

**No IME.** `TextInput` gets display and caret. Input-method editing is a platform concern that does not belong inside a governed renderer, and pulling it in would breach the trust-zone boundary established in E5.

## `SignalTrace` design note

Samples come from a **caller-owned ring buffer** via a binding slot; expansion writes into the pre-sized vertex budget from E7. Each segment becomes a quad with round-cap overdraw rather than mitred joins — cheaper, and visually indistinguishable at 1-3 px stroke widths.

---

*Children promoted to individual issues on 28 August 2026, when #15 closed.*

ambroise-leclerc:
**Note from #18, now closed: S2 and S4 have a waiting consumer.**

[...]

PR #130, opened 2026-08-02
S2 — Author and bake the MduX UI shaders

### LINKED ISSUE #120: S2 — Author and bake the MduX UI shaders

ambroise-leclerc:
Closes #120. Second story of #13. **Stacked on #129 (#119) — merge that first.**

The first committed artifact under `generated/`, and the first `mdux_bake_artifact()` call site. `ctest -L evidence` selected **zero** tests before this; it now selects `evidence.shader.mdux-ui` and byte-compares a real package on every toolchain leg.

## The shaders

One fixed pipeline contract with three fragment paths selected per-vertex — solid, R8 coverage (glyphs, #14), sampled RGBA (images, #17). One pipeline for all three is what makes a fixed-budget renderer possible: a frame is one bind and one draw per budget, with no pipeline switching whose cost depends on what the screen happens to contain.

The 24-byte vertex is the layout `mdux.draw`'s `UiVertex` will declare in #123:

| offset | | format | |
|---|---|---|---|
| 0 | `position` | `R32G32_SFLOAT` | pixels, origin top-left |
| 8 | `uv` | `R32G32_SFLOAT` | atlas coordinates |
| 16 | `color` | `R8G8B8A8_UNORM` | 4 bytes, normalised in the shader |
| 20 | `mode` | `R32_UINT` | selects the fragment path |

Positions arrive in **pixels** and are converted once in the vertex shader from a push constant, so a governed draw list holds no projection maths and no dependency on the surface it will eventually be drawn to.

The atlas is bound even for a wholly solid draw. A descriptor set whose shape changed with the content would put a conditional in the renderer's hot path *and in its budget*, which is the opposite of fixed.

## The `.spv` files are committed source assets, not build outputs

The recipe points at SPIR-V checked into `recipes/shader/mdux-ui/`, compiled once by hand. This is the decision most worth pushing back on if you disagree, so the reasoning in full:

- **glslangValidator's output varies between its own versions.** Compiling GLSL during the build would make every committed package depend on whichever shader compiler the CI image shipped — and the three legs do not ship the same one. Byte-identity would fail for a reason that has nothing to do with MduX.
- It would make a shader compiler a **build requirement for everyone**, which ADR-007 avoids deliberately.
- ADR-007's own diagram is `recipes/<kind>/<id>.toml + assets/… ──[baker]──▶ generated/…`. The `.spv` *is* the asset.

The `.glsl` beside each `.spv` is what a reviewer reads; the `.spv` is what the digests cover. The recipe documents the regeneration command. `.gitignore` gains a `!recipes/**` negation mirroring `!generated/**`, since the blanket `*.spv` rule would otherwise swallow them.

## A cross-platform byte-identity trap, fixed before it bit

`.gitattributes` had **no end-of-line policy**, and GitHub's Windows runners default to `core.autocrlf=true`. The recipe would have been checked out with CRLF on Windows and LF everywhere else — a different SHA-256 in `report.json`, and an `evidence.*` test that **fails on the MSVC leg alone** while every other leg stays green.

`recipes/**` and `generated/**` are now `-text`, so bytes that are hashed are never transformed by git. Worth knowing for every baker that follows.

## Contract tests, because the evidence gate cannot cover this

`evidence.shader.mdux-ui` proves the committed artifact is what the baker produces from the committed sources. It **cannot** prove the artifact is what the renderer needs: re-baking an edited shader updates both sides at once, so the comparison stays green while the pipeline contract silently moves.

`UiPackageTests` is the other half — descriptor set/binding and kind, push-constant size and stage visibility, both stages present — so changing the contract is a deliberate edit here rather than a surprise in #124 or a validation-layer message on a device. It also checks the package's own claims against the bytes beside it, and that each recorded range really is SPIR-V rather than merely correctly digested.

## Verification

Everything below was **run**, not just compiled — I found a way past the macOS link gap that blocked the earlier PRs (CMake builds `std.cc` with `-fmodule-only`, so no object exists; compiling it by hand against the BMI CMake already built produces one, and passing it via `CMAKE_EXE_LINKER_FLAGS` links everything).

```
ctest --test-dir build-mac                     # 256/257 passed
ctest --test-dir build-mac -L evidence         # evidence.shader.mdux-ui .... Passed
```

The one failure is `InstallTreeConsumer`, which runs its own nested CMake configure for a throwaway consumer and so does not inherit that linker workaround — a local-environment artifact, not a finding.

**The evidence gate was checked for teeth**, not just for green. Flipping one byte of the committed sidecar:

```
shaders.spv: first differs at byte 100 - baked 0x2b, committed 0xff (sizes: baked 3032, committed 3032)
  cmake --build <build-dir> --target mdux-bake-update
```

**Source-tree cleanliness** was verified the meaningful way — a full clean build that *did* run the bake, then `git status --porcelain` empty. (My first attempt at this was clean for the wrong reason: executables had not linked, so the bake never ran.)

Python lints: `mdux-evidence-lint` OK (15 files), `mdux-docs-lint` OK (71 files).

## Correction to #119's note on the GCC modules defect

**The characterisation I committed in #119 was wrong**, and this PR corrects it in place. The trigger is not re-exporting a governed type from a host-tools module. It is **`std::optional<T>` where `T` is, or contains, a class imported from another named module**. `std::expected` is unaffected — which is why `ShaderPackage::parse()` has always worked from plain translation units. I found this writing `UiPackageTests`, where `std::optional<ShaderPackage>` failed identically.

The practical rule for #121 and #124: prefer `mdux::core::Result` over `std::optional` for a type that crosses a module boundary.

---

- **Predecessor**: #129 (issue #119). This PR **cannot** merge before it — `mdux_bake_artifact(TOOL mdux-shaderbake …)` requires that target to exist, and CMake fails at configure time otherwise.
- **Shared registries edited**: `CMakeLists.txt` (one `mdux_bake_artifact()` call, appended after `add_subdirectory(tools)`), `tests/CMakeLists.txt` (one source added to `shader_tests`, one compile definition), `.gitignore`, `.gitattributes`.
- **Post-merge gate**: `develop` CI green before #121 opens.


recipes/shader/mdux-ui/ui.frag
```diff
@@ -0,0 +1,40 @@
...
+
+layout(set = 0, binding = 0) uniform sampler2D uAtlas;
+
+layout(location = 0) in vec4 fragColor;
+layout(location = 1) in vec2 fragUv;
+layout(location = 2) flat in uint fragMode;
+
+layout(location = 0) out vec4 outColor;
+
+const uint kModeSolid = 0u;
```
camille-martin-paris: Please drop the `k` prefixes here (`modeSolid`, `modeCoverageR8`). The repository naming rule uses lowerCamelCase for constants too; the language being GLSL does not make these names part of a different project convention.

tests/shader/UiPackageTests.cpp
```diff
@@ -0,0 +1,159 @@
...
+import mdux.test;
+
+#include "../framework/MduXTest.hpp"
+
+namespace {
+
+namespace shader = mdux::shader;
+namespace evidence = mdux::evidence;
+
+const std::filesystem::path kPackageDir =
```
camille-martin-paris: Please rename this constant to lowerCamelCase without a `k` prefix (for example `packageDir`), as required for `const` variables by `CONTRIBUTING.md`.

tests/shader/UiPackageTests.cpp
```diff
@@ -0,0 +1,159 @@
...
+    if (!bytes.has_value()) {
+        return mdux::core::err(shader::SchemaError::MalformedPackage);
+    }
+    const std::string_view text{reinterpret_cast<const char*>(bytes->data()), bytes->size()};
+    return shader::ShaderPackage::parse(text);
+}
+
+}  // namespace
+
+TEST_CASE("The committed UI package parses and validates", "evidence-unit") {
```
camille-martin-paris, line 68: Please rewrite this new suite in BDD form using `SCENARIO` and `GIVEN`/`WHEN`/`THEN`, rather than `TEST_CASE`, per ADR-002.

tests/shader/UiPackageTests.cpp
```diff
@@ -0,0 +1,159 @@
+/**
+ * @file UiPackageTests.cpp
```
camille-martin-paris: Please remove the `@file` tag. The file-level Doxygen rule is `@brief` only, with no `@file` or `@author` tags.

ambroise-leclerc:

## On the BDD request

recipes/shader/mdux-ui/ui.frag
ambroise-leclerc: Renamed in the GLSL: `recipes/shader/mdux-ui/ui.frag:26-27` now declares `modeSolid` and `modeCoverageR8`. The paired comment in `mdux.draw` that named the old spellings has been corrected too.

tests/shader/UiPackageTests.cpp
ambroise-leclerc: Renamed.

tests/shader/UiPackageTests.cpp
ambroise-leclerc: Removed.

ambroise-leclerc (commented):

Already handled
- `k` prefixes in GLSL (`modeSolid`/`modeCoverageR8`), \`UiPackageTests\` constant, \`@file\` removed. BDD declined — agreed.

Two Copilot items

1. `tests/shader/UiPackageTests.cpp:42\` — \`readFile()\` assumes \`tellg()\` returns a non-negative value. On stream error \`tellg()\` returns -1, the \`static_cast<size_t>\` makes a huge allocation, and the function throws \`bad_alloc\` rather than returning \`nullopt\`. Guard \`tellg()\` and the subsequent read before allocating. Same pattern recurs on #131 (`Emit.cpp:51`).

2. `tests/shader/UiPackageTests.cpp:17,146\` — the SPIR-V contract check only validates the magic number and 4-byte alignment when it could validate each declared module's execution model against the package's stage field using the hosted \`mdux.tools.spirv::reflect()\`. The PR description says this is the "contract test because the evidence gate cannot cover this"; only checking magic is a weaker contract than the one described. Either reflect and assert the stage matches, or note in a comment why reflection is deliberately not used here.

tests/shader/UiPackageTests.cpp
ambroise-leclerc: Leaving this open rather than actioning it, because the framework it asks for is not available here and its replacement is blocked upstream.

 `tests/shader/UiPackageTests.cpp` is in that first batch.

tests/shader/UiPackageTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+    }
+    const auto size = static_cast<std::streamsize>(file.tellg());
+    file.seekg(0);
+    std::vector<std::byte> bytes(static_cast<std::size_t>(size));
+    file.read(reinterpret_cast<char*>(bytes.data()), size);
+    if (!file) {
+        return std::nullopt;
+    }
+    return bytes;
+}
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `cf765e1` — negative `tellg()` and a failed `seekg` both return `nullopt` before anything is sized. Same pattern fixed in `Emit.cpp` on #131.

tests/shader/UiPackageTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * These tests are the other half. They assert the contract in one place, so changing a descriptor
+ * binding, a push-constant size, or which stage sees what is a deliberate edit here rather than a
+ * surprise in #124 or a validation-layer message on a device.
+ */
+import std;
+import mdux.core.result;
+import mdux.evidence.digest;
+import mdux.shader.schema;
+import mdux.test;
```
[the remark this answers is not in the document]
ambroise-leclerc: Good call, and taken in `cf765e1`. The test was named "well-formed SPIR-V for its declared stage" while checking four bytes, which is a claim it could not support. It now reflects each range through `mdux.tools.spirv` and asserts the execution model *and* entry point match the package.

The case this catches and the magic number could not: two modules swapped in the sidecar are both valid SPIR-V and both correctly digested, so the first symptom would have been a vertex shader bound to the fragment stage on a device.

Issue #122, opened 2026-08-01
S4 — Migrate VulkanSCTriangleExample to embedded shaders


ambroise-leclerc:
Part of #13. Depends on the emitter child and #48's compiler/example resolution.

Migrate `VulkanSCTriangleExample` from working-directory shader file loads to the generated embedded shader package. Remove the mismatched build/runtime shader paths and retain caller-owned Vulkan/windowing setup.

Acceptance: the example builds on every toolchain where it is enabled, performs no runtime shader file I/O, and its shader bytes match the committed evidence artifact.

Issue #108, opened 2026-08-01
S1 — Re-baseline ADR numbering and supersede the Catch2 decision


ambroise-leclerc:
Part of #10 (Wave 3 documentation architecture).

## Problem

`docs/adr/` still has two files presented as ADR-002, while the Catch2 decision was never implemented and the in-repository `MduXTest` framework is already the tested reality. The authoritative index therefore cannot give a reader one contiguous, truthful decision trail.

## Work

- Preserve the numbers of accepted ADR-001 and ADR-003 through ADR-007 so existing citations remain valid.
- Mark ADR-002 (Catch2) `Superseded` and point it at a new accepted ADR-008 describing `MduXTest`, its discovery model, labels, JSON output, and zero-SOUP rationale.
- Remove `ADR-002-implementation-plan.md` from the authoritative ADR directory. Record its disposition as historical implementation material; do not renumber it into a new decision.
- Rebuild `docs/adr/README.md` as a one-row-per-decision contiguous index with no duplicate decision number.
- Update live references that still describe Catch2 as selected or `MduXTest` as pending.

## Acceptance

- The ADR index contains exactly ADR-001 through ADR-008, with ADR-002 visibly superseded by ADR-008.
- No active document claims Catch2 is used.
- Documentation lint, internal-link lint, and all affected tests pass.
- No normative standard text is introduced.

PR #131, opened 2026-08-02
S3 — Shader C++ emitter with module and header outputs

### LINKED ISSUE #121: S3 — Shader C++ emitter with module and header outputs

ambroise-leclerc:
Closes #121. Third story of #13. **Stacked on #130 → #129 — merge those first.**

Renders a committed shader package as C++ the renderer can hold as `constexpr` data: a module interface for consumers that `import`, and a header for consumers that cannot.

## Generated code lives in the build tree, never in git

The reviewed artifact is `generated/shader/<id>/{package.json, shaders.spv}` — read as JSON and digests, byte-compared in CI. The C++ is a mechanical rendering of exactly those bytes, regenerated on every build, and **is not committed**.

Committing it would put a second representation of one artifact under review, and the second one is the one nobody reads: a reviewer cannot meaningfully check three thousand hex bytes, so a diff in generated source is noise that hides signal. Reviewing the JSON and the digests, and deriving the C++ mechanically, is the arrangement where the thing a human checks is the thing that matters.

**The emitter refuses to render a sidecar whose digest does not match the one `package.json` records.** Without that check, a hand-edited sidecar would be compiled into a binary while every artifact check stayed green — the digest under review describing one thing and the linked bytes another. That refusal is what makes "generated code needs no review" defensible rather than merely convenient. It has its own test.

## C arrays, not `std::array`

Namespace-scope C arrays with `std::span` accessors, as the epic requires. A `std::array` of a few thousand elements in a module interface triggers the GCC 15/16 modules ICE this repository has already hit twice (`tests/CMakeLists.txt`'s `-O0` workaround; `Units.cppm`'s note on friend-defaulted comparisons). Not a compromise — the form that compiles.

## The view layer belongs in the schema

`mdux.shader.schema` gains `ModuleView` and `PackageView` beside the owning `ShaderPackage` they mirror. Two representations of one format, in one module, deliberately: the owning form is right for a baker assembling an artifact and a reader validating one; the view form is right for a device that wants a contract with no allocation and no parsing. A view type declared in the *emitter* would be a second definition of the format — the arrangement epic #13 names as the Wave 2 lesson.

## Its own executable, not an `emit` mode of `mdux-shaderbake`

Two reasons, and the second is the real one:

- `mdux::tools::cli::parse` implements one grammar shared by every baker. Widening it would put a mode there that only some tools have.
- Emission is **kind-specific in a way baking is not**. A font atlas and an ML model will each want to generate different types, so a shared `emit` verb would be a shared verb with nothing shared behind it.

Diagnostics still go through `cli::Diagnostic`/`render`/`exitStatus`, so `--format=json` is the same envelope every other MduX tool emits — which is the part an agent actually depends on.

It also keeps this PR off `tools/common/Cli.*`, which #128 is editing. That is the #117 rule in practice rather than in principle.

## Tests

**Both generated forms are compiled into one binary and compared** — that is #121's acceptance, and it works because the module form is attached to a named module and the header form is not, so they are distinct entities that can coexist. Three separate things are then checked, which are not the same thing:

1. The two forms expose byte-identical SPIR-V and an identical contract. If they diverged, whichever of #122 or #124 used the other one would be linking different shaders than the reviewer read.
2. The generated data agrees with `generated/shader/mdux-ui/package.json`.
3. The bytes are the *reviewed* bytes — checked against the digest the committed package records, which is the number a reviewer actually looked at.

Plus the emitter's refusals: unreadable package, a package that is not a shader package, a missing sidecar, a digest mismatch, a sidecar of the wrong length. And `write()` leaving an unchanged file untouched, so a rebuild that changes nothing does not restamp it and force every consumer to recompile.

The CMake regex that predicts the generated filenames is **asserted against `identifierFor()`** rather than trusted, because a mismatch would surface as a missing file with no explanation.

## Verification

All executed locally with GCC 15.2:

```
shader_tests                      95/95 passed
ctest --test-dir build-mac        272/273 passed
git status --porcelain            empty after a full clean build
```

The empty `git status` after a clean build **is** the "no normal build changes tracked files" acceptance criterion, checked rather than assumed. The one test failure is `InstallTreeConsumer`, which runs its own nested CMake configure and cannot inherit this machine's linker workaround — a local-environment artifact.

Lints: `mdux-evidence-lint` OK (18 files), `mdux-docs-lint` OK (71 files).

One thing I fixed after reading the first output: the provenance comment the emitter writes had an **absolute path** in it. The generated files are not committed, but two developers' copies should still be identical — an absolute path there would defeat a shared compiler cache for no benefit. The custom command now runs from the repository root and passes a relative path, as `mdux_bake_artifact()` does.

---

- **Predecessor**: #130 (issue #120), itself on #129 (#119). This PR cannot merge before them — `mdux_emit_shader_package()` fails at configure time if `generated/shader/mdux-ui/package.json` is absent.
- **Shared registries edited**: `CMakeLists.txt` (include + one `mdux_emit_shader_package()` call), `tools/CMakeLists.txt` (`Emit.cppm`/`Emit.cpp` added to the existing lib, one new executable), `tests/CMakeLists.txt` (four sources, a named file set, an include dir). **Deliberately does not touch** `tools/common/Cli.*` or `tests/tools/CliTests.cpp`, which #128 edits.
- **Post-merge gate**: `develop` CI green before #123 opens.


tests/shader/EmitTests.cpp
```diff
@@ -0,0 +1,255 @@
...
+    std::vector<std::string> codes;
+    for (const cli::Diagnostic& diagnostic : diagnostics) {
+        codes.push_back(diagnostic.code);
+    }
+    return codes;
+}
+
+}  // namespace
+
+TEST_CASE("identifierFor maps a package id to a C++ identifier", "evidence-unit") {
```
camille-martin-paris, line 100: Please convert both new test suites (`EmitTests.cpp` and `GeneratedTests.cpp`) from `TEST_CASE` to Catch2 BDD `SCENARIO`/`GIVEN`/`WHEN`/`THEN`, per ADR-002.

tools/shader/Emit.cpp
```diff
@@ -0,0 +1,335 @@
...
+import mdux.evidence.digest;
+import mdux.shader.schema;
+import mdux.tools.cli;
+
+namespace mdux::tools::shaderemit {
+
+namespace {
+
+// Stable diagnostic codes; see docs/governance/schemas/diagnostic.schema.json.
+constexpr std::string_view kPackageUnreadable = "SHE001";
```
camille-martin-paris: Please remove `k` prefixes from constants throughout this change (`packageUnreadable`, `packageUnparsed`, `toolName`, etc.). `CONTRIBUTING.md` explicitly requires lowerCamelCase for `const` and `constexpr` variables.

tools/shader/Emit.cpp
```diff
@@ -0,0 +1,335 @@
...
+/// outputs cannot describe different bytes or a different contract.
+[[nodiscard]] std::string renderBody(const shader::ShaderPackage& package,
+                                     std::span<const std::byte> sidecar,
+                                     std::string_view identifier) {
+    std::string out;
+
+    out += "namespace mdux::shader::generated::" + std::string{identifier} + " {\n\n";
+
+    out += "/// The package id this data was generated from.\n";
+    out += "inline constexpr std::string_view kId = \"" + escape(package.header.id) + "\";\n\n";
```
camille-martin-paris: The emitted C++ must conform to the same naming rules: please generate `id`, `spirvSha256`, `spirvBytes`, `modules`, `descriptors`, and `pushConstants` rather than `kId`, `kSpirvSha256`, and the other `k*` names. Update the string-based emitter tests with the generated contract.

tools/shader/Emit.cpp
```diff
@@ -0,0 +1,335 @@
+/**
+ * @file Emit.cpp
```
camille-martin-paris: Please remove `@file` tags from all new C++ files in this PR. The project requires file-level Doxygen to use `@brief` only.

tools/shader/ShaderEmitMain.cpp
```diff
@@ -0,0 +1,99 @@
+/**
```
camille-martin-paris, line 1: Please rename this `.cpp` file in UpperCamelCase (for example `ShaderEmitMain.cpp`) and update CMake, per the repository file-naming rule.

ambroise-leclerc:

## On the BDD request

tools/shader/Emit.cpp
ambroise-leclerc: Removed throughout.

tools/shader/Emit.cpp
ambroise-leclerc: Done — the generated C++ follows the same rule as hand-written code. `tools/shader/Emit.cpp:129-193` emits `id`, `spirvSha256`, `spirvBytes`, `modules`, `descriptors` and `pushConstants"`.

tools/shader/Emit.cpp
ambroise-leclerc: Removed.

tools/shader/ShaderEmitMain.cpp
ambroise-leclerc: Renamed to `ShaderEmitMain.cpp`, CMake updated.

ambroise-leclerc (commented):

Already handled
- `@file` removed, `k` prefixes dropped (including in *generated* C++), `ShaderEmitMain.cpp` renamed. BDD declined — agreed.

Three Copilot items, two are real bugs

1. `tools/shader/ShaderEmitMain.cpp:65\` — \`--format=text\` never resets the output format. \`--format=json --format=text\` still emits JSON because last-wins is not implemented. Real CLI bug; the format flag should set, not merge. Worth a one-line test in `CliTests.cpp` to pin it.

2. `cmake/MduXShaderEmit.cmake:57\` — the CMake identifier regex diverges from \`identifierFor()\`. \`identifierFor("2d")\` yields \`_2d\` (a valid identifier); the regex would produce \`2d\` (not a valid module/interface filename). The PR description even claims the regex is "asserted against `identifierFor()`" — if it is, the digit-leading case is the one it would miss. Add a digit-leading id to the assertion fixture or document the constraint on id authors.

3. `tools/shader/Emit.cpp:51\` — same \`tellg()\`-returns-−1 pattern as #130. Guard before the \`static_cast<size_t>\`.

tests/shader/EmitTests.cpp
ambroise-leclerc: Leaving this open rather than actioning it, because the framework it asks for is not available here and its replacement is blocked upstream.

 `tests/shader/EmitTests.cpp` and `tests/shader/GeneratedTests.cpp` are in that first batch.

tools/shader/ShaderEmitMain.cpp
```diff
@@ -0,0 +1,98 @@
...
+            std::print(std::cout, "{}", usage());
+            return 0;
+        }
+        if (argument == "--format=json") {
+            format = cli::Format::Json;
+            continue;
+        }
+        if (argument == "--format=text") {
+            continue;
+        }
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `07bcd57` — it assigns now, so last flag wins. It could only ever have been a no-op as written.

tools/shader/Emit.cpp
```diff
@@ -0,0 +1,334 @@
...
+    file.seekg(0);
+    std::vector<std::byte> bytes(static_cast<std::size_t>(size));
+    if (size > 0) {
+        file.read(reinterpret_cast<char*>(bytes.data()), size);
+        if (!file) {
+            return std::nullopt;
+        }
+    }
+    return bytes;
+}
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `07bcd57`. Note the `size > 0` guard did not help: the vector was sized before it.

cmake/MduXShaderEmit.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+    # The identifier mdux-shaderemit derives from the id: every character that is not
+    # alphanumeric becomes an underscore. Kept in step with identifierFor() in Emit.cpp, and
+    # asserted by a test rather than trusted, because a mismatch here would surface as a missing
+    # file with no explanation.
+    string(REGEX REPLACE "[^A-Za-z0-9]" "_" identifier "${ARG_ID}")
+
+    set(output_dir "${CMAKE_BINARY_DIR}/mdux_generated/shader")
+    set(module_file "${output_dir}/${identifier}.cppm")
+    set(header_file "${output_dir}/${identifier}.hpp")
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `07bcd57`, and this one had a second problem underneath it worth naming.

The derivation is now `mdux_shader_identifier()` and applies the leading-underscore rule. But `GeneratedTests` claimed to keep CMake and C++ in step while only asserting `identifierFor()`'s return values — a C++ assertion cannot observe a CMake regex, so it structurally could not have caught this, and did not. The build now writes its own answers for eight ids to a file at configure time and the test compares those against `identifierFor()`. Both halves execute, which is what makes the parity claim true.

Issue #126, opened 2026-08-01
S8 — First lavapipe pixel test


ambroise-leclerc:
Part of #13. Depends on the offscreen renderer child.

Add the first pixel-level CI test using lavapipe: render known solid geometry offscreen, read it back, and compare exact expected bounds/colors with actionable diff output. Register it with the `pixel` label and preserve source-tree cleanliness.

Acceptance: the positive case passes in Linux CI, deliberate one-pixel/color changes fail with coordinates and expected/actual values, and no golden is silently rewritten.

Issue #125, opened 2026-08-01
S7 — Headless offscreen renderer and CPU readback


ambroise-leclerc:
Part of #13. Depends on the Vulkan renderer child.

Add a headless offscreen render target and deterministic CPU readback surface for tests. Keep governed draw construction independent from Vulkan and confine image/resource ownership to the adapter test harness.

Acceptance: fixed-size surface creation, clear/solid-rectangle rendering, readback, teardown, and failure paths are covered without a desktop window.

PR #132, opened 2026-08-02
S5 — mdux.draw fixed-budget draw types

### LINKED ISSUE #123: S5 — mdux.draw fixed-budget draw types

ambroise-leclerc:
Closes #123. Fourth story of #13. **Stacked on #131 → #130 → #129.**

The governed description of a frame: `UiVertex`, `DrawCommand`, `DrawBudget`, and a bounded, non-allocating `DrawList`.

## No Vulkan anywhere, which is the point

There is no Vulkan handle, descriptor, or device concept in this module — `mdux_verify_trust_zones()` fails the configure step if there ever is, and `draw_tests` links `MduX::Core` only so a boundary slip shows up as a link error rather than a review comment.

A `DrawList` is a description of a frame that can be inspected by a test, compared against an expected frame, or rendered by something else entirely, **with no GPU present**. That is what makes #125's headless verification and #126's pixel test possible rather than merely convenient.

## Fixed budget means the storage is the caller's

A `DrawList` never allocates. It is built over spans the caller owns — typically a `static` array sized from a budget the `.medui` compiler computed for the screen — and fails explicitly the moment a frame would exceed them. A UI that silently grew its buffers would have a per-frame cost nobody bounded, and on a device the first symptom of that is a missed deadline rather than an error.

Two implementation details worth a look:

- **Every `add` either records a primitive completely or records nothing.** All budget checks run before the first write, so a refused primitive leaves no half-written geometry. There is a test for exactly that, because the failure it prevents renders as stray triangles rather than as an error.
- **Checks are subtractions against remaining capacity, never additions against the limit.** With counts near the top of their range an addition could wrap and read as acceptable.

Consecutive primitives sharing a clip extend the current command rather than starting a new one, which is what makes a command budget a meaningful number — it tracks clip changes, not rectangle count. `setClip` records rather than applies, so a clip nobody drew under produces no empty command consuming a budget slot.

## The layout is the shader's, pinned

`UiVertex` is exactly the 24-byte layout `recipes/shader/mdux-ui/ui.vert` declares, and `DrawMode`'s values are the shader's `kModeSolid` / `kModeCoverageR8` constants. `static_assert`s pin the size, alignment and every member offset.

Nothing here can read GLSL, so **the assertions and the paired comments are the mechanism** — there is no automatic check, and changing one side without the other should be uncomfortable. Flagging that honestly rather than implying more coverage than exists.

`packColor` writes bytes in memory order rather than shifting into a `std::uint32_t`, because the shader reads them as `R8G8B8A8_UNORM` in *memory* order. The test checks the four bytes individually rather than the integer — an integer comparison would pass on a host whose byte order made the shader read them backwards.

**16-bit indices cap a list at 65536 vertices.** A deliberate ceiling, not an oversight: it halves the index buffer, and a governed screen needing more than sixteen thousand rectangles in one frame has a budget problem a wider index would only hide. A budget above the cap is rejected at construction rather than producing wrapped indices that point at the wrong vertices with no error anywhere.

## Tests

25 cases. The ones that earn their keep are at the boundaries:

- A frame that **exactly fills** its budget succeeds (the off-by-one that would make a budget one primitive smaller than it claims), and the very next primitive fails with nothing recorded.
- Each budget can bind independently — storage generous on vertices and tight on indices proves the index check fires rather than the vertex one.
- A command budget of one is **not** exceeded while the clip holds, which guards the command check against firing per primitive instead of per change.
- The same primitives produce byte-identical buffers from two lists with different storage sizes, one of which drew a different frame first and was reset. That is the property #126's pixel comparison rests on: a frame is a function of its primitives and nothing else.

## Two module-system notes

`offsetof` is a macro and `import std;` exports no macros, so `<cstddef>` comes through the global module fragment — and in the test it must precede the imports, since included afterwards it redefines `std::byte` against the module. Same ordering the Vulkan SC tests already use for `<vulkan/vulkan.h>`.

Braced initialisers had to be hoisted out of `CHECK()`/`REQUIRE()`: their commas read as macro argument separators.

## Verification

```
draw_tests                       25/25 passed
ctest --test-dir build-mac       297/298 passed
cmake --build build-mac          0 errors, 0 warnings under -Werror
mdux_verify_trust_zones          2 governed target(s) clean of Vulkan/windowing dependencies
```

The one failure is `InstallTreeConsumer`, the known local-environment case.

---

- **Predecessor**: #131 (issue #121). Rebased onto it after its GCC 15 fix; no structural dependency, but it is next in the epic's merge order.
- **Shared registries edited**: `CMakeLists.txt` (one module + one source added to the `MduXCore` file set), `tests/CMakeLists.txt` (one new target, appended). Nothing reordered or removed.
- **Post-merge gate**: `develop` CI green before #124 opens.


include/mdux/draw/Draw.cppm
```diff
@@ -0,0 +1,201 @@
...
+static_assert(offsetof(UiVertex, u) == 8, "uv must be at offset 8");
+static_assert(offsetof(UiVertex, color) == 16, "colour must be at offset 16");
+static_assert(offsetof(UiVertex, mode) == 20, "mode must be at offset 20");
+
+/// Indices are 16-bit, which caps a list at 65536 vertices. That is a deliberate ceiling rather
+/// than an oversight: it halves the index buffer, and a governed screen that needs more than
+/// sixteen thousand rectangles in one frame has a budget problem the index width would only hide.
+using Index = std::uint16_t;
+
+inline constexpr std::uint32_t kMaxVertices = 65536;
```
camille-martin-paris: Please remove the `k` prefix from constants in this PR. This public name should be lowerCamelCase, and the implementation/test constants should likewise become `verticesPerRect`, `indicesPerRect`, `rect`, and `red`. `CONTRIBUTING.md` explicitly includes `const` and `constexpr` variables in this rule.

tests/draw/DrawTests.cpp
```diff
@@ -0,0 +1,478 @@
...
+constexpr core::Rect kRect{.x = 10, .y = 20, .width = 30, .height = 40};
+constexpr core::ColorRgba8 kRed{.r = 255, .g = 0, .b = 0, .a = 255};
+
+}  // namespace
+
+// ---------------------------------------------------------------------------
+// Layout
+// ---------------------------------------------------------------------------
+
+TEST_CASE("UiVertex is the 24-byte layout the shader declares", "evidence-unit") {
```
camille-martin-paris, line 56: Please rewrite this new suite in Catch2 BDD style using `SCENARIO` with meaningful `GIVEN`/`WHEN`/`THEN` sections, not `TEST_CASE`, per ADR-002.

include/mdux/draw/Draw.cppm
```diff
@@ -0,0 +1,201 @@
+/**
+ * @file Draw.cppm
```
camille-martin-paris: Please remove the `@file` tags from the new module, implementation, and test sources. File-level Doxygen in this project is `@brief` only.

ambroise-leclerc:

## On the BDD request

include/mdux/draw/Draw.cppm
ambroise-leclerc: Renamed — the public constant is `maxIndexableVertices`, which also reads better than `maxVertices` next to `budget.maxVertices`.

include/mdux/draw/Draw.cppm
ambroise-leclerc: Removed.

ambroise-leclerc (commented):

Already handled
- `k` prefix removed (`maxIndexableVertices`), `@file` removed. BDD declined — agreed.

Two Copilot items, both worth a fix

1. `include/mdux/draw/Draw.cppm:135\` — \`packColor()\` is documented as writing bytes in memory order specifically to stay correct on big-endian hosts, but the implementation uses shifts, which makes the in-memory byte order endianness-dependent. The comment and the shader contract (R8G8B8A8_UNORM read in memory order) say one thing; the code does another. Either write four named bytes as the comment claims, or delete the endianness rationale. The test checking four individual bytes would catch a host-byte-order bug but not the contradiction on a little-endian host where shifts happen to coincide.

2. `include/mdux/draw/Draw.cppm:198\` and \`src/draw/Draw.cpp:89\` — \`clipSet_\` is written by \`reset()\` and \`setClip()\` but never read anywhere. It is dead state. Under Clang \`-Wall\` it can produce \`-Wunused-private-field\` once the warnings settings are tightened; more importantly it makes the clip/command logic harder to reason about. Either remove \`clipSet_\` entirely or wire it into a real check. If it was intended to suppress empty-clip commands, the test that earns its keep says that already works via \`commandCount_\`, so removing it should be safe.

tests/draw/DrawTests.cpp
ambroise-leclerc: Leaving this open rather than actioning it, because the framework it asks for is not available here and its replacement is blocked upstream.

 `tests/draw/DrawTests.cpp` is in that first batch.

include/mdux/draw/Draw.cppm
```diff
@@ -0,0 +1,200 @@
...
+/// Packs a colour into the 32-bit form the vertex carries, in memory order R, G, B, A.
+///
+/// Written byte by byte rather than by shifting into a `std::uint32_t`, because the shader reads
+/// these four bytes as `R8G8B8A8_UNORM` in *memory* order - a shift-based packing would produce
+/// different bytes on a big-endian host, and the whole point of a governed type is that it means
+/// the same thing everywhere.
+[[nodiscard]] constexpr std::uint32_t packColor(mdux::core::ColorRgba8 color) noexcept {
+    return static_cast<std::uint32_t>(color.r) | (static_cast<std::uint32_t>(color.g) << 8) |
+           (static_cast<std::uint32_t>(color.b) << 16) | (static_cast<std::uint32_t>(color.a) << 24);
+}
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `d82db52` with `std::bit_cast` from a `std::array<std::uint8_t, 4>`, which states the memory order directly and stays constexpr.

Worth recording how this got past: the test beside it (`packColor writes bytes in memory order R, G, B, A`) already asserted the bytes individually, precisely so an integer comparison could not hide the problem. So the contract was right, documented, and tested — and the implementation still did not meet it, because every machine anyone ran it on was little-endian. The comment was the only thing describing the intent, and it was describing code that did not exist.

include/mdux/draw/Draw.cppm
```diff
@@ -0,0 +1,200 @@
...
+    std::span<UiVertex> vertices_;
+    std::span<Index> indices_;
+    std::span<DrawCommand> commands_;
+    DrawBudget budget_{};
+    std::uint32_t vertexCount_{0};
+    std::uint32_t indexCount_{0};
+    std::uint32_t commandCount_{0};
+    mdux::core::Rect clip_{};
+    bool clipSet_{false};
+};
```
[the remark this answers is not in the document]
ambroise-leclerc: Removed in `d82db52`.

src/draw/Draw.cpp
```diff
@@ -0,0 +1,167 @@
...
+    clip_ = {};
+    clipSet_ = false;
+}
+
+void DrawList::setClip(const mdux::core::Rect& clip) noexcept {
+    // Recorded rather than applied: the next primitive notices the change and starts a command.
+    // Doing it here would emit an empty command for a clip nothing was drawn under.
+    clip_ = clip;
+    clipSet_ = true;
+}
```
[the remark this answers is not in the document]
ambroise-leclerc: Removed in `d82db52` — your reading is right, `clip_` alone is sufficient because a default-constructed `Rect` already means "no clip", which is exactly what `reset()` restores. The remaining member now carries a note saying why no flag is needed, so it does not get re-added.

Issue #59, opened 2026-07-26
S4 — Determinism enforcement and the cross-toolchain ctest


ambroise-leclerc:
Correct kernel code is necessary but not sufficient in C++. Three hazards Rust does not have, and the guards for each.

**1. FP contraction.** GCC and Clang default to `-ffp-contract=fast` outside strict modes and **will** fuse `acc += w * x` into an FMA, changing the result. Set explicitly on the kernels target:

- GCC / Clang: `-ffp-contract=off -fno-fast-math`
- ~~MSVC: `/fp:precise /fp:contract`~~ — **both wrong; see the correction below.**

> **Correction (applied in #146, kept here so the original text is not reapplied).**
>
> The MSVC flags above are wrong in *both* directions, and neither mistake is obvious from reading
> them:
>
> - **`/fp:contract` enables contraction into FMA.** It is MSVC's opt-in for exactly the
>   transformation this issue exists to forbid. It now appears in the *forbidden* list rather than
>   being applied.
> - **`/fp:precise` breaks the build.** `cl` defines `_M_FP_PRECISE` when the flag is passed
>   explicitly, the prebuilt `std` module is compiled without it, and importing `std` from a
>   translation unit whose command line disagrees raises `C5050`
>   ("Possible incompatible environment while importing module 'std'"). With warnings as errors,
>   every governed module that imports `std` fails to compile. This was observed in CI, not
>   theorised — it broke six PRs mid-stack.
>
> **What was implemented instead: nothing is passed on MSVC.** `/fp:precise` is already the default
> and VS2022 does not contract under it, so an unadorned MSVC build already behaves as ADR-008
> decision 3 requires. Enforcement on MSVC is therefore entirely
> `mdux_verify_fp_determinism()`'s job, which rejects `/fp:fast` *and* `/fp:contract` if either
> ever arrives.

**2. `-ffast-math` / `/fp:fast` creeping in** via a preset, a toolchain file, or a dependency's interface flags. Add a CMake guard that inspects the effective compile options of every governed target and `message(FATAL_ERROR)`s on `-ffast-math`, `-Ofast`, `-funsafe-math-optimizations` or `/fp:fast`. This is the guard that catches the change nobody thought was related.

**3. x87 excess precision** — only bites 32-bit x86, where the FPU computes at 80 bits and rounds unpredictably on spill. Windows is 64-bit-only here; require SSE2 math on any 32-bit Linux target or refuse to build.

[...]

PR #143, opened 2026-08-03
docs(adr): record the zero-SOUP ML inference decision

### LINKED ISSUE #56: S1 — ADR: zero-SOUP ML inference pipeline
The MduX equivalent of TrustSC's ADR-017, adapted to C++.

**Context.** Manufacturers want to embed learned models — signal classification is the entry case — without pulling ONNX Runtime, PyTorch or `tract` into a Class B/C device. They also want to prototype against openly available pretrained weights before they have their own qualified training data, and then move to production **without re-validating the inference engine**.

**Decision, in outline:**

1. **Four modules, not three.** TrustSC splits schema / authoring / runtime. MduX adds `mdux.ml.kernels` as a separate governed module imported by *both* the runtime and the host baker. In Rust "authoring uses the same kernels as the runtime" is a discipline; here it is one definition, one object file, one set of flags. If host and device disagree, it is the FPU, not the code.
2. **Weights are data, not code.** A `ModelPackage` is baked offline and committed under `generated/model/<id>/`, byte-verified like every other artifact. Swapping demonstrator weights for clinically-qualified weights is a re-bake — **zero application source change**.
3. **Determinism by strictly-ordered scalar arithmetic** — fixed accumulation order, plain `f32` multiply-then-add, never FMA, never SIMD intrinsics.
4. **Golden vectors baked into the package and re-run at construction**, comparing **bit patterns**. `create()` fails closed on any divergence, detecting toolchain miscompilation, target FP drift, or a corrupted package before the device classifies anything real.
5. **v1 scope:** `f32` only; Dense, Conv1D, MaxPool1D/AvgPool1D, Flatten, Relu/Sigmoid/Softmax. `int8`, recurrent layers and attention are out of scope and need a follow-up ADR, not an unreviewed runtime extension.
6. **Weight import adds no runtime-adjacent SOUP** — safetensors is hand-parsed (a `u64` header length, a JSON header, raw bytes) in a host-only tool that never runs in CI and never touches a device module.

**Consequences to record:** no foreign inference stack is ever linked into a device target; a demonstrator build and a production build differ only in one committed JSON file; the golden self-test costs bounded startup work, not per-frame work; and model complexity is deliberately capped by v1's kernel set.

Part of #18.

ambroise-leclerc:
ADR-008, the reference the rest of epic #18 is built against — the MduX adaptation of TrustSC's ADR-017.

**Four modules rather than three.** `mdux.ml.kernels` is separate so the device runtime and the host golden-vector generator import the same object code. That identity is what makes a golden-vector mismatch diagnostic rather than ambiguous — there is no second implementation that could have drifted.

Also records the two guarantees C++ gives away that Rust holds: FP contraction defaults to fusing `acc += w * x` into an FMA, and `-ffast-math` can arrive from a preset or a dependency's interface flags. Both need explicit enforcement (#59) rather than correct kernel source alone.

First of a 9-PR stack for epic #18. Each PR targets its predecessor.

Closes #56.

### DIFF
--- a/docs/adr/ADR-008-zero-soup-ml-inference.md
+++ b/docs/adr/ADR-008-zero-soup-ml-inference.md
@@ -1,0 +1,261 @@
+# ADR-008: Zero-SOUP ML inference
+
+## Status
+Accepted (2026-08-03)
+
+## Context
+Manufacturers want to embed learned models in a device — 1-D signal classification is the entry
+case — without pulling ONNX Runtime, PyTorch, LiteRT or `tract` into a Class B/C build. Any of those
+is SOUP of a particularly awkward kind: large, fast-moving, transitively dependent on a
+general-purpose numerics stack, and impossible to characterise as "the code that computes this
+device's output" in a way a reviewer can follow.
+
+There is a second, less obvious requirement. Manufacturers want to prototype against openly
+available pretrained weights *before* they have their own qualified training data, and then move to
+production **without re-validating the inference engine**. If swapping weights means recompiling
+application source, the demonstrator and the product are different software and the demonstrator's
+verification evidence buys nothing.
+
+The evidence pipeline (ADR-007) already answers "which tool, at which version, from which input,
+produced this artifact" for baked assets, and `mdux.evidence.digest` (issue #50) already exists and
+already anticipates this use. What is missing is the decision about *where inference itself lives*
+and what makes its arithmetic trustworthy across two compilers and a device FPU.
+
+This ADR is the MduX equivalent of [TrustSC ADR-017](https://github.com/ambroise-leclerc/TrustSC),
+adapted to C++. The adaptation is not cosmetic: two of the decisions below exist because C++ gives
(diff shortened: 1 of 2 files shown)


0xfritz-dev (approved):
LGTM. Clean ADR that correctly establishes the 4-module split (schema/kernels/runtime/host-tools) so runtime + golden generator share identical object code — this is what makes golden mismatches diagnostic rather than ambiguous. Documents the two C++ FP hazards (FMA contraction default, -ffast-math leakage) that the rest of the stack addresses. Sets up epic #18 correctly.

ambroise-leclerc:
Thanks both.

**TrustSC link → ADR-017 file (declined, with reason).** Kept as a directory link deliberately. I can't verify the exact filename in the TrustSC repo from this environment, and a deep link that 404s is worse for an auditor than one that lands on the index — the ADR number is already in the link text, so nothing is ambiguous about *which* document is meant. Happy to make it a file link if you paste the filename.

PR #144, opened 2026-08-03
feat(ml): add mdux.ml.schema, the canonical model package format

### LINKED ISSUE #57: S2 — mdux.ml.schema
`include/mdux/ml/Schema.cppm`. **Governed zone**, header-only, no I/O, no parsing.

```cpp
enum class LayerKind  : std::uint8_t { dense, conv1d, maxPool1d, avgPool1d, flatten };
enum class Activation : std::uint8_t { none, relu, sigmoid, softmax };

struct TensorRef {                       // offset into the weight blob, NOT a pointer
    std::uint64_t byteOffset;
    std::array<std::uint32_t, 3> shape;
    std::uint8_t rank;
};

struct LayerDesc {
    LayerKind kind; Activation activation;
    std::uint32_t inLength, inChannels, outLength, outChannels;
    std::uint32_t kernelSize, stride;
    TensorRef weights, bias;
};

struct GoldenVector {
    std::span<const std::uint32_t> inputBits, expectedOutputBits;
};

struct ModelPackage {
    std::string_view id;
    std::uint32_t schemaVersion;
    std::array<std::uint8_t, 32> weightsDigest;
    std::uint64_t weightsByteLength;
    std::span<const LayerDesc>    layers;
    std::span<const GoldenVector> goldens;
    std::uint32_t inputLength, outputLength, maxScratchFloats;
};
```

**Two design points worth defending in review:**

`TensorRef` holds a **byte offset, not a pointer**. That is what lets `ModelPackage` be a `constexpr` object while the weights live in a separately-supplied blob — a pointer would make it non-constexpr and force the weights into generated source.

`GoldenVector` stores `u32` **bit patterns**, not floats. Comparison is bitwise. Never decimal, never an epsilon. An epsilon comparison would silently accept exactly the FP drift this mechanism exists to detect.

The schema **rejects any dtype other than `f32`** — v1 scope, per S1.

Part of #18.

ambroise-leclerc:
The record every other part of epic #18 imports rather than restating.

Two choices worth the review time:

- **`TensorRef` holds a byte offset, not a pointer**, which is what lets a `ModelPackage` be `constexpr` while its weights live in a separately-supplied blob. A pointer would force multi-megabyte weights into generated source.
- **`GoldenVector` stores `u32` bit patterns, not floats**, because comparison has to be bitwise — an epsilon would accept exactly the FP drift the goldens exist to detect.

Header-only, so a generated package validates at compile time. The spec asserts that with a `static_assert` on the reference architecture, and gives every `SchemaError` a mutation that produces exactly it — 28 invariants, each breaking one field of an otherwise-valid package.

Stacked on #56.

Closes #57.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,6 +71,7 @@ code*; treat this subsection as the direction that code is moving in.
 | `mdux.evidence.digest`, `.json`, `.report` | `include/mdux/evidence/` | `src/evidence/` |
 | `mdux.governance`, `mdux.governance.compliance` | `include/mdux/governance/` | `src/governance/` |
 | `mdux.shader.schema` (governed) | `include/mdux/shader/Schema.cppm` | `src/shader/Schema.cpp` |
+| `mdux.ml.schema` (governed) | `include/mdux/ml/Schema.cppm` | header-only |
 | `mdux.draw` (governed) | `include/mdux/draw/Draw.cppm` | `src/draw/Draw.cpp` |
 | `mdux.render.vulkan`, `mdux.render.offscreen` (adapter) | `include/mdux/render/` | `src/render/` |
 | `mdux.vulkansc.memory` | `include/mdux/vulkansc/MemoryPoolManager.cppm` | `src/vulkansc/MemoryPoolManager.cpp` |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,7 @@ target_sources(MduXCore
             include/mdux/governance/Governance.cppm
             include/mdux/governance/Compliance.cppm
             include/mdux/shader/Schema.cppm
+            include/mdux/ml/Schema.cppm
             include/mdux/draw/Draw.cppm
     PRIVATE
         src/evidence/Digest.cpp
--- a/include/mdux/ml/Schema.cppm
+++ b/include/mdux/ml/Schema.cppm
@@ -1,0 +1,478 @@
+/**
+ * @file Schema.cppm
+ * @brief Governed-zone ML model types: the canonical shape of every model package.
+ *
+ * @compliance ADR-004 Trust zones in C++ (governed zone: std only, no Vulkan, no windowing)
+ * @compliance ADR-005 Error handling and exceptions policy (Result-returning, noexcept)
+ * @compliance ADR-007 Evidence pipeline doctrine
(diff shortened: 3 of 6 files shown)


0xfritz-dev (approved):
LGTM. Strong design choices:  uses byte offset (not pointer) enabling  packages with weights in separate blob;  stores u32 bit patterns (not floats) for bitwise comparison — an epsilon would accept exactly the FP drift goldens exist to detect. 28 mutation tests each breaking one invariant is thorough. Header-only with compile-time validation via static_assert.

0xfritz-dev (approved):
LGTM. Strong design choices: `TensorRef` uses byte offset (not pointer) enabling `constexpr` packages with weights in separate blob; `GoldenVector` stores u32 bit patterns (not floats) for bitwise comparison — an epsilon would accept exactly the FP drift goldens exist to detect. 28 mutation tests each breaking one invariant is thorough. Header-only with compile-time validation via static_assert.

ambroise-leclerc:
All three addressed.

**`byteEnd()` overflow — real bug, fixed.** You're right, and it was exploitable in the way that matters: `byteOffset + byteLength()` wraps, so a large offset reports a tiny end and the tensor validates, after which the runtime reads off the end of the blob. Now compared as two subtractions that cannot overflow.

While fixing it I found the same class of bug one level down: `elementCount()` multiplies up to three `uint32` extents, so it can exceed `uint64` on its own. Both it and `byteLength()` now **saturate** rather than wrap — saturation turns an absurd shape into an absurd length, which the bounds check rejects; wrapping produced a *small, plausible* length that passed. Two corpus cases added, one per overflow path.

**`static_assert(sizeof(float) == 4)` — added.** Right that the on-wire assumption should be stated where `byteLength()` depends on it.

**Empty goldens — enforced, but in the runtime rather than here.** The gap is real and now closed: `Classifier1D::create()` rejects a package with no goldens (new `MlError::Code::NoGoldens`, PR #147), because a loop over an empty set reports success without checking anything, which is worse than having no control at all.

Kept out of `validate()` on purpose, and the module comment now explains why: `validate()` is a *structural* check the baker runs on a half-assembled package — it validates an architecture against imported weights *before* any goldens exist, and the golden generator needs a validated layer chain to run at all. Rejecting empty here would make that bootstrap impossible and put the safety argument at the wrong layer. The control belongs where an unverified model would actually execute.

PR #145, opened 2026-08-03
feat(ml): add mdux.ml.kernels, the arithmetic host and device share

### LINKED ISSUE #58: S3 — mdux.ml.kernels
`include/mdux/ml/Kernels.cppm` + `src/ml/Kernels.cpp`. **Governed zone.** The single most safety-relevant file in the ML subsystem, and the one imported by both the device runtime and the host baker.

**v1 kernel set:** `Dense`, `Conv1D`, `MaxPool1D`, `AvgPool1D`, `Flatten`, and the `Relu` / `Sigmoid` / `Softmax` activations. `f32` only.

**The determinism rule, documented per kernel:** every accumulation happens in a single fixed order — Conv1D sums over input channel then kernel tap; Dense sums over input feature — using plain `f32` multiply-then-add.

- **Never `std::fma` or `f32::mul_add`.** Fused multiply-add rounds once instead of twice and produces different results from the equivalent scalar sequence.
- **Never SIMD intrinsics.** Same reason, plus lane-order dependence.
- Write them as plain, readable loops. That is not a compromise — it is the specification.

**Not a hazard, and state it in a comment so nobody "fixes" it:** auto-vectorisation **cannot** reorder floating-point reductions without `-ffast-math`. `-O2` and `-O3` are safe. Nobody should be disabling optimisation here out of superstition.

`-fno-exceptions`-compatible, `noexcept` throughout, no allocation, no recursion, no VLAs. Every kernel takes explicit input/output/scratch spans.

**Tests:** per-kernel numerical cases with hand-computed expected values; shape-edge cases (kernel wider than input, stride > 1, single channel); and a test asserting that a Conv1D over a known input produces a bit-exact expected output, so a change in accumulation order fails immediately rather than at golden-vector time.

Part of #18.

ambroise-leclerc:
Dense, Conv1D, MaxPool1D, AvgPool1D, Flatten and the relu/sigmoid/softmax activations, as plain scalar loops in a fixed accumulation order. One module, imported by both the runtime and the host golden generator.

**Includes a hand-written `expF32()`, which the issue did not ask for.** `std::exp` cannot carry the cross-toolchain claim: neither the standard nor IEEE 754 requires a correctly-rounded `expf`, and glibc and the UCRT are different implementations — so a golden vector holding a softmax output could differ between the two CI legs *while both compilers were correct*. `expF32` uses only exactly-specified f32 operations and is a few ULP from `std::exp`, which is the wrong property to maximise here anyway. ADR-008 is amended to record it.

The accumulation-order spec is the one worth reading. Rather than freezing a bit pattern taken from the implementation — which would bless whatever the code happened to do — it picks a window where f32 addition is provably non-associative, states the value the documented order must give, and asserts the reverse order gives a different one.

Stacked on #57.

Closes #58.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,6 +72,7 @@ code*; treat this subsection as the direction that code is moving in.
 | `mdux.governance`, `mdux.governance.compliance` | `include/mdux/governance/` | `src/governance/` |
 | `mdux.shader.schema` (governed) | `include/mdux/shader/Schema.cppm` | `src/shader/Schema.cpp` |
 | `mdux.ml.schema` (governed) | `include/mdux/ml/Schema.cppm` | header-only |
+| `mdux.ml.kernels` (governed) | `include/mdux/ml/Kernels.cppm` | `src/ml/Kernels.cpp` |
 | `mdux.draw` (governed) | `include/mdux/draw/Draw.cppm` | `src/draw/Draw.cpp` |
 | `mdux.render.vulkan`, `mdux.render.offscreen` (adapter) | `include/mdux/render/` | `src/render/` |
 | `mdux.vulkansc.memory` | `include/mdux/vulkansc/MemoryPoolManager.cppm` | `src/vulkansc/MemoryPoolManager.cpp` |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,12 +157,14 @@ target_sources(MduXCore
             include/mdux/governance/Compliance.cppm
             include/mdux/shader/Schema.cppm
             include/mdux/ml/Schema.cppm
+            include/mdux/ml/Kernels.cppm
             include/mdux/draw/Draw.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
         src/evidence/Report.cpp
         src/shader/Schema.cpp
+        src/ml/Kernels.cpp
         src/draw/Draw.cpp
         src/governance/Governance.cpp
         src/governance/Justification.cpp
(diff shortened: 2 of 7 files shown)


0xfritz-dev (approved):
LGTM. Correct implementation of shared kernels with fixed accumulation order. Custom `expF32()` is necessary and well-justified — neither C++ standard nor IEEE 754 requires correctly-rounded `expf`, so glibc vs UCRT would diverge on softmax. Accumulation-order spec tests (reverse order differs) validates the contract. ADR-008 amended to record `expF32`.

ambroise-leclerc:
Four of the five addressed; one is a stack artefact.

**`std::floor` vs the "no libm" claim — fixed properly rather than softened.** You were right that the ADR sentence was false. Rather than weaken it, `expF32()` now does the floor by integer truncation with a correction step, so the claim is true as written. Verified mechanically:

```
$ nm -u -C build/CMakeFiles/MduXCore.dir/src/ml/Kernels.cpp.o
(no output)
```

Zero undefined symbols in the kernels object. The frozen determinism bit patterns are unchanged, which is the check that mattered — integer floor and `std::floor` agree exactly over this range.

**expF32's ±88 range claim — wrong, rewritten.** Your counterexample is exactly right: `{0, -100}` reaches -100 after the max subtraction. The comment now says the *lower* limit is reachable and the saturation is load-bearing (a flushed zero is the right probability *and* the right bit pattern on every toolchain, where a denormal would be neither), and that it's the *upper* limit that's unreachable.

**NaN comment — corrected.** It claimed the comparisons would return 0; in fact control would reach `static_cast<int>(floor(NaN))`, which is UB. Now says so, and notes the guard is a determinism decision rather than an optimisation.

**KernelTests header — corrected.** Now states that kernel cases are bit-exact and names the two scenarios that deliberately use a tolerance (`expF32` vs `std::exp`, softmax summing to 1), both being comparisons against a reference this code is explicitly *not* required to match bit for bit.

**FP-contraction flags on `MduXCore` — already handled, one PR up.** This is a stacked series; `mdux_enforce_fp_determinism()` lands in #146, which is the child of this PR. Correct observation about this commit in isolation, and the reason the two are adjacent in the stack.

Issue #61, opened 2026-07-26
S6 — tools/ml/GoldenGen, mdux-mlbake, and generated/model/ecg-demo/


ambroise-leclerc:
The baker that turns weights into a committed, byte-verified `ModelPackage`.

**`tools/ml/GoldenGen.cppm`** generates the golden input→output pairs by running the package **through `mdux.ml.kernels`** — the same governed module the device runtime will execute. Not a reimplementation, not a reference implementation: the same object code. That identity is what makes a golden-vector mismatch on device unambiguous evidence of an FPU or toolchain problem rather than a possible authoring bug.

Inputs should be deterministic and cover the input domain meaningfully: fixed patterns plus a seeded PRNG whose algorithm and seed are recorded in the recipe. **Never `<random>`'s default engine** — its output is not specified across standard library implementations, which would break byte-identity across the two CI legs.

**`mdux-mlbake`** follows E6/S5's contract exactly:

```sh
mdux-mlbake bake   recipes/model/ecg-demo.toml generated/model/ecg-demo/
mdux-mlbake verify recipes/model/ecg-demo.toml generated/model/ecg-demo/package.json \
                                               generated/model/ecg-demo/report.json
```

Outputs `package.json` (layers, golden vectors as `u32` bit patterns, `weightsDigest`, `maxScratchFloats`), `report.json`, and `weights.bin` as the sidecar.

**`generated/model/ecg-demo/`** — the committed demonstrator. Document its provenance in the recipe: where the weights came from, their license, and the explicit statement that they are a demonstrator with no clinical validation. This is the artifact whose replacement by a manufacturer's own weights is meant to require zero source changes; S9's weight-swap test proves it.

Part of #18.

ambroise-leclerc:
Wave 3 roadmap reconciliation: the former #12 blocker is complete. Use the implemented MduXCore/evidence interfaces, current BakeReport (semantic toolVersion and no commit SHA), shared diagnostic envelope, and #11 stack-safe integration policy. Cross-toolchain evidence currently covers MSVC and GCC 15/16; Clang is added only after #48 is green.

PR #146, opened 2026-08-03
feat(ml): enforce fp determinism and freeze the cross-toolchain contract

ambroise-leclerc:
Correct kernel source is necessary but not sufficient in C++, so this is the enforcement half of ADR-008 decision 3.

`mdux_enforce_fp_determinism()` sets `-ffp-contract=off -fno-fast-math -Wframe-larger-than=4096` on MduXCore. `mdux_verify_fp_determinism()` then walks the governed link graph, the directory options and `CMAKE_CXX_FLAGS` looking for `-ffast-math`, `-Ofast`, `-funsafe-math-optimizations`, `/fp:fast` and friends, and fails the configure step naming where the option came from. **Both paths were exercised before committing** — via `CMAKE_CXX_FLAGS`, and via an interface option on a target MduXCore links transitively.

**Deviation from the issue text worth a look:** #59 specifies `/fp:precise /fp:contract` for MSVC, but `/fp:contract` *enables* contraction into FMA — the exact transformation decision 3 forbids. Only `/fp:precise` is applied. I could not test this on MSVC from here.

The frozen bit patterns in `tests/ml/DeterminismTests.cpp` are the contract: not "the right answer", but what the specified accumulation order produces, so every toolchain must produce the same ones. CI runs them under the `determinism` label on both the MSVC and GCC legs. Weights come from an explicit LCG because `<random>` is not specified bit-for-bit across standard libraries.

Stacked on #58.

Refs #59 — the bake-reproducibility half lands with the baker in #61.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,16 @@ jobs:
       run: ctest --preset ninja-msvc -L evidence
       shell: cmd
 
+    # Floating-point determinism (ADR-008, issue #59). The frozen bit patterns in
+    # tests/ml/DeterminismTests.cpp are the contract, and this step is one half of what makes it a
+    # *cross-toolchain* contract - the Linux/GCC step is the other. Neither leg alone can falsify
+    # "these kernels compute the same bits everywhere", so do not reduce this to one.
+    #
+    # A failure here is never fixed by updating the constants. See that file's header comment.
+    - name: Verify ML determinism (MSVC)
+      run: ctest --preset ninja-msvc -L determinism
+      shell: cmd
+
     - name: Verify no source-tree writes
       shell: pwsh
       run: |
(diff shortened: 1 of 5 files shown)


0xfritz-dev (approved):
LGTM. Strong FP determinism enforcement at CMake level: `mdux_enforce_fp_determinism()` sets correct flags on MduXCore, `mdux_verify_fp_determinism()` scans governed link graph for forbidden flags (`-ffast-math`, `-Ofast`, `/fp:fast`, etc.) and fails configure with source location. Both paths exercised (CMAKE_CXX_FLAGS + transitive interface).

**Note**: MSVC deviation from #59 — `/fp:contract` enables FMA (forbidden), so only `/fp:precise` applied. Untested on MSVC from this environment. Should be validated on Windows CI leg before merge.

ambroise-leclerc:
**Option scan matching — real gap, fixed.** You're right that `STREQUAL` after stripping one genex layer only matched when the flag was the *entire* token. `SHELL:/fp:fast /Zi`, `$<$<COMPILE_LANGUAGE:CXX>:-ffast-math>`, a nested genex, or two flags in one quoted argument all slipped through — precisely the "arrived transitively from somewhere nobody was looking" cases the guard exists for.

Now substring matching. Both previously-missed forms verified to fire:

```
probe [SHELL:/fp:fast /Zi]                        -> caught
probe [$<$<COMPILE_LANGUAGE:CXX>:-ffast-math>]  -> caught
```

No false positive on `-fno-fast-math` (which this module itself sets) — it does not contain the string `-ffast-math`. Confirmed by a clean configure.

---

@0xfritz-dev on the MSVC note: it was validated on the Windows leg, and it **failed** — good call. `cl` defines `_M_FP_PRECISE` when `/fp:precise` is passed explicitly, the prebuilt `std` module is compiled without it, and importing `std` from a TU whose command line disagrees raises `C5050`; with warnings-as-errors that broke every governed module.

Resolved by passing **nothing** on MSVC. `/fp:precise` is already the default and VS2022 doesn't contract under it, so an unadorned build already behaves as ADR-008 requires. `/fp:contract` is now in the forbidden list instead, so enforcement on MSVC is entirely the guard's job. All six checks green on every PR in the stack.

Worth noting issue #59's specified flags are wrong in *both* directions: `/fp:contract` enables the fusion the ADR forbids, and `/fp:precise` breaks `import std`.

PR #147, opened 2026-08-03
feat(ml): add Classifier1D with its fail-closed golden self-test

### LINKED ISSUE #62: S7 — mdux.ml.runtime::Classifier1D with fail-closed self-test
`include/mdux/ml/Runtime.cppm` + `src/ml/Runtime.cpp`. **Governed zone.** The device-side engine: no file I/O, no dynamic graph construction, no allocation in `predict`.

```cpp
static std::expected<Classifier1D, MlError>
create(const ModelPackage& pkg,
       std::span<const std::byte> weights,   // caller-supplied: mmap, ROM, flash, or a linked blob
       std::span<float> scratch) noexcept;   // caller-supplied, >= pkg.maxScratchFloats

std::expected<void, MlError>
predict(std::span<const float> input, std::span<float> output) const noexcept;
```

**Weights are caller-supplied, deliberately.** A real model is megabytes; a multi-megabyte constexpr array takes minutes to compile and may exhaust MSVC. The runtime does no I/O — the caller mmaps, reads from ROM, or links a blob via `mdux_embed_blob()` (S9).

**`create()` sequence, in order:**

1. Validate `pkg.schemaVersion`.
2. Verify `sha256(weights) == pkg.weightsDigest` — E6/S2's digest, and the mechanism that makes "weights are data" safe.
3. Check `scratch.size() >= pkg.maxScratchFloats`.
4. **Re-run every golden vector through the real kernels and compare bit patterns.** Any divergence → return an error; the object is never constructed.

Step 4 is a genuine Class C safety control, not a unit test that happens to run late. It detects toolchain miscompilation, target floating-point drift, and a corrupted or mismatched package **before the device ever classifies a real signal**.

**`MlError` carries evidence, not just a code:**

```cpp
struct MlError {
    enum class Code : std::uint8_t { digestMismatch, schemaVersion, unsupportedLayer,
                                     shapeMismatch, scratchTooSmall, goldenMismatch };
    Code code;
    std::uint32_t layerIndex, goldenIndex, elementIndex;
    std::uint32_t expectedBits, actualBits;
};
```

That struct **is** the determinism evidence record when a device fails closed in the field. A bare enum would throw away exactly the information the incident report needs.

`Classifier1D` holds only spans and PODs; assert `std::is_trivially_destructible_v<Classifier1D>`.

Part of #18.

ambroise-leclerc:
The device-side engine. `create()` validates the package, verifies `sha256(weights)` against the digest it was baked with, checks the scratch budget, and then re-runs every golden vector through the real kernels comparing bit patterns. Any divergence returns an error and the object is never constructed.

Step 4 is the point of the whole subsystem: it detects toolchain miscompilation, target FP drift and a corrupted or mismatched package **before the device classifies anything real**, for bounded startup work and nothing per frame.

`MlError` carries the layer, golden and element indices plus both bit patterns, because when a device fails closed in the field that record is the evidence — and the divergence is by definition not reproducible on the bench, or CI would have caught it.

The self-test and `predict()` share `runFromScratch()`, so the self-test exercises the identical path rather than being evidence about a different one.

One implementation note: the weight blob arrives as bytes and is read as f32, so `create()` checks the blob's alignment and fails closed rather than assuming it.

Stacked on #59.

Closes #62.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,7 @@ code*; treat this subsection as the direction that code is moving in.
 | `mdux.shader.schema` (governed) | `include/mdux/shader/Schema.cppm` | `src/shader/Schema.cpp` |
 | `mdux.ml.schema` (governed) | `include/mdux/ml/Schema.cppm` | header-only |
 | `mdux.ml.kernels` (governed) | `include/mdux/ml/Kernels.cppm` | `src/ml/Kernels.cpp` |
+| `mdux.ml.runtime` (governed) | `include/mdux/ml/Runtime.cppm` | `src/ml/Runtime.cpp` |
 | `mdux.draw` (governed) | `include/mdux/draw/Draw.cppm` | `src/draw/Draw.cpp` |
 | `mdux.render.vulkan`, `mdux.render.offscreen` (adapter) | `include/mdux/render/` | `src/render/` |
 | `mdux.vulkansc.memory` | `include/mdux/vulkansc/MemoryPoolManager.cppm` | `src/vulkansc/MemoryPoolManager.cpp` |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,7 @@ target_sources(MduXCore
             include/mdux/shader/Schema.cppm
             include/mdux/ml/Schema.cppm
             include/mdux/ml/Kernels.cppm
+            include/mdux/ml/Runtime.cppm
             include/mdux/draw/Draw.cppm
     PRIVATE
         src/evidence/Digest.cpp
@@ -173,6 +174,7 @@ target_sources(MduXCore
         src/evidence/Report.cpp
         src/shader/Schema.cpp
         src/ml/Kernels.cpp
+        src/ml/Runtime.cpp
         src/draw/Draw.cpp
         src/governance/Governance.cpp
         src/governance/Justification.cpp
--- a/include/mdux/ml/Runtime.cppm
+++ b/include/mdux/ml/Runtime.cppm
@@ -1,0 +1,171 @@
+/**
+ * @file Runtime.cppm
+ * @brief Governed-zone ML runtime: the device-side classifier, and its fail-closed self-test.
+ *
+ * @compliance ADR-004 Trust zones in C++ (governed zone: std only, no Vulkan, no windowing)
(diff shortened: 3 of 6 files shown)


0xfritz-dev (approved):
LGTM. Core safety mechanism: `create()` validates package, verifies `sha256(weights)`, checks scratch budget, **re-runs all goldens through real kernels bitwise** — any divergence returns error, object never constructed. Self-test + `predict()` share `runFromScratch()` (identical path). `MlError` carries layer, indices, both bit patterns (field evidence). Alignment check on weight blob (fail-closed). Startup cost, zero per-frame cost — correct architecture.

ambroise-leclerc:
All six addressed — this was the most productive review of the stack.

**`WeightsUnaligned` — added.** Agreed that reusing `ShapeMismatch` conflated "a kernel rejected validated shapes" with "the caller handed us a blob off the f32 grid", which would mislead an incident log. New code, its own `describe()` message, and a scenario that builds a genuinely misaligned span (offset by one byte inside an oversized buffer) and asserts the specific code.

**`SchemaVersion` was unreachable — fixed.** `create()` now maps `SchemaError::UnsupportedSchemaVersion` to `MlError::Code::SchemaVersion` and leaves everything else as `SchemaInvalid`. It's the one rejection a caller can act on without reading the schema diagnostic — it means the artifact and the binary are different vintages. Test expectation updated.

**`layerIndex` was a count — fixed** to `maxSupportedLayers`, the first index that doesn't fit. The count is recoverable from the package; the index is the actionable half.

**`GoldenMismatch` placeholders — removed.** They contradicted the header's own "only meaningful for the codes that set them" rule. Left at defaults now, with a comment saying why.

**`maxLayers` / `maxSupportedLayers` desync — fixed.** The array is sized from `maxSupportedLayers` directly, so the capacity and the rejection threshold are the same constant.

**Empty goldens (raised on #144, enforced here).** `create()` now rejects a package with no goldens. The self-test loop over an empty set succeeded trivially, which meant the strongest control in the subsystem reported success without checking anything.

That change had a cost worth flagging: the test fixture used to generate its goldens by building a classifier with no goldens and predicting through it, which is no longer possible — and shouldn't be, since generating goldens is what happens *before* a package can be verified. It now runs the chain directly through `applyLayer()`, the same way `GoldenGen` does. Same for the no-heap harness in #148.

PR #148, opened 2026-08-03
feat(ml): verify no-heap-in-predict three independent ways

### LINKED ISSUE #63: S8 — No-heap-in-predict, verified three ways
"No allocation in `predict`" is easy to claim and easy to break silently. Three independent layers, all of them real checks rather than review discipline.

**Achieving it:** `predict()` operates entirely within the `scratch` span captured at `create()`, sized from `pkg.maxScratchFloats` (computed by the baker as the max over layers of the double-buffered activation footprint). No `std::vector`, no `new`, no recursion, no VLAs.

**1. Runtime interposition — the test that actually earns the claim.**
A test executable defining global `operator new`, `operator new[]`, `operator delete` and `operator delete[]` to increment a counter. Snapshot the counter, call `predict()` 1000 times, assert zero delta. Portable to MSVC, GCC and Clang, and it catches **everything** — including an allocation hidden inside a `std` call nobody expected to allocate.

**2. Object-file symbol scan — catches it at build time, in the PR that caused it.**
A ctest running `nm -C` (or `dumpbin /SYMBOLS` on MSVC) over the kernels and runtime object files, failing on any undefined reference to `operator new`, `_Znwm`, `_Znam`, `malloc`, `calloc`, `realloc` or `__cxa_throw`. Cheap, and it fails at the right time rather than at runtime.

**3. Static analysis and a stack bound.**
`cppcoreguidelines-no-malloc` and `cppcoreguidelines-owning-memory` as errors in `src/ml/.clang-tidy`, plus `-Wframe-larger-than=4096` — so "no heap" does not quietly become "enormous stack", which on a device is the same bug wearing a different hat.

Layer 2 also catches `__cxa_throw`, which double-checks E5/S2's no-throwing rule for the governed zone.

Part of #18.

ambroise-leclerc:
"No allocation in `predict()`" is easy to claim and easy to break silently, so none of the three layers is review discipline.

**Layer 1** replaces the whole global `operator new` family with counting versions and asserts a zero delta across 1000 predictions. It is also the layer most able to become worthless — a broken interposition would simply never move the counter and every case would pass — so the first scenario deliberately allocates and asserts the counter *does* move. The error path and `create()`'s self-test are measured separately, since a formatted diagnostic would creep into the error path first.

**Layer 2** scans the compiled ML objects with `nm`/`dumpbin` for undefined references to `operator new`, `malloc` and friends, plus `__cxa_throw`, which double-checks ADR-005's no-throwing rule. It fails in the PR that introduced the reference rather than whenever someone next runs that path. The scan refuses to run on an empty object list, and was checked against a known-allocating object to confirm it is not vacuous.

**Layer 3** is `src/ml/.clang-tidy` making `cppcoreguidelines-no-malloc` and `-owning-memory` errors.

Stacked on #62.

Closes #63.

### DIFF
--- a/cmake/MduXNoHeapScan.cmake
+++ b/cmake/MduXNoHeapScan.cmake
@@ -1,0 +1,116 @@
+# MduXNoHeapScan.cmake
+#
+# Layer 2 of issue #63: scan the compiled ML objects for any undefined reference to the allocation
+# and deallocation family, or to the exception-throwing runtime.
+#
+# The delete operators are in the forbidden set alongside the new ones deliberately. A reference to
+# `operator delete` means the object owns heap memory just as surely as a reference to `operator
+# new` does - often more visibly, because a destructor emitted for a std container is what pulls it
+# in. So a failure naming a delete symbol means the same thing as one naming new: something in the
+# governed ML zone acquired owning memory. Run as a ctest via `cmake -P`, not at configure time, because
+# the object files do not exist until after a build.
+#
+# Where layer 1 (tests/ml/NoHeapTests.cpp) proves predict() does not allocate *when run*, this
+# proves the kernels and the runtime cannot allocate *at all* - no path through them, taken or not,
+# reaches operator new or malloc. It is also far cheaper, and it fails in the PR that introduced
+# the reference rather than whenever someone next executes that path.
+#
+# __cxa_throw is in the forbidden set for a second reason: it double-checks ADR-005's no-throwing
+# rule for the governed zone. A std::vector::at() or a std::stoi() that crept in would show up here
+# as a throw reference even if it never allocated.
+#
+# Expected variables (passed with -D):
+#   MDUX_NOHEAP_OBJECT_LIST - path to a newline-separated file of object paths, written by
+#                             file(GENERATE) so that a generator expression resolves per config
+#   MDUX_NOHEAP_TOOL        - "nm" or "dumpbin"
+#   MDUX_NOHEAP_COMMAND     - path to that tool
+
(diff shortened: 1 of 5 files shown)


0xfritz-dev (approved):
LGTM. Defense-in-depth no-heap verification across 3 independent layers:
1. Runtime: global `operator new` interposition + counter, asserts zero delta over 1000 predictions (self-check allocates to prove counter works)
2. Static scan: `nm`/`dumpbin` for `operator new`, `malloc`, `__cxa_throw` refs in ML objects — fails at first introduction
3. Clang-tidy: `cppcoreguidelines-no-malloc` / `-owning-memory` as errors in `src/ml/`

Layer 1 could be defeated by broken interposition but self-check mitigates. Layer 2 catches what Layer 1 misses. Layer 3 is compile-time. Strong approach.

ambroise-leclerc:
All three addressed.

**Layer 1's scope was overstated — corrected.** You're right: it sees the global `operator new` family and nothing else. A direct `std::malloc`, a pool allocator, or anything bypassing `operator new` is invisible to it — which is exactly why layer 2's symbol scan exists and why the three layers aren't redundant. The file header and the `@brief` now say that rather than claiming it "catches everything".

**`leaked` — renamed to `probe`.** It was deleted two lines later; the name was just wrong.

**Scan docstring vs the delete family — corrected.** The comment now explains why the delete operators belong in the forbidden set: a reference to `operator delete` means owning heap memory just as surely as `operator new` does, often more visibly, since it's typically a std container's destructor that pulls it in.

Two other changes landed here from review of adjacent PRs: the harness now bakes real goldens (`create()` rejects an empty golden set as of #147), and the aligned-allocation path is now exercised — the counter-is-live scenario allocates an `alignas(64)` type and asserts both the count and the alignment, since nothing else in this binary would have touched that hand-written pointer arithmetic.

PR #149, opened 2026-08-03
feat(ml): add the safetensors reader and architecture validator

### LINKED ISSUE #60: S5 — tools/ml/Safetensors and ArchValidate
**Host tools zone** — never linked into a device target, never run in CI, never fed by a build step. A malformed or hostile input file therefore cannot affect a build.

**`tools/ml/Safetensors.cppm`.** The format is small enough to parse by hand and that is the point: a `u64` little-endian header length, a JSON header describing each tensor's name / dtype / shape / byte range, then raw tensor bytes. Parse it directly with E6/S3's JSON reader rather than adding a dedicated dependency — the same reasoning TrustSC applies, and the same reason the SOUP register (E3/S3) stays short.

Validate aggressively: header length within the file, every byte range in bounds and non-overlapping, declared shape consistent with the byte length, dtype recognised.

**`tools/ml/ArchValidate.cppm`.** Checks the recipe's declared architecture against the imported tensors before anything is baked. Rejections, all of them explicit errors rather than silent coercion:

- any dtype other than `f32` — v1 scope from S1
- any layer kind outside `{dense, conv1d, maxPool1d, avgPool1d, flatten}`
- rank > 3
- a weight tensor whose shape disagrees with the layer's declared in/out dimensions
- a missing bias where the layer declares one
- a `maxScratchFloats` that does not cover the computed worst-case activation footprint

Diagnostics use E13/S3's envelope, so `--format=json` gives an agent a machine-readable failure.

**Tests:** a fixture corpus of valid and deliberately malformed safetensors files — truncated header, out-of-bounds range, overlapping tensors, `f16` dtype, rank-4 tensor — each asserting the specific expected diagnostic code.

Part of #18.

ambroise-leclerc:
Host-tools zone: never linked into a device target, never run in CI, never fed by a build step, so a malformed or hostile weights file can only make a developer's mlbake invocation fail. The format is a `u64` header length, a JSON header and raw bytes — small enough to read with the evidence JSON parser, which is why it was chosen over anything needing a dependency.

**Two rejection families, distinguishable by code.** `malformed*` means the file is not a well-formed container; `unsupported*` means the container is fine and MduX v1 does not accept it — an f16 checkpoint or a rank-4 tensor. An author who sees `unsupportedDtype` needs to know their model is out of scope, not that their file is broken.

**ArchValidate deliberately restates none of the shape rules.** It resolves each layer's named tensors, records their shapes as the *file* declares them, packs the blob, and hands the result to `ModelPackage::validate()` — the same governed function the runtime calls. The shapes coming from the file while the dimensions come from the recipe is what makes that check mean something; filling both sides from the recipe would make it vacuous.

The corpus is built in memory rather than committed as binary fixtures, and every case asserts its specific diagnostic code.

Stacked on #63.

Closes #60.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -543,6 +543,36 @@ endif()
 
 mdux_discover_tests(ml_spec)
 
+# ML host-tools suite (issue #60)
+#
+# Links MduX::MlBakeLib, the host-tools-zone target. Deliberately separate from ml_spec, which
+# links MduX::Core only - keeping the two apart is what preserves ml_spec's standing evidence that
+# the governed ML modules reach nothing but std.
+add_executable(ml_tools_spec
+    ml/MlToolsSpecMain.cpp
+    ml/SafetensorsTests.cpp
+)
+
+target_link_libraries(ml_tools_spec PRIVATE MduX::MlBakeLib speclab::speclab)
+target_include_directories(ml_tools_spec PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+set_target_properties(ml_tools_spec
+    PROPERTIES
+        CXX_STANDARD 23
+        CXX_STANDARD_REQUIRED ON
+        CXX_EXTENSIONS OFF
+)
+
+if(TARGET __CMAKE::CXX23)
+    set_target_properties(ml_tools_spec PROPERTIES CXX_MODULE_STD ON)
+endif()
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    target_compile_options(ml_tools_spec PRIVATE /experimental:module /std:c++latest)
+endif()
+
+mdux_discover_tests(ml_tools_spec)
+
 # No-heap verification, layer 1 (issue #63): global operator new interposition.
 #
 # Its own binary because it replaces the global allocator - see tests/ml/NoHeapSpecMain.cpp.
--- a/tests/ml/MlToolsSpecMain.cpp
+++ b/tests/ml/MlToolsSpecMain.cpp
@@ -1,0 +1,20 @@
+/**
+ * @brief Entry point for the ML host-tools SpecLab executable (issue #60).
+ *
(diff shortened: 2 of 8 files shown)


0xfritz-dev (approved):
LGTM. Clean host-tools separation: never linked to device, never run in CI. Two rejection families distinguished by code: `malformed*` (bad container) vs `unsupported*` (valid container, out of scope — f16, rank-4) — operator-friendly. ArchValidate resolves named tensors from file, records shapes, packs blob, calls `ModelPackage::validate()` (same governed fn as runtime) — correct reuse. Corpus built in-memory with specific diagnostic assertions.

ambroise-leclerc:
All five addressed.

**"Never runs in CI" was simply false — corrected.** `ml_tools_spec` is part of the ordinary CTest suite, exactly as you say. The comment now draws the distinction it should have drawn from the start: this code is never linked into a device target and **no build step ever feeds it a weights file**; its own unit tests do run in CI, on fixtures this repository authored. Testing the parser is not the same as parsing untrusted input during a build, and the containment claim only ever rested on the second.

**Shape/byte-length overflow — fixed.** Right that it contradicted the file's own header about explicit overflow guards. `elementCount() * dtypeByteWidth()` is now computed with a saturating check, because a wrapped product can land exactly on the declared byte range and make a tensor describing gigabytes validate against a few bytes. New corpus case with three near-2^32 extents.

**`uint64` → `uint32` shape truncation in `appendTensor()` — fixed, both paths.** This was the nastier of the two: the appended bytes stay full length while the `TensorRef` describing them wraps, so the package validates against a blob it mis-describes. Now reports `mdux.ml.arch.shapeMismatch` naming the layer and the tensor instead of narrowing silently.

**Hardcoded header length `111` — fixed.** Now read back out of the file the way the parser does. A literal there would have stopped testing anything the moment the fixture's header string changed length.

The same overflow class turned up in `mdux.ml.schema` (raised on #144) and in `PackageLoad` (#151); all three are fixed.

Issue #64, opened 2026-07-26
S9 — mdux_embed_blob(), the ECG demonstrator, and a weight-swap test


ambroise-leclerc:
The end-to-end proof that "weights are data".

**`mdux_embed_blob(<target> FILE weights.bin SYMBOL ecgModelWeights)`** — a CMake helper generating a plain `.cpp` byte array plus an accessor returning `std::span<const std::byte>`. For demonstrators and for devices with no filesystem. Plain `.cpp`, not a module and not constexpr, so a multi-megabyte blob compiles in reasonable time.

**The demonstrator: a 1-D ECG rhythm classifier.**

```
Conv1D → MaxPool1D → Conv1D → Flatten → Dense → Softmax
```

It exercises **every** v1 kernel, produces a visible result, and makes the architecture tangible: its output drives a `StatusIndicator` in the E9 `.medui` screen, and it reads from the **same sample ring buffer** a `SignalTrace` (E11/S2) renders — a clean demonstration that the governed zone is self-consistent, with the classifier and the waveform provably looking at identical data.

**The weight-swap test — the one that proves the headline claim.**

Bake a second `ModelPackage` from different weights with the same architecture. Point the build at it. Assert:

1. The application source is **byte-identical** — no `#include`, no type, no call site changes.
2. Only `generated/model/<id>/package.json`, `report.json` and `weights.bin` differ.
3. `Classifier1D::create()` succeeds against the new package and its new golden vectors.
4. Swapping in the *old* `weights.bin` with the *new* `package.json` **fails closed** on `digestMismatch` — proving the digest check is load-bearing rather than decorative.

Point 4 is the one worth writing carefully. It is the difference between an integrity check and a comment claiming there is one.

Part of #18.

ambroise-leclerc:
Wave 3 roadmap reconciliation: the former #12 blocker is complete. Use the implemented MduXCore/evidence interfaces, current BakeReport (semantic toolVersion and no commit SHA), shared diagnostic envelope, and #11 stack-safe integration policy. Cross-toolchain evidence currently covers MSVC and GCC 15/16; Clang is added only after #48 is green.

PR #150, opened 2026-08-03
feat(ml): add mdux-mlbake and commit the ECG demonstrator package

### LINKED ISSUE #61: S6 — tools/ml/GoldenGen, mdux-mlbake, and generated/model/ecg-demo/

ambroise-leclerc:
The baker that turns weights into a byte-verified `ModelPackage`, following ADR-007's contract and the same `run()`/`write()`/`verify()` shape as `mdux-shaderbake`.

GoldenGen runs the goldens through `mdux.ml.kernels`, the same governed module the device executes — the same object code, which is what makes a golden mismatch on device unambiguous. Inputs are four fixed patterns plus an explicit LCG whose algorithm and seed are both recorded in the recipe and the report; `<random>` is not specified bit-for-bit across standard libraries, which would break byte-identity between the two CI legs.

`weights.bin` is repacked in layer order rather than being the safetensors file with its header removed, so two exporters emitting the same tensors in different orders produce the same blob.

⚠️ **The committed demonstrator's weights are synthetic.** Generated by a committed script from a seeded LCG, trained on nothing, **no clinical validity, none claimed**. The recipe says so at length. They exist to demonstrate the pipeline end to end and to be the "before" half of #64's weight-swap test. I could not obtain licensed pretrained ECG weights, and stating this plainly seemed better than implying otherwise.

This also completes the bake-reproducibility half of #59 that #61 was carrying: `evidence.model.ecg-demo` re-bakes and byte-compares on both CI legs.

Stacked on #60.

Closes #61.


0xfritz-dev (approved):
LGTM. Baker follows `run()`/`write()`/`verify()` shape (like `mdux-shaderbake`). GoldenGen runs goldens through `ml.kernels` (same object code as device — key property). LCG for inputs ensures bit-identity across CI legs. `weights.bin` repacked in layer order (exporter-order independent). **Weights are synthetic** — explicitly called out, generated from seeded LCG, trained on nothing, **no clinical validity claimed**. Completes bake-reproducibility half of #59.

ambroise-leclerc:
All three addressed.

**`generateGoldens()` empty-layer OOB — real bug, fixed.** Exactly as described: `requiredScratchFloats()` returns 0 for an empty span so the scratch check passes, and then `layers[0]` reads off the end. New `GoldenError::NoLayers`. Your framing is the right one — every current caller validates first, but a function that is only safe because of what its callers happen to do is one refactor from being unsafe.

**`requireUnsigned()` silent wrap — fixed.** `inputLength = -1` became 4294967295 and surfaced as a baffling layer-chain error; `maxScratchFloats = -1` would have asked the runtime for a 16 GB buffer. Now throws `TomlError` with the offending key's line, so the diagnostic points at the mistake. Added a `requireCount()` for the `size_t` case.

**Padded-header fixture — added, and you were right that it was missing.** The trimming shipped with no test at all, so nothing would have caught it accepting spaces but not NULs, or trimming into the JSON. Two cases now, one per padding byte, both asserting the file parses.

Re-baked and byte-compared after all three: `evidence.model.ecg-demo` unchanged.

Issue #15, opened 2026-07-26
Epic: .medui compiler and CMake build integration


ambroise-leclerc:
Adopt `.medui` as the single authored UI source. It compiles at build time into immutable packages; runtime code performs no DSL parsing, layout solving, or text shaping.

The evidence kernel (#12) and install/export foundation are implemented. Work remains dependent on #14 for text and #13 for visible output.

## Children

- [x] #190 — S1: ADRs for the deterministic `.medui` boundary and generated artifacts
- [x] #191 — S2: diagnostics and MDX-E code registry, using #19's shared envelope
- [x] #192 — S3: lexer/parser/AST and fixture corpus
- [x] #193 — S4: semantic analysis for themes and locales
- [x] #194 — S5: bounded layout
- [x] #195 — S6: text-budget checks
- [x] #196 — S7: safety-critical golden references
- [x] #197 — S8: canonical package and C++ emitters
- [x] #198 — S9: CMake/host-tool integration
- [x] #199 — S10: allocation-free screen runtime
- [x] #200 — S11: mdux-medui-check
- [x] #201 — S12: end-to-end windowed example

**Blocked by:** #14; #13 to render output. #12 and install/export are complete.
**Blocks:** #16 and #17.

PR #151, opened 2026-08-03
feat(ml): embed blobs, the ECG demonstrator, and the weight-swap proof

### LINKED ISSUE #64: S9 — mdux_embed_blob(), the ECG demonstrator, and a weight-swap test

ambroise-leclerc:
`mdux_embed_blob()` links a binary file into a target as a byte array plus a span accessor. Plain `.cpp`, not `constexpr` and not a module, because a multi-megabyte `constexpr` array takes minutes to compile and can exhaust MSVC. `alignas(16)` is load-bearing rather than decorative: `create()` reads the blob as f32 and refuses one that is not aligned, and a bare `unsigned char` array has alignment 1 — without it the demonstrator would fail closed on its own weights, intermittently, depending on section layout.

`EcgClassifierExample` opens no files at all — both `weights.bin` and `package.json` are linked in — and exercises the whole chain: digest check, four goldens re-run through the real kernels, then classification off a sample ring.

**The weight-swap test is the point of #64.** It uses the two *committed* packages rather than fixtures it built itself. It asserts they are architecturally identical, that the swapped one loads and reproduces its own goldens, and — the one worth writing carefully — that the old `weights.bin` against the new `package.json` fails closed on `digestMismatch`. Without that last one the digest field could be decorative and everything else would still pass. Verified by hand too: repointing the two embed lines at `ecg-demo-alt` loads the other model with `EcgClassifierExample.cpp` byte-identical.

**Two things #64 describes are not here**, because what they depend on is not built yet: the `.medui` `StatusIndicator` (#15) and the `SignalTrace` (E11/S2). The example notes both. It also parses the embedded `package.json` through the host-tools loader; a `constexpr` package emitter — the equivalent of the shader pipeline's #121 — would remove even that.

Final PR of the epic #18 stack.

Closes #64.


0xfritz-dev (approved):
LGTM. `mdux_embed_blob()` links binary as aligned byte array + span (plain `.cpp`, not `constexpr` — avoids MSVC compile-time explosion). `alignas(16)` is load-bearing: `create()` reads as f32, refuses unaligned. `EcgClassifierExample`: zero file I/O, both weights + package embedded.

**Weight-swap test (point of #64)** is excellent: uses two *committed* packages, asserts architectural identity, swapped loads own goldens, **old weights vs new package → `digestMismatch` fail-closed**. Without this last assertion, digest field could be decorative. Verified by hand: repointing embed lines at `ecg-demo-alt` loads other model with example byte-identical.

Notes missing `.medui` StatusIndicator (#15) and SignalTrace (E11/S2) — tracked separately.

ambroise-leclerc:
One fixed, one is a false positive.

**JSON `uint64` → `uint32` casts — fixed.** `inputLength`, `outputLength`, `maxScratchFloats` and the tensor shape extents all go through a range-checked `readUInt32()` now. Your reasoning is the right one: after a wrap, `validate()` is checking a different number than the file contained, so it passes on a package that describes something else entirely.

**Missing `make_directory` — declined; `file(WRITE)` creates parent directories.** That's documented CMake behaviour, and I checked it rather than assuming:

```
$ cmake -DOUT=/tmp/fwtest -P fw.cmake     # file(WRITE "${OUT}/deep/nested/file.txt" "hello")
$ ls /tmp/fwtest/deep/nested/file.txt
-rw-rw-r-- 1 leclerc leclerc 5 file.txt
```

It's also exercised on every clean build in CI — `EcgClassifierExample` links two embedded blobs and the Windows and Linux legs both build it from scratch. That said, "the output directory is never created" is a reasonable thing to think when reading the function, so I've added a comment recording the reliance.

PR #152, opened 2026-08-03
feat(ml): land the ML host tooling and demonstrator on develop (#60, #61, #64)

### LINKED ISSUE #59: S4 — Determinism enforcement and the cross-toolchain ctest

### LINKED ISSUE #60: S5 — tools/ml/Safetensors and ArchValidate

### LINKED ISSUE #61: S6 — tools/ml/GoldenGen, mdux-mlbake, and generated/model/ecg-demo/

### LINKED ISSUE #64: S9 — mdux_embed_blob(), the ECG demonstrator, and a weight-swap test

ambroise-leclerc:
## Why this PR exists

PRs #149, #150 and #151 were **merged into `feat/63-ml-no-heap` rather than into `develop`**, so a third of epic #18 is not actually on `develop` — no `tools/ml/`, no `recipes/model/`, no `generated/model/`, no demonstrator, no weight-swap test. Issues #60, #61 and #64 are correspondingly still open.

The cause is mechanical, not anyone's mistake: GitHub only re-targets a stacked PR when its base branch is **deleted**. `feat/63-ml-no-heap` was merged but kept, so the three children below it kept their original base and merged there.

This PR carries that work onto `develop`. Same commits, already reviewed and approved in #149/#150/#151 — nothing new.

## Contents

| Commit | Issue |
|---|---|
| `d9b800d` safetensors reader + architecture validator | #60 |
| `6c81ad3` `mdux-mlbake`, golden generation, committed `generated/model/ecg-demo/` | #61 |
| `dbb6a15` `mdux_embed_blob()`, ECG demonstrator, weight-swap proof | #64 |

## Verified before opening

Merged locally into `develop` and built clean with GCC 16.1: **360/360 tests pass**, including `ctest -L evidence` (both committed model packages re-bake byte-identically), `-L determinism` and `-L noheap`. The demonstrator runs and its four goldens reproduce bit for bit. No conflicts.

## Note on #59

Closing it here too. #146 said "Refs #59" rather than "Closes" because the bake-reproducibility half of that issue — re-baking the demonstrator and byte-comparing `package.json` — could not exist until the baker did. It lands with `6c81ad3` as `evidence.model.ecg-demo`, which runs on both CI legs.

Closes #60.
Closes #61.
Closes #64.
Closes #59.


ambroise-leclerc:
All three addressed in `11562f0`.

**`generate-alt-weights.py` pointing at the wrong script (both places) — fixed, and the second one mattered.**

You caught the consequence of how that file was made: I produced it by string-replacing the base script, and two references escaped. The usage snippet told the reader to run `generate-demonstrator-weights.py`, which regenerates the *other* file — so following the instructions silently does the wrong thing.

The suppressed comment about the `__metadata__` provenance is the more serious of the two, and I'd rather it hadn't been suppressed. That string is the artifact's own record of where it came from, embedded in the `.safetensors` itself, and it named a script that did not produce it. For a file whose entire justification is documented provenance, that is the one field that has to be right.

Fixing it changes the source bytes, so `ecg-demo-alt` was re-baked. The diff is one line:

```diff
   "path": "recipes/model/ecg-demo/ecg-demo-alt.safetensors",
-  "sha256": "50e5ecc6..."
+  "sha256": "6f679ec8..."
```

`package.json` and `weights.bin` are byte-identical, which is the useful confirmation: the change was confined to metadata strings and left every weight value untouched.

**Layer dimension arrays cast to uint32 unchecked — fixed.** Right, and I'd missed it: I range-checked the scalar TOML reads in an earlier round but not the `[layers]` arrays. `inLengths = [-1, ...]` became 4294967295 and surfaced much later as an incomprehensible layer-chain error. Now narrowed through a checked helper that names the layer and the field.

**`requiredScratchFloats()` truncation — fixed, and your reasoning about the symptom is exactly why it was worth fixing.** The wrapped value is *smaller*, so `validate()` then rejected the package as `ScratchTooSmall` — blaming a budget the baker had derived itself moments earlier, which sends the author looking in precisely the wrong place. Now reported explicitly as "the architecture is too large for the v1 package format".

Re-verified after all three: 360/360 tests pass under GCC 16.1, including `ctest -L evidence` (both committed packages re-bake byte-identically).

Issue #114, opened 2026-08-01
S8 — Verify IEC 81001-5-1 corpus mappings against an authorized source


ambroise-leclerc:
Part of #10 (Wave 3 documentation architecture). This is a Wave 3 closure blocker.

Verify the IEC 81001-5-1 corpus structure and mappings against a lawfully accessible, authorized copy of the edition identified by the corpus.

## Work

- Check every clause number, lifecycle-practice grouping, heading assignment, and cross-reference used by `docs/iec81001/`.
- Replace inferred practice placement with verified mappings while retaining conservative scoping where MduX has no mechanism.
- Update the README confidence statement with edition/source metadata, reviewer, and review date without quoting or closely paraphrasing normative text.
- Reassess the earlier JSON-export exclusion: enable it only if the verified corpus now meets the same confidence and validation bar as the other standards; otherwise document the concrete remaining reason.
- Regenerate and validate all resulting indexes/schemas/tests.

## Acceptance

- A maintainer records completion of the authorized-source review in the PR.
- Generated indexes are current and documentation/schema checks pass.
- No copyrighted standard text is uploaded or committed.

Issue #113, opened 2026-08-01
S7 — Verify IEC 62366-1 corpus mappings against an authorized source


ambroise-leclerc:
Part of #10 (Wave 3 documentation architecture). This is a Wave 3 closure blocker.

Verify the IEC 62366-1 corpus structure and mappings against a lawfully accessible, authorized copy of the edition identified by the corpus.

## Work

- Check every clause number, heading assignment, and cross-reference used by `docs/iec62366/`.
- Correct the corpus where the current professional-familiarity mapping differs from the authorized source.
- Update the README confidence statement to record what was verified, the edition/source metadata, reviewer, and review date without quoting or closely paraphrasing normative text.
- Regenerate and validate `AI-Reference.md` and `AI-Reference.json` and update affected schemas/tests.

## Acceptance

- A maintainer records completion of the authorized-source review in the PR.
- Generated indexes are current and documentation/schema checks pass.
- No copyrighted standard text is uploaded or committed.

ambroise-leclerc:
Done in #154 (`d50428b`). Verified clause by clause against IEC 62366-1 Edition 1.0 (2015-02), a licensed copy held by the maintainer — nothing from it committed, uploaded or quoted.

The corpus's caution was vindicated. It had withheld sub-clause numbers rather than assert unverified ones, and the check found three substantive errors:

1. **§4 was titled "General requirements".** The standard's §4 is *Principles*; "General requirements" is §4.1.
2. **§5.1 was "application specification"** — the superseded IEC 62366:2007 term, renamed *use specification* in 2015. It had propagated into the schema's `process_step` vocabulary too.
3. **§5.3, §5.4, §5.5 and §5.10 were missing entirely** — the hazard-identification steps that connect this process to ISO 14971, and *user interface of unknown provenance*, the clause an integrator is most likely to apply to MduX itself.

Issue #127, opened 2026-08-01
S9 — Retire the HTML/CSS runtime path


ambroise-leclerc:
Part of #13. Land last, after the replacement renderer and pixel test are green.

Delete the HTML/CSS runtime path: `MedicalUiContent`, reload/watcher types, `MedicalUiConfig`, and `MedicalUiRenderer` implementation. Retain the small facade/version/compliance surfaces explicitly listed by epic #13. Update examples, CMake, tests, public documentation, SDF descriptions, and agent guidance so nothing implies HTML/CSS remains supported.

Acceptance: no active source or documentation references the deleted API, install-tree consumption passes, and full CI is green after removal.

Issue #115, opened 2026-08-01
S9 — CI lint for internal documentation links and retired paths


ambroise-leclerc:
Part of #10 (Wave 3 documentation architecture) and the Wave 2 RETEX.

Wave 2 deleted or renamed documents while live README/agent references survived until the integration repair. Extend the documentation checks so this fails in the originating PR.

## Work

- Validate repository-relative Markdown link targets and local heading anchors across tracked live documentation.
- Read retired paths from `docs/governance/superseded-documents.md` and reject exact references to them outside that disposition record.
- Keep external HTTP links out of the deterministic check; network availability must not make CI flaky.
- Add fixtures for missing files, dead anchors, paths escaping the repository, a permitted disposition entry, and a forbidden live retired-path reference.
- Run the check in the existing Documentation Lint job.

## Acceptance

- The known Wave 2 dangling-reference cases fail the new tests.
- The current repository passes with zero findings.

ambroise-leclerc:
Done in #154 (`808361d`). Two checks added to `mdux-docs-lint`: repository-relative link targets and local heading anchors resolve, and paths retired by `superseded-documents.md` are rejected anywhere else. External links are deliberately not fetched — network availability must not decide whether CI passes.

The scan now includes README.md, AGENTS.md and CONTRIBUTING.md. It stopped at `docs/` before, which is exactly why the Wave 2 references in the root files survived the checks meant to catch them.

Switched on, it found three real defects of the shape it was written for: a link to `../iec62304/02-planning-and-requirements.md` that has never existed, and two corpus READMEs naming retired monoliths by path. 14 fixtures cover each failure mode; 105 tests pass and the repository lints clean across 76 files.

PR #154, opened 2026-08-04
docs: rebuild the top-level documentation from the built tree (#10)

### LINKED ISSUE #108: S1 — Re-baseline ADR numbering and supersede the Catch2 decision

ambroise-leclerc:
Seven of epic #10's nine children. **Two cannot be done by me** — see the end.

Every claim in these documents was checked against the build rather than written from memory, and the last commit adds a lint so the next one has to be.

| Commit | Issue |
|---|---|
| ADR re-baseline, Catch2 superseded | #108 |
| Retire the IEC 62304 monolith | #112 |
| `architecture.md` + `getting-started.md` | #109, #110 |
| README rewritten from built reality | #111 |
| AGENTS.md aligned | #65 |
| Internal-link and retired-path lint | #115 |

## Three things the issues asked for that the tree had already changed

**#108 specified ADR-008 for the testing decision. That number was taken** by epic #18's zero-SOUP ML ADR, which landed while #108 sat in the backlog. It is **ADR-009**.

**#109 asked for the current `MedicalUiRenderer` behaviour to be documented.** #127 deleted that path before this epic started. `architecture.md` records the deletion under a heading saying the HTML/CSS path is gone rather than planned.

**#111's list of fictional APIs to remove was mostly already removed** — they appeared as explicit denials. The real problem was that the README's *top half* still advertised a "Medical Device Software Manufacturer framework" with a Risk Analysis Engine, a CAPA system and post-market surveillance, while its bottom half correctly recorded that none of that code exists. A reader had no way to know which half to believe, and the overclaiming half reads first. That framing is gone; the file is 180 lines, down from 352.

## What writing the docs turned up

**The consumer snippet was wrong.** My first version omitted `CMAKE_EXPERIMENTAL_CXX_IMPORT_STD`, which must be set *before* `project()` — without it configure fails at generate time with "requires that the `__CMAKE::CXX23` target exist", which does not point back at the missing line. The corrected snippet was compiled and run verbatim against an install tree before being committed.

**SpecLab was an unregistered dependency.** Writing ADR-009 down made its absence from `soup-register.toml` a contradiction. It now has an entry: `tests` zone, no runtime deployment, pinned by commit because upstream has no tagged release.

**The lint found three real defects the moment it was switched on**, all of the shape it was written for: a link to `../iec62304/02-planning-and-requirements.md` that has never existed, and two corpus READMEs naming retired monoliths by path. All three fixed.

**Two dangling references in AGENTS.md** — `examples/BasicExample.cpp` described as an inactive file that is not in the tree, and a `sdf-documents` skill marked "Planned — `software_development_file/` doesn't exist yet" when it does.

## Verified

- **360/360 tests** pass under GCC 16.1 from a clean build
- **105 docs-lint tests** pass, 14 of them new
- `mdux-docs-lint` clean across **76 files**, up from 73 — the scan now includes README, AGENTS and CONTRIBUTING, which is exactly why the Wave 2 references in those files survived the checks meant to catch them
- Both new failure modes demonstrably rejected when re-introduced

## Not in this PR: #113 and #114

They require verifying IEC 62366-1 and IEC 81001-5-1 clause mappings against **lawfully accessible, authorized copies** of those standards, and both say a maintainer records the review in the PR.

I have no access to them and will not fabricate a review record — least of all in the epic whose purpose is documentation honesty. Both are marked closure blockers on #10, so **Wave 3 cannot close, and there is no v0.4.0, until someone with lawful access does them.**

Refs #10. Closes #108, #109, #110, #111, #112, #115, #65.


ambroise-leclerc:
## #113 done, #114 partially — with the sources now available

@ambroise-leclerc supplied a licensed IEC 62366-1 PDF and a draft of 81001-5-1. The two turned out to have very different standing, so they get very different outcomes.

### #113 — verified, and the corpus's caution was justified

Checked clause by clause against **IEC 62366-1 Edition 1.0 (2015-02)**, ISBN 978-2-8322-2281-2. Nothing from it is committed, uploaded or quoted — only the corpus's own clause numbers and headings changed.

The corpus had deliberately withheld sub-clause numbers rather than assert unverified ones. That was the right call: the check found three substantive errors.

1. **§4 was titled "General requirements".** The standard's §4 is **Principles**; "General requirements" is §4.1. Citing §4 under its sub-clause's name sends a reader to the wrong scope. Module renamed `02-principles.md`.
2. **§5.1 was "application specification"** — the superseded **IEC 62366:2007** term. The 2015 edition calls it *use specification*. It had also propagated into the schema's `process_step` vocabulary, corrected with it.
3. **§5.3, §5.4, §5.5 and §5.10 were missing entirely.** The first three are the hazard-identification and scenario-selection steps — how this process joins a device's ISO 14971 file. A description jumping from UI characteristics straight to UI specification omits how hazards enter at all. **§5.10 is *user interface of unknown provenance*, the clause an integrator is most likely to apply to MduX itself**, and it now says what MduX can and cannot offer against it.

A listed "step" that is not a sub-clause — *iterate the design as evaluation results require* — is folded into §5.8, where the standard puts iteration.

The §5.7 sub-divisions were verified but deliberately get **no** index rows: MduX has nothing to offer against them, and the generator's guard against rows pointing at stubs (#32) is right to refuse them.

### #114 — not closed, and should not be

The file supplied is **IEC/DIS 81001-5-1:2020(E), document 62D/1419/CDV** — a Committee Draft for Vote, not the published IEC 81001-5-1:2021 the corpus cites. It is also an **iTeh preview**: the contents pages covering §1 through §5.8.3 are absent; the first entry present is §5.8.4.

Confirmed against the draft: §1 Scope, §2 Normative references, §3 Terms and definitions — matching what the corpus asserts. **Not confirmed: §4's title.** The "general requirements" assumption stands unverified, because the pages that would show it are not in the preview.

Clause numbering changes between a CDV and the published standard, and this draft is both a draft *and* incomplete. Recording it as an authorized-source verification of the 2021 edition would be exactly the overclaim this epic exists to remove — so the README carries it as an explicitly-labelled **partial** review, and **#114 stays open**. Finishing it needs the paper copy you hold, to confirm §4 and add practice-level numbering (#31 tracks the same work).

One thing worth having even from the draft: the security practices sit far deeper than this directory's filenames suggest — development §5, maintenance §6, **security risk management §7**, configuration management §8, problem resolution §9, QMS §10. The `02-`, `03-`, `04-` prefixes are file ordering, not clause numbers. No module ever claimed otherwise, but it is close enough to mislead a skimmer, so it is now stated.

### Verified

360/360 tests, 105 docs-lint tests, `mdux-docs-lint` clean across 76 files, generated indexes current, schema drift clean.

Updates the PR scope: **Closes #113** as well. #114 remains open.

ambroise-leclerc:
## #114 closed — the official preview was enough

That file is the **published** standard: **IEC 81001-5-1, Edition 1.0, 2021-12**, an official IEC preview carrying the full bilingual table of contents. That is sufficient to check every clause number and heading this directory asserts. It also carries interpretation sheet **ISH1:2025**, now recorded alongside it.

**All confirmed as asserted:** §1 Scope, §2 Normative references, §3 Terms and definitions, and §4 **General requirements** — the last of which the corpus had been carrying as an *assumption* from IEC front-matter convention. The assumption was right.

### Declining the draft was not pedantry

Comparing the CDV against the published standard shows the numbering moved by a **whole clause level**:

| Subject | CDV draft | Published |
|---|---|---|
| Security risk management | §7 | **§4.2** |
| Quality management | §10 | **§4.1** |

Anything numbered from that draft would have been wrong. The comparison stays in the README so the next reader knows why a draft was refused rather than quietly used.

### Added: a verified clause map

The practice modules are organised by topic, and their `02-`/`03-`/`04-` prefixes are file ordering — close enough to clause numbers to mislead. The map now states where each actually sits:

| Module | Published clauses |
|---|---|
| `01-scope-and-relationship-to-iec62304.md` | §1, §2, §3 |
| `02-security-risk-management.md` | §4.2, §7 (§7.1–§7.5) |
| `03-secure-design-and-implementation.md` | §5.3, §5.4, §5.5 |
| `04-security-verification-and-update-management.md` | §5.7, §5.8, §6, §9 |

Still **not** asserted: requirement-level numbering inside the practice modules — a heading reads "Secure design", not "§5.4.2 Secure design". That is **#31**, which is now *unblocked* rather than waiting on access.

### Epic #10 is complete

All nine children. **Closes #114** in addition to the eight already listed.

Verified: 360/360 tests, 105 docs-lint tests, lint clean across 76 files, generated indexes current, schema drift clean.

With #10 done, **Wave 3 closes** — #10, #13 and #18 all delivered. That clears the way for **v0.4.0** whenever you want it cut.

PR #156, opened 2026-08-04
docs(81001): carry verified clause numbers into the practice modules

ambroise-leclerc:
The last item before v0.4.0.

The corpus already knew which clauses each module covered — the README's clause map recorded it — but the headings inside still read "Secure design" and "Security update management". Anyone citing one of those had a topic name rather than a clause, which is precisely what kept this corpus from being usable compliance-facing.

Every section heading now carries its clause:

| Module | Clauses |
|---|---|
| `02-security-risk-management.md` | §4.2, §7.1–§7.3, §7.4, §7.5 |
| `03-secure-design-and-implementation.md` | §5.3–§5.4, §5.5 |
| `04-security-verification-and-update-management.md` | §5.7, §9, §6, §5.8 |

All taken from the structure verified against IEC 81001-5-1 Edition 1.0 (2021-12) under #114.

## Where numbering stops, and why that is a decision

At the level where MduX has something to say — §7.4 rather than §7.4.1, §5.7 rather than §5.7.3 — and the README states this rather than leaving it to read as an omission.

Splitting §5.7 into five sub-clause headings so the corpus could write "no MduX mechanism" five times would be padding rather than precision. The index generator's guard against rows pointing at stubs (#32) would refuse them anyway, and it would be right to.

## A stale pointer removed

Two places cited **issue #31** as though it tracked this work. #31 is **closed** — it was the issue that *created* this corpus during Wave 2, and its only numbering instruction was to carry a "numbering is provisional" caveat. The authorized-source verification has replaced that caveat with fact, so the pointer was both stale and pointing at the wrong thing.

No open issue tracked this work; it is done here.

## Verified

360/360 tests, 105 docs-lint tests, `mdux-docs-lint` clean across 76 files, generated index current and now carrying clause numbers in its Clause column, schema drift clean.


ambroise-leclerc:
Both addressed in `cb38f7f`.

**The stale README sections — a bad miss, and the right thing to flag.** Its Structure table still listed practice names with no clauses, and the AI-Reference description still claimed *"every practice row shows `—`, because that is what this corpus asserts… it will gain numbers when the modules do"* — a sentence this very PR falsified. A file contradicting itself about whether its own clause numbers exist is precisely the failure this epic was opened to remove, so that was the wrong place to leave one.

Both sections now describe what the corpus is. The two rows that legitimately show `—` are named rather than left looking like gaps: the IEC 62304 relationship mapping, which is a relationship *between* standards rather than a clause of this one, and the section recording what no MduX mechanism covers.

**Module 04's ordering — reordered to §5.7, §5.8, §6, §9.** You're right that it was invisible while the headings were topic names and obvious once numbered, and that it contradicted the order in the file's own title. The opening sentence already ran verification → handling → deployment, so the narrative now matches the numbering instead of fighting it. Index regenerated.

**One thing your comment surfaced that I've stated rather than fixed.** Reordering module 04 makes its rows monotonic, but the index as a whole still is not: it follows *module* order, which is thematic, so §7 precedes §5 across files because risk management is discussed before secure design.

I kept that. The reading order is the useful one for someone working out what MduX does and does not provide, and renaming modules to force global clause order would churn every path for a property a reader can get by sorting the clause column. But it is now **said** in the README rather than left to be discovered — within a module, clause order; across modules, thematic.

Verified: 360/360 tests, 105 docs-lint tests, lint clean across 76 files, index current, schema drift clean.

Issue #162, opened 2026-08-06
S6 — Coverage draw path and text pixel tests


ambroise-leclerc:
Part of #14. Depends on the S4 atlas baker (#160) and S3 rasteriser (#159) children; uses the #13 draw path.

Wire the baked R8 atlas into `UiRenderer`'s existing single-texture binding (#13 already created the atlas/sampler mechanism — this child replaces its contents, not the mechanism) and add a `DrawList` helper that records `CoverageR8` rectangles from a baked glyph run. Add the text pixel tests under `tests/render/` reusing `HeadlessDevice.hpp` and `PixelExpectation.hpp`: render known static text offscreen on lavapipe and compare exact bounds and coverage steps. Bounds and colour checks are exercisable before any glyph exists; localized-text presence follows.

Acceptance: the positive case passes in Linux CI with the `pixel` label; deliberate one-pixel or coverage changes fail with coordinates and expected/actual values; no golden is silently rewritten.

Issue #160, opened 2026-08-06
S4 — Atlas packer and font baker


ambroise-leclerc:
Part of #14. Depends on the S3 rasteriser child (#159).

Pack rasterised glyphs into a power-of-two R8 atlas with deterministic shelf or skyline placement, and complete `mdux-textbake` so it produces a committed `generated/font/<id>/` artifact through `mdux_bake_artifact()`: `package.json`, `report.json`, `atlas.bin`. The recipe uses parallel-array-of-scalars tables (the `mdux.tools.toml` subset supports no `[[tables]]`). A first real font is committed in this child; the recipe lists its supported locales and the atlas dimensions are recorded in the resolved options.

Acceptance: `mdux-bake-update` produces the reviewed artifacts; `ctest -L evidence` is byte-identical on MSVC and GCC; atlas dimensions are power-of-two and the packer rejects an over-budget glyph set with a stable code; no normal build writes a tracked file.

ambroise-leclerc:
Landed on `develop` as three stacked PRs:

- **#173** — `cmap` + `hmtx` in `mdux.tools.truetype`. A prerequisite this issue did not name: without a code-point-to-glyph mapping a recipe cannot express a charset, and without advance widths a baked glyph cannot carry the advance ADR-010 requires.
- **#174** — `mdux.tools.atlaspacker`, deterministic shelf packing into a power-of-two sheet.
- **#176** — the baker, the recipe format, seven `TXT` diagnostics, and the first committed font package.

Acceptance:

| | |
|---|---|
| `mdux-bake-update` produces the reviewed artifacts | yes — `generated/font/dejavu-ui/` |
| `ctest -L evidence` byte-identical on MSVC and GCC | yes — the atlas baked on GCC re-bakes byte-identically under MSVC |
| atlas dimensions power-of-two, over-budget set refused with a stable code | yes — 128x128; `AtlasBudgetExceeded` / `GlyphTooLarge` |
| no normal build writes a tracked file | yes |

The committed package is 95 printable-ASCII glyphs from DejaVu Sans at 16 px/em, on a 128x128 sheet at 54% occupancy. The font is redistributed unmodified under the Bitstream Vera + Arev terms and registered in `docs/governance/soup-register.toml`.

S5 (#161) continues from here.

Issue #159, opened 2026-08-06
S3 — Deterministic rasteriser with coverage AA


ambroise-leclerc:
Part of #14. Depends on the S2 TrueType parser child (#158).

Rasterise parsed `glyf` outlines to an R8 coverage atlas using deterministic coverage anti-aliasing. Edge accumulation and scan conversion are integer-only or use float operations encoded as `u32` bit patterns where determinism requires it (ADR-007); no `std::fma`, no operation whose result differs between MSVC and libc/UCRT. The rasteriser is governed code importable by the host baker; if it is ever shared with a device path, that is one module (ADR-008 decision 1).

Acceptance: the same glyph rasterises byte-identically on MSVC and GCC; `ctest -L evidence` and `-L determinism` pass on both legs; coverage values are in [0,255] with no clamping surprises; rasteriser unit tests cover corners, curves and edge cases.

ambroise-leclerc:
Landed on `develop`.

- #158 (S2, hand-parsed TrueType `glyf`) merged via #171, commit `c3c82ed` then `774b3f7`.
- #159 (S3, deterministic rasteriser with coverage AA) merged via #172, commit `b63f534`.

Both are built and tested on the MSVC and GCC 16 legs. Closing; S4 (#160) continues from here.

Issue #158, opened 2026-08-06
S2 — Hand-parsed TrueType (glyf only)


ambroise-leclerc:
Part of #14. Depends on the S1 schema/baker child (#157).

Add a host-only TrueType parser that walks the `glyf` table only, deliberately not linking `freetype` or any other font SOUP — the parallel of `tools/shader/Spirv.cpp` for SPIR-V and `tools/ml/Safetensors.cpp` for safetensors (ADR-008 decision 6). v1 supports TrueType `glyf` outlines and Latin/Cyrillic/Greek LTR only; composite glyphs, CFF/CFF2, hinting, GPOS, ligatures and RTL/complex scripts are rejected with stable codes — they never reach the runtime. The parser returns `Result<…, ParseError>` and the baker converts each failure to a shared-envelope `cli::Diagnostic`.

Acceptance: malformed/truncated/unsupported font fixtures fail with stable codes; the baker links no third-party font library; `mdux_verify_trust_zones()` still passes; parser unit tests cover `glyf` simple and composite outlines, and every rejection path.

ambroise-leclerc:
Landed on `develop`.

- #158 (S2, hand-parsed TrueType `glyf`) merged via #171, commit `c3c82ed` then `774b3f7`.
- #159 (S3, deterministic rasteriser with coverage AA) merged via #172, commit `b63f534`.

Both are built and tested on the MSVC and GCC 16 legs. Closing; S4 (#160) continues from here.

PR #163, opened 2026-08-06
feat(text): mdux.text.schema and mdux-textbake skeleton (S1, #157)

### LINKED ISSUE #157: S1 — mdux.text.schema and mdux-textbake skeleton
Part of #14. The evidence kernel (#12) is complete and the draw path (#13) is closed; this is the wave's opening move.

Introduce the governed `mdux.text.schema` module as the canonical shape of a baked text package — positioned glyph runs per approved locale, no runtime shaping — and a host-only `mdux-textbake` baker skeleton linked through `MduX::ToolsCommon`. The schema is imported by both the device runtime and the host baker: one definition, one set of compile flags (ADR-008 decision 1, applied to text). The baker has one code path for bake and verify (ADR-007). No artifact is committed in this child; the first committed text package lands in S4. Register the module in `MduXCore`'s `FILE_SET CXX_MODULES` and the baker in `tools/CMakeLists.txt` so `mdux_verify_trust_zones()` and `ctest -L evidence` still pass.

Acceptance: schema `validate()` rejects malformed packages with stable codes; the governed-zone check passes after registration; `text_spec` schema tests pass on MSVC and GCC; no normal build writes a tracked file.

ambroise-leclerc:
Closes #157. Opens Wave 4 of epic #14 (Font & text pipeline).

## What lands

S1 — the governed `mdux.text.schema` module and the `mdux-textbake` host baker skeleton — together with ADR-010, which records the architectural commitment the rest of the wave leans on.

- **`include/mdux/text/Schema.cppm` + `src/text/Schema.cpp`** — governed, ADR-004/005/007/010 annotated. `TextPackage` is positioned glyph runs per locale against a font-package atlas id; the runtime consumes runs, it does not shape them (ADR-010). `PackageView` is the non-owning `constexpr` form for generated code, mirroring `mdux.shader.schema`'s owning/view pair.
- **`tools/text/{TextBake.cppm, TextBake.cpp, TextBakeMain.cpp}`** — host baker skeleton linked through `MduX::ToolsCommon`, one code path for bake and verify (ADR-007). `run()` produces a no-run package at S1; S4 (#160) commits the first real artifact once the TrueType parser (#158) and rasteriser (#159) land.
- **`tests/text/{TextSpecMain.cpp, SchemaTests.cpp}`** — 17 SpecLab scenarios covering `validate`/`parse`/`write` round-trip and every `SchemaError` rejection case. Links `MduX::Core` only, mirroring `shader_spec` and `ml_spec`.
- **ADR-010** records the no-on-device-shaping commitment, inherited from ADR-008 decision 1 (host/device schema sharing). Indexed in `docs/adr/README.md`.

No artifact is committed (no `generated/` or `recipes/` call site); the first baked text package is S4 (#160). `mdux_verify_trust_zones()` passes with the new module registered; `ctest -L evidence` stays clean.

## Verification

`cmake --build build --target MduXCore mdux-textbake text_spec` green. Full evidence/spec suite:

```
ctest --test-dir build -R '^(text_spec|shader_spec|ml_spec|draw_spec|evidence|tools_spec|bridge_spec)'
100% tests passed, 0 tests failed out of 314
```

Trust-zone check at configure time:

```
mdux_verify_trust_zones: 2 governed target(s) clean of Vulkan/windowing dependencies
```

## Out of scope for this PR

- **S2 (#158)** — host TrueType parser. Starts in a follow-up branch `14-truetype-parser` once this merges; per the #13 one-PR-per-child pattern.
- **The pre-existing adapter ICE** on `VulkanRenderer.cpp` and `VulkanSCTriangleExample` under GCC 16.0.1 is unrelated and tracked under issue #48. It reproduces on `develop` with this branch's edits stashed; CI's `gcc:16` container is unaffected. S1 touches only `MduXCore` and the host-tools zone, neither of which ICE.
- Wave 6 child issue promotion (#14 → epic body update) was done in the same session: see the #14 epic body where S1–S6 are now linked to their child issue numbers (#157–#162).


devjean-75:
## Review — S1 text schema + textbake skeleton

Reviewed the diff and surrounding code; I did **not** build or run the suite, so the green results in the description are yours, not independently verified here.

### Overview

Lands Wave 4 S1 of epic #14: the governed `mdux.text.schema` module (owning `TextPackage` + non-owning `PackageView`), the `mdux-textbake` host baker skeleton, 17 SpecLab scenarios, and ADR-010. No artifact committed; `run()` deliberately produces a no-run package until S4.

What's genuinely well done: the CMake blocks are a faithful match to the `MduXShaderBakeLib`/`MduXMlBakeLib` conventions (module file sets, `MduX_warnings`, `CXX_MODULE_STD`, MSVC flags) — `MDUX_TOOL_VERSION` correctly arrives via `MduX::ToolsCommon`'s PUBLIC link to `MduX::BuildInfo`, so no redefinition was needed. The bounds check in `validate()` (`byteLength > sidecarByteLength - byteOffset`) is overflow-safe by construction rather than by hope. `verify()` re-comparing the sidecar bytes rather than trusting the package's recorded digest is the right instinct, and the comment explaining why is correct. ADR-010 follows `template.md` section-for-section. And the `findLocatesRun` test comment identifying a latent dangling-pointer read in the *shader* schema's equivalent test is a real catch.

---

### 1. `recordSizeWire` is never emitted — its doc comment claims otherwise

`include/mdux/text/Schema.cppm:103`. The comment says it is *"recorded in `package.json` so an auditor can read the stride without consulting this module."* `toJson()` writes `schemaVersion`/`id`/`kind`/`atlas`/`locale`/`sidecar`/`runs` — never `recordSize`. The constant's only consumer is a test asserting it equals `"6"`. It's also a *string* `"6"`, where a JSON member would be the unsigned number `6`, which suggests it was drafted for something else.

Unlike `stageWireValues` in the shader schema, this isn't a wire encoding — nothing encodes or decodes through it. Either emit it or delete it and fix the comment.

**This is the cheapest moment to decide.** No artifact is committed until S4 (#160). Adding a member to `package.json` after that changes the digest of every committed text package and every `report.json` that records it.

### 2. `MduXTextBakeLib` has zero test coverage

`tests/CMakeLists.txt:558` — `text_spec` links `MduX::Core` only. Both siblings deliberately do the opposite:

- `shader_spec` → `MduX::ShaderBakeLib` (comment at `tests/CMakeLists.txt:105`: *"Links MduX::ShaderBakeLib rather than the executable, so the scenarios drive run()/write()/verify()"*)
- `ml_tools_spec` → `MduX::MlBakeLib`

So ~300 lines of new baker code ship untested: `parseRecipe()`'s four diagnostic paths, `run()`, `write()`, `verify()`, `readFile()`'s chunked-read loop, and `compareArtifact()`'s first-differing-byte reporting — the last being exactly the kind of index arithmetic that's wrong once and then wrong quietly forever. `TextBake.cppm` itself argues the library exists *so it can be tested* (*"everything worth testing is in the library"*), and then nothing tests it.

A `text_tools_spec` target mirroring `ml_tools_spec` would close this. Even at skeleton stage, `run()` on a valid recipe → `parse()` the resulting `packageJson` round trip is a few lines and pins the S1 contract before S4 changes it.

### 3. `PackageView` / `RunView` are untested

`include/mdux/text/Schema.cppm:186-215`. No scenario constructs either. `runBytes()` has a real bounds guard whose own comment says it exists because *"a view can also be assembled by hand in a test, and a span past the end of the sidecar is the one mistake here that would not fail visibly"* — and there is no such test. This is the type the device runtime consumes, so it's the half of the module that matters most at runtime.

### 4. ADR-010 Decision ¶1 contradicts the code it governs

> *"A font baker (`mdux-textbake`, child #157 / #160) produces a committed `generated/font/<id>/` artifact ... an R8 coverage atlas, glyph advances, kerning pairs ... and a restricted-charset table"*

The code says the opposite: `packageKind = "text"`, `run()` sets `header.kind = text`, so the artifact is `generated/text/<id>/`, and `Schema.cppm:47-56` is explicit that the atlas lives in a *separate* font package (`mdux.font.schema`, #161) which a text package **references by id** rather than embedding — with two well-argued reasons for that split. The Consequences section repeats the error (*"auditable from `git log -- generated/font/`"*), as does Decision ¶4 (*"fail the font baker"*).

This matters more than a stray comment: ADR-010 is marked Accepted and is cited by `@compliance` blocks in five new files as the architectural basis for the whole wave. Worth reconciling before S2 builds on it.

### 5. `docs/architecture.md` not updated

This PR adds `include/mdux/text/Schema.cppm` to `MduXCore`'s `target_sources` and a new host-tools library, but three inventories go stale:

- Line 23 — trust-zone table, host-tools row lists `MduXToolsCommon, MduXShaderBakeLib, MduXMlBakeLib`
- Line ~54 — governed `MduXCore` module table (needs a `mdux.text.schema` row)
- Line ~86 — host-tools table (needs `MduXTextBakeLib` | `tools/text/` | `mdux-textbake`)

The `generated/` artifact table at line 122 correctly needs nothing yet.

### 6. `TXT000` comment contradicts the code in the same PR

`tools/text/TextBake.cpp:39` reserves it: *"TXT000 stays free as a top-of-file marker a future caller may want; the first numbered code begins at TXT001."* `tools/text/TextBakeMain.cpp:33` then uses `TXT000` for "cannot read recipe". Since these codes are documented as stable API that agents key off, pick one — either document `TXT000` as "recipe unreadable" or renumber it to `TXT009`.

### 7. Recipe-level validation is asymmetric

`tools/text/TextBake.cpp:126-138`. Empty `atlas` and `locale` get dedicated diagnostics with fix hints (TXT003/TXT004). Empty `id`, and a `sidecar` containing a path separator, get no recipe-level check — they fall through to schema validation and surface as the generic TXT005 *"assembled package is not valid: …"*, pointing at the recipe with line 0 and no fix hint. Both are recipe-authoring mistakes and cheap to catch where the other two are caught.

---

### Minor / nits

- **`byteEnd()` on unvalidated runs** (`src/text/Schema.cpp:172-178`): the pairwise loop calls `runs[j].byteEnd()` before `runs[j]` has been bounds-checked, so the doc comment's *"never overflows for a validated package"* doesn't cover this call site. Not a bug — unsigned wraparound is well-defined, and any wrapping run necessarily fails its own bounds check when the outer loop reaches it, so no invalid package escapes. But the reasoning is load-bearing and undocumented; hoisting the per-run bounds check into a first pass before the pairwise loop would make it airtight for an O(n) cost.
- **`UnsupportedSchemaVersion` is unreachable**: `header.validate()` checks `schemaVersion` first and maps to `ReportRejected`, so the branch in `TextPackage::validate()` is dead. The test comment already documents this, and it mirrors `src/shader/Schema.cpp:230` exactly — so it's consistent, pre-existing, and fine to leave. Fix both or neither.
- **`RunView` uses `std::size_t` where `TextRun` uses `std::uint64_t`**: no conversion path exists in this PR, but the #15 emitter will need one, and it narrows on a 32-bit host.
- **ADR-010 typo**: "The architecturalhazards" (missing space), IEC 62304 §5.1.2 bullet.
- **S-number drift**: `Schema.cppm:19` attributes the coverage draw path to *"S5 of #14/#161"*, while `tests/CMakeLists.txt` and `Schema.cppm:40` put it at #162/S6 and #161 at font schema.

### Security & performance

Nothing concerning. No new attack surface in the governed zone — no Vulkan types, no heap growth at runtime, no floats (so ADR-007's `{"bits": N}` encoding correctly never arises). Parsing happens only in the host-tools zone where throwing is permitted. `validate()`'s O(n²) pairwise loop and `find()`'s linear scan are both correct choices at the stated scale and are justified in comments.

### Verdict

Solid, conventional, well-argued work — the module design and the host/device sharing rationale are right, and the CMake integration is exact. I'd want **#1** (decide `recordSizeWire` before any artifact is committed) and **#2** (baker test coverage, matching both siblings) addressed before merge, and **#4/#5** as doc corrections that get more expensive once S2 builds on them. The rest are follow-ups.

ambroise-leclerc:
Addressed the seven substantive items and the nits in `3e78e09`. Build green, tests green.

| Item | Fix |
|---|---|
| 1. `recordSizeWire` never emitted | Deleted the constant; the stride is pinned by `schemaVersion` and not serialised. Test assertion removed. |
| 2. `MduXTextBakeLib` untested | Added `tests/text/TextToolsSpecMain.cpp` + `TextBakeTests.cpp` (9 SpecLab scenarios) and a `text_tools_spec` target mirroring `ml_tools_spec`/`shader_spec`. Covers `parseRecipe` (valid + TXT001/002/003/004/009/010), `run()` round-trip through `TextPackage::parse()`, `write()` into a temp dir, and `verify()` TXT007/TXT008/identical paths. |
| 3. `PackageView`/`RunView` untested | Added four governed scenarios to `text_spec` covering `runBytes()` happy path and both out-of-bounds guards (length past end, offset past end). The view bytes-content path is checked against byte values, not just length. |
| 4. ADR-010 contradicts code | Decision ¶1 now distinguishes the text baker (`generated/text/<id>/`, #157) from the font baker (`generated/font/<id>/`, #160); Consequences lists both git-log paths; Decision ¶5 names which baker fails for which case; fixed the "architecturalhazards" typo. |
| 5. `docs/architecture.md` stale | Added `mdux.text.schema` to the governed module table, `MduXTextBakeLib` to the host-tools row (line 23) and table, and updated the #14 Planned-not-built entry to note S1 has landed. |
| 6. `TXT000` comment contradicts code | Removed the reserving comment. `TXT000` is documented as the recipe-unreadable code, matching `SHB000` in the shader baker and what `TextBakeMain.cpp` already emits. |
| 7. Recipe-level validation asymmetric | Added `TXT009` (empty `id`) and `TXT010` (sidecar path separator) checks in `parseRecipe` so those recipe-authoring mistakes are diagnosed with a fix hint at the recipe layer rather than falling through to schema validation as a generic `packageInvalid` at line 0. |

Nits taken:
- `validate()` restructured: per-run bounds pass before the pairwise overlap pass, so `byteEnd()` is only ever called on runs whose bounds have already been verified. The overflow-safety reasoning is now airtight and the comment makes it explicit.
- `RunView` offsets widened to `std::uint64_t` for parity with `TextRun`, so the #15 emitter casts rather than narrows. `PackageView::runBytes()` narrows at the span boundary where it is well-defined.
- The `Schema.cppm:19` /`TextBake.cppm:33`/`TextBake.cpp:142` S-number drift (`S5/#161` for the coverage draw path and the atlas/font package) corrected to `S6/#162` and `S4/#160` respectively.

The unreachable `UnsupportedSchemaVersion` is left as-is (consistent with the shader schema; the test now documents it produces `ReportRejected` via the header rather than itself).

Verification:
```
cmake --build build --target MduXCore mdux-textbake text_spec text_tools_spec
ctest --test-dir build -R '^(text_spec|text_tools_spec|shader_spec|ml_spec|draw_spec|evidence_tests|tools_tests|tools_spec|bridge_spec)'
100% tests passed, 0 tests failed out of 323
```

Trait-zone check still passes at configure time (2 governed target(s) clean of Vulkan/windowing dependencies). The pre-existing adapter-zone ICE on `VulkanRenderer.cpp` under GCC 16.0.1 is unrelated and tracked under issue #48; CI's `gcc:16` container is unaffected. Copilot's five inline comments (typo, the two S-number misreferences, the `S4/#160` recipe example, and the baker coverage ask) are all addressed by items 1, 2, and 4 above.

devjean-75:
## Follow-up review — 3e78e09

Re-reviewed against `origin/develop`. Same caveat as before: I read the diff, I did not build, so the 323/323 figure is yours.

All seven substantive items are genuinely addressed, and three of the four nits. The new baker suite is the real win — `text_tools_spec` mirrors `ml_tools_spec` exactly and pins each diagnostic to its stable code rather than to "it was rejected", which is the version that keeps working when the messages get reworded.

### Verified

| # | Item | Status |
|---|---|---|
| 1 | `recordSizeWire` | Deleted, with the stride pinned by `kSchemaVersion` instead — better than emitting it |
| 2 | Baker coverage | `text_tools_spec` + 9 scenarios; CMake matches `ml_tools_spec` |
| 3 | `PackageView`/`RunView` | 4 scenarios: find, runBytes happy path, both bounds guards |
| 4 | ADR-010 baker confusion | Text vs font baker now correctly separated |
| 5 | `docs/architecture.md` | All three tables updated |
| 6 | `TXT000` | Documented as recipe-unreadable, matching `SHB000` |
| 7 | Recipe-level checks | TXT009 / TXT010 with fix hints |
| nit | `byteEnd()` on unvalidated runs | Two-pass restructure is correct and the new comment is accurate |
| nit | `RunView` width | Widened to `uint64_t`; the `runBytes()` comparison genuinely widens rather than narrows |
| nit | S-number drift, ADR typo | Both fixed; the S4/S5 split now reads consistently across all six files |

---

### New findings

#### 1. ADR-010: the Decision list lost its sixth number (line 93)

Inserting the font baker as ¶2 renumbered 2–4 into 3–5, but the last item became an unordered `-` bullet at 2-space indent instead of `6.`. The section now renders as an ordered list 1–5 followed by an orphan bullet.

#### 2. ADR-010: stale cross-reference (line 197)

> *"The build-fail enforcement described in Decision paragraph 4 is delivered by the restricted-charset validation in #161 (S5)"*

That enforcement is now paragraph **5**. Paragraph 4 is the runtime `DrawList`/`CoverageR8` path, which is not a build-fail rule at all.

#### 3. The TXT008 test doesn't exercise the path it says it does (`tests/text/TextBakeTests.cpp:412-416`)

The comment contradicts itself — *"so the first-differing-byte path is exercised: the file is shorter than the produced bytes, so the comparison fails by length at byte 1"* — and the second half is what actually happens. Writing `{` as the whole committed file gives `common = min(produced.size(), 1) = 1`; byte 0 is `{` in both, so the comparison loop finds nothing and falls through to the **length-differs** branch.

Since the assertion only checks `code == "TXT008"` and both branches emit TXT008, it passes either way. The offset-reporting loop — the index arithmetic I flagged as the part that goes wrong quietly — is still untested. Writing a same-length copy with one byte flipped mid-file would cover it, and would also pin the reported offset, which is the diagnostic's whole value.

#### 4. The offset-past-end view assertion can't fail (`tests/text/SchemaTests.cpp:769`)

`RunView{.byteOffset = 8, .byteLength = 0}` against a 4-byte sidecar: the expected result is an empty span, but a *guardless* implementation returns an empty span here too, because the length is zero. The scenario can't distinguish guard-present from guard-absent.

`.byteOffset = 8, .byteLength = 6` fixes it — then an empty span is only reachable through the guard. (The companion length-past-end case at offset 0 / length 6 is already sound.)

#### 5. An empty `sidecar` still falls through to the generic TXT005

`sidecar = ""` in a recipe passes the new separator check (`find` returns `npos` on an empty string), then surfaces as *"assembled package is not valid: sidecar path is empty"* at line 0 with no fix hint — exactly the shape TXT009/TXT010 were added to remove. One more line beside the separator check, or fold it in as `recipe.sidecar.empty() || …has a separator`.

#### 6. The diagnostic-code registry is out of numeric order (`tools/text/TextBake.cpp:33-42`)

TXT009 and TXT010 are declared between TXT004 and TXT005. Assigning new numbers rather than renumbering was the right call — the codes are published API — but for a block documented as a stable registry, listing them after TXT008 costs nothing and reads much better.

### Nits

- **Formatting slip**: `state->outputs = produceValid();` is flush-left at `tests/text/TextBakeTests.cpp:404`, inside an otherwise-indented `Given` block. clang-format will take it.
- **`architecture.md:205`** still opens the Planned section with *"Nothing below exists in the tree today"*, while the #14 row directly beneath now says S1 landed. Either soften the preamble or scope the row's note.
- **The record layout is now undocumented.** Deleting `recordSizeWire` also removed the only description of what the 6 bytes are (*"glyphIndex … x and y (signed pixel positions)"*). #162/S6 has to interpret them, and `recordSize = 6` alone won't tell that wave what the fields were meant to be. Worth keeping a line of it on `recordSize` at `Schema.cppm:96`.
- **Trailing newlines**: all eight new `tests/text/`, `src/text/`, `include/mdux/text/` and `tools/text/` files end without one. On develop, 108 of 112 C++ files have a trailing newline — the four that don't are pre-existing (`mdux.cppm`, `src/mdux.cpp`, `tests/tools/CliTests.cpp`, `ToolsSpecMain.cpp`). Not enforced by `.clang-format`, so it's convention rather than rule, and this was my miss in the first pass — flagging it now rather than later.

---

Nothing here blocks merge. #1–#2 are a two-line doc fix; #3–#4 are worth taking while the tests are fresh, since both are scenarios that currently pass without proving what they claim.

ambroise-leclerc:
Addressed all six findings and four nits in `cc94b8b`. 323/323 still green.

| Finding | Fix |
|---|---|
| 1. ADR-010 Decision list lost its sixth number | Orphan bullet renumbered to `6.`; the section now renders as 1–6 |
| 2. ADR-010 stale cross-reference | *Decision paragraph 4* → *paragraph 5* at line 197 (¶4 is the runtime path, not a build-fail rule) |
| 3. TXT008 test doesn't exercise the offset-reporting loop | Rewritten: flip one mid-file byte in a same-length committed `package.json`, not truncate to 1 byte. The first-differing-byte branch actually fires; the assertion pins the reported offset against the flipped index |
| 4. Offset-past-end view assertion can't fail | `RunView.byteLength` widened from 0 to 6 — an empty span is now only reachable through the guard, not through a zero-length `subspan(offset, 0)` |
| 5. Empty `sidecar` falls through to generic TXT005 | Recipe-level check added (TXT010) with a fix hint, and a matching case added to the `parseRecipe` rejection corpus |
| 6. Diagnostic-code registry out of numeric order | TXT009/TXT010 declarations moved after TXT008 — codes unchanged, only the declaration layout reads better |

Nits:
- Record layout restored on `recordSize` at `Schema.cppm:96`: the 6 bytes are `uint16_t glyphIndex`, `int16_t x`, `int16_t y`. Deleting `recordSizeWire` had removed the only description of what the fields are; #162/S6 needs it.
- `architecture.md` Planned-not-built preamble softened from "Nothing below exists in the tree today" to note that the #14 row's slice has landed — consistent with the row's updated wording.
- Trailing newlines added to all nine new C++ files.

```
cmake --build build --target text_spec text_tools_spec mdux-textbake
ctest --test-dir build -R '^(text_spec|text_tools_spec|shader_spec|ml_spec|draw_spec|evidence_tests|tools_tests|tools_spec|bridge_spec)'
100% tests passed, 0 tests failed out of 323
```

The two correctness items (#3, #4) were the ones worth taking while the tests are fresh: both now prove what they claim.

Issue #161, opened 2026-08-06
S5 — mdux.font.schema and restricted-charset validation


ambroise-leclerc:
Part of #14. Depends on the S4 atlas baker child (#160).

Introduce the governed `mdux.font.schema` module — the canonical font-package shape: atlas metrics, glyph advances, kerning pairs the baker chose to bake, and the restricted-charset table for dynamic text. The compiler rejects any dynamic-text format that could escape the table; this is what turns "no shaping on device" from a slogan into a compile-time error. Tabular-figure assertion: digit glyphs share an advance so numeric displays do not jitter. The baker validates the declared charset against the committed atlas and fails closed on a missing glyph.

Acceptance: a dynamic-text format referencing an unbaked code point fails the baker with a stable code; tabular-figure mismatches fail the baker; `font_spec` schema tests pass on MSVC and GCC; the governed-zone check still passes.

ambroise-leclerc:
Landed on `develop` via #177.

`mdux.font.schema` is now the canonical font-package shape in MduXCore, imported by the baker that writes a package and by whatever reads one — replacing the informal JSON `mdux-textbake` emitted by hand in S4. `write()` validates on the way out and `parse()` on the way in, so no caller holds an unchecked `FontPackage`.

Acceptance:

| | |
|---|---|
| tabular-figure mismatches fail the baker | yes — `TXT019`, over `SchemaError::TabularFigureMismatch`. Verified against the committed font first: DejaVu Sans' ten digits all advance 1303 |
| `font_spec` schema tests pass on MSVC and GCC | yes — 5 scenarios, both legs |
| the governed-zone check still passes | yes — "2 governed target(s) clean of Vulkan/windowing dependencies" |
| a dynamic-text format referencing an unbaked code point fails the baker with a stable code | **partly** — the `restrictedCharset` table and `permits()` are here, and `validate()` refuses a package whose charset names a code point it has no glyph for. The compiler-side rejection of a *format* belongs to `.medui` (#15), which this unblocks |

That last row is the one caveat: this child supplies what #15 will enforce against, not the enforcement itself.

S6 (#162) is the last child of #14.

Issue #14, opened 2026-07-26
Epic: Font and text pipeline


ambroise-leclerc:
Deterministic text prepared offline. The completed evidence kernel (#12) supplies the artifact machinery; the draw path (#13) is closed and `DrawMode::CoverageR8` is already wired into `mdux.draw`. This epic opens Wave 4.

## Scope

- Static text is shaped/rasterized offline per approved locale.
- Dynamic text uses a compiler-validated restricted charset with tabular digits.
- v1 supports TrueType `glyf` outlines and Latin/Cyrillic/Greek LTR only; unsupported scripts fail compilation.
- No hinting, GPOS, ligatures, HarfBuzz, CFF/CFF2, or runtime shaping.

See ADR-010 for the architectural commitment that the runtime performs no parsing, no layout and no shaping; the enforced compile-fail for unsupported scripts lands in #161 (S5).

## Children

- [ ] #157 — S1 — mdux.text.schema and mdux-textbake skeleton
- [ ] #158 — S2 — Hand-parsed TrueType (`glyf` only)
- [ ] #159 — S3 — Deterministic rasteriser with coverage AA
- [ ] #160 — S4 — Atlas packer and font baker
- [ ] #161 — S5 — mdux.font.schema and restricted-charset validation
- [ ] #162 — S6 — Coverage draw path and text pixel tests

**Blocked by:** none; #13 is closed. Children are dependency-ordered: S2 (#158) after S1 (#157); S3 (#159) after S2; S4 (#160) after S3; S5 (#161) after S4; S6 (#162) after S3 and S4.
**Blocks:** #15.

## Done when

A baked font package produces a committed `generated/font/<id>/` artifact byte-identical on MSVC and GCC; a text pixel test renders static glyphs on lavapipe and matches expected bounds and coverage; a dynamic-text format referencing an unbaked code point fails the baker; and `mdux_verify_trust_zones()` still passes with the new `mdux.text.schema` and `mdux.font.schema` modules registered.

ambroise-leclerc:
Closing. All six children are merged, and the pipeline runs end to end: a `.ttf` in `recipes/` becomes a committed package, and a run baked from that package reaches a frame.

| | | landed as |
|---|---|---|
| S1 | Text schema and baker skeleton | `mdux.text.schema`, `mdux-textbake` (#157) |
| S2 | Hand-parsed TrueType, `glyf` only | `mdux.tools.truetype`, host-only (#158) |
[...]

PR #165, opened 2026-08-07
S2 — Hand-parsed TrueType (glyf only) (#158)

ambroise-leclerc:
Part of #14. Depends on #157 (closed).

## What this lands

A host-only TrueType parser that walks `glyf` only, deliberately not linking `freetype` or any other font SOUP — the parallel of `tools/shader/Spirv.cpp` for SPIR-V and `tools/ml/Safetensors.cpp` for safetensors (ADR-008 decision 6, mirrored onto text by ADR-010).

- `tools/text/Truetype.cppm` + `.cpp` — module `mdux.tools.truetype`. `parse()` walks the offset table, `head`/`maxp`/`loca` and the implicit outline check; `parseGlyph()` walks one glyf record against the resulting `Font`. Both return `Result<T, ParseError>`, paralleling `Spirv.cpp` and `Safetensors.cpp`.
- `tests/text/TruetypeTests.cpp` — six SpecLab scenarios (minimal valid font, `true`/`typ1` container variants, a square glyph with 4 on-curve points, an empty glyph) plus two table-driven rejection corpora pinning the stable `ParseError` for every directory-walk and per-glyph rejection path.
- Wired into `MduXTextBakeLib` (`tools/CMakeLists.txt`) as a public module so S4's font baker (#160) can `import mdux.tools.truetype` through the same target that hosts the recipe model. Joined to `text_tools_spec` in `tests/CMakeLists.txt`.
- `docs/architecture.md` updated: `MduXTextBakeLib` row notes the new module; the #14 row in "Planned, not built" records S2 as landed.

## v1 scope and rejections

v1 supports TrueType `glyf` outlines and Latin/Cyrillic/Greek LTR only. Per ADR-010 decision paragraph 5, the structural half of the enforcement is this parser, and it rejects with stable codes (never reaching the runtime):

- Composite glyphs (`numberOfContours == -1`): `CompositeGlyphRejected`
- CFF / CFF2 outlines (`glyf` absent): `CffOutlinesRejected`
- Hinting instruction bytecode per glyph: `HintingRejected`
- `GPOS` table present: `GposRejected`
- `GSUB` table present (carries ligatures and contextual substitutions): `GsubRejected`
- Apple AAT layout tables `morx` / `mort` present: `AppleLayoutRejected`

RTL and complex scripts cannot be detected from the font binary alone — they are a property of the string-to-glyph mapping, which is the `.medui` compiler's (#15) and the restricted-charset's (#161) job, not this parser's. The mechanical enforcement delivered by #161 (S5) is the second half; this module is the structural one.

## Acceptance

- Malformed/truncated/unsupported font fixtures fail with stable codes — covered by the two table-driven rejection corpora.
- The baker links no third-party font library — `MduXTextBakeLib`'s `target_link_libraries` is unchanged (still `PUBLIC MduX::ToolsCommon`, `PRIVATE MduX_warnings`); the parser depends only on `mdux.core.result` and `std`.
- `mdux_verify_trust_zones()` still passes — configure reports "2 governed target(s) clean of Vulkan/windowing dependencies".
- Parser unit tests cover `glyf` simple and composite outlines, and every rejection path — `TruetypeTests.cpp` registers six scenarios in `text_tools_spec`; all 15 cases in the binary pass.

## Local verification

Verified locally with `g++ 16.1.0` (the default `/usr/bin/g++` on this machine is a 16.0.1 experimental trunk snapshot that ICEs on `VulkanRenderer.cpp` — see issue #48, pre-existing and unrelated to this change). Configured and built the full tree (`MDUX_BUILD_EXAMPLES=ON MDUX_BUILD_TESTS=ON`):

- `mdux_verify_trust_zones()` clean, no new third-party link dependency.
- `ctest -E "InstallTreeConsumer|offscreen_tests|ml.noheap.symbolScan"`: **393 / 393 pass**.
- Exclusions: `InstallTreeConsumer` (pre-existing flakiness on this container), `offscreen_tests` (needs a Vulkan ICD that is not present here), `ml.noheap.symbolScan` (needs `nm`/`dumpbin` — separately affected by the container's tooling). None related to this change.

```sh
CC=gcc-16.1 CXX=g++-16.1 cmake -B build -S . -G Ninja -DCMAKE_BUILD_TYPE=Release -DMDUX_BUILD_EXAMPLES=ON -DMDUX_BUILD_TESTS=ON
cmake --build build
ctest --test-dir build -E "InstallTreeConsumer|offscreen_tests|ml.noheap.symbolScan"
```

## Out of scope (deliberately)

- No `mdux_bake_artifact()` call site is registered in this PR. S4 (#160) is the wave that feeds a `.ttf` recipe through this parser and commits a font artifact.
- The baker-facing conversion from `ParseError` to a `TXT`-prefixed `cli::Diagnostic` is deferred to S4 (#160) — the same wave that introduces the recipe fields the baker needs to call this parser at all. The `ParseError` codes this PR publishes are the stable contract that conversion will switch on.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -86,7 +86,7 @@ performs no checking and confers no compliance.
 | `MduXToolsCommon` | `tools/common/` | TOML subset reader, CLI parser, shared diagnostic envelope |
 | `MduXShaderBakeLib` | `tools/shader/` | `mdux-shaderbake`, `mdux-shaderemit` |
 | `MduXMlBakeLib` | `tools/ml/` | `mdux-mlbake` |
-| `MduXTextBakeLib` | `tools/text/` | `mdux-textbake` |
+| `MduXTextBakeLib` | `tools/text/` | `mdux-textbake`; also hosts `mdux.tools.truetype` (the host-only glyf parser, #158) |
 
 Host tools parse untrusted input, so they are deliberately outside the governed zone. They are
 never linked into `MduXCore` or `MduX` and are absent from the install/export set.
@@ -209,7 +209,7 @@ tracking issue; the issue is authoritative for what remains.
 
 | Planned | Issue | Note |
 |---|---|---|
-| Font and text pipeline | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) | S1 (`mdux.text.schema` + `mdux-textbake` skeleton) landed; S2–S6 open |
+| Font and text pipeline | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) | S1 (`mdux.text.schema` + `mdux-textbake` skeleton) landed. S2 (`mdux.tools.truetype` glyf parser, #158) landed; S3–S6 open |
 | `.medui` compiler | [#15](https://github.com/ambroise-leclerc/MduX/issues/15) | the replacement for the retired HTML/CSS path |
 | Rendered-truth verification | [#16](https://github.com/ambroise-leclerc/MduX/issues/16) | beyond the current pixel test |
 | Content components | [#17](https://github.com/ambroise-leclerc/MduX/issues/17) | `SignalTrace`, `StatusIndicator`, `NumericDisplay` and the rest |
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -586,6 +586,7 @@ mdux_discover_tests(text_spec)
 add_executable(text_tools_spec
     text/TextToolsSpecMain.cpp
     text/TextBakeTests.cpp
+    text/TruetypeTests.cpp
 )
 
 target_link_libraries(text_tools_spec PRIVATE MduX::TextBakeLib speclab::speclab)
--- a/tests/text/TruetypeTests.cpp
+++ b/tests/text/TruetypeTests.cpp
@@ -1,0 +1,1063 @@
+/**
+ * @file TruetypeTests.cpp
+ * @brief BDD scenarios for the host-only TrueType (glyf) parser (issue #158).
+ *
+ * @compliance ADR-004 Trust zones in C++ (host-tools zone)
+ * @compliance ADR-007 Evidence pipeline doctrine
+ * @compliance ADR-010 No on-device text shaping
+ *
+ * The corpus is built in memory rather than committed as `.ttf` files. A malformed font is a
+ * handful of bytes described by one line of code here; as a committed blob it would be an
+ * opaque artifact whose intended defect a reviewer has to take on trust (the same argument
+ * `SpirvTests.cpp` makes for `SpirvFixtures.hpp` and `SafetensorsTests.cpp` makes for its
+ * fixtures).
+ *
+ * Every rejection scenario asserts the specific `ParseError` code, not merely that parsing
+ * failed. A parser that rejects everything with one generic error passes "it was rejected"
+ * and is useless to the author who has to fix the font - and useless to S4 (#160), which has
+ * to convert each code into a distinct `TXT` diagnostic. The first scenario belows holds the
+ * success case; the two table-driven scenarios hold all rejection paths for `parse()` and
+ * `parseGlyph()`, so a new rejection added later lands next to its enumerator by index rather
+ * than spread across the file.
+ */
+
+import std;
+import speclab;
+import mdux.core.result;
+import mdux.tools.truetype;
+
+#include "../framework/SpecLabBridge.hpp"
+
+namespace {
+
+namespace tt = mdux::tools::truetype;
+using tt::ParseError;
+
+constexpr std::uint32_t kSfntTrue  = 0x00010000u;
+constexpr std::uint32_t kHeadMagic = 0x5F0F3CF5u;
+
+// ---------------------------------------------------------------------------
+// An assembler for in-memory TrueType files, plus helpers for the glyf records each scenario
+// uses. A real `.ttf` would be the same bytes in a different file; building them in code keeps
+// the tests free of binary fixtures whose defect a reviewer would have to inspect with a hex
+// dump. The assembler emits directory records in a fixed order (head, maxp, loca, glyf,
+// extras) so the per-table offsets are deterministic and a `poke()` on a known location is
+// safe across the whole corpus.
+// ---------------------------------------------------------------------------
+
+class Builder {
+public:
+    Builder& sfnt(std::uint32_t v) {
+        sfnt_ = v;
+        return *this;
+    }
+    Builder& unitsPerEm(std::uint16_t v) {
+        upe_ = v;
+        return *this;
+    }
+    Builder& indexToLocFormat(std::int16_t v) {
+        idxFmt_ = v;
+        return *this;
+    }
+    Builder& numGlyphs(std::uint16_t n) {
+        ng_ = n;
+        return *this;
+    }
+    Builder& headMagic(std::uint32_t v) {
+        headMagic_ = v;
+        return *this;
+    }
+
+    /// Replaces the glyf record at index `i`. `numGlyphs` is bumped to `i+1` if lower.
+    Builder& rawGlyph(std::size_t i, std::vector<std::byte> bytes) {
+        if (i + 1 > rawGlyphs_.size()) {
+            rawGlyphs_.resize(i + 1);
+        }
+        rawGlyphs_[i] = std::move(bytes);
+        if (ng_ < i + 1) {
+            ng_ = static_cast<std::uint16_t>(i + 1);
+        }
+        return *this;
+    }
+
+    Builder& extraTable(std::string tag, std::vector<std::byte> bytes) {
+        extras_.emplace_back(std::move(tag), std::move(bytes));
+        return *this;
+    }
+
+    Builder& dropHead() {
+        dropHead_ = true;
+        return *this;
+    }
+    Builder& dropMaxp() {
+        dropMaxp_ = true;
+        return *this;
+    }
+    Builder& dropLoca() {
+        dropLoca_ = true;
+        return *this;
+    }
+    Builder& dropGlyf() {
+        dropGlyf_ = true;
+        return *this;
+    }
+
+    /// An override for the byte length the `loca` directory record declares - the directory
+    /// length is what `readLoca()` reads when sizing the table, so a smaller value here is the
+    /// `LocaSizeMismatch` fixture. Zero means "use the derived length".
+    Builder& locaDeclaredLength(std::uint32_t v) {
+        locaLen_ = v;
+        return *this;
+    }
+
+    /// An override for the byte length the `head` directory record declares - a smaller value
+    /// fires `TruncatedHead`, since the `head` subspan shrinks below 54 bytes.
+    Builder& headDeclaredLength(std::uint32_t v) {
+        headLen_ = v;
+        return *this;
+    }
+
+    /// An override for the byte length the `maxp` directory record declares.
+    Builder& maxpDeclaredLength(std::uint32_t v) {
+        maxpLen_ = v;
+        return *this;
+    }
+
+    /// An override for the head table's directory `offset` field; setting it past `bytes().size()`
+    /// is the `TableOutOfBounds` fixture.
+    Builder& headDeclaredOffset(std::uint32_t v) {
+        headOff_ = v;
+        return *this;
+    }
+
+    /// Sets `numTables` declared in the offset table to a different value than the directory's
+    /// physical length - the parser reads only as many records as `numTables` claims, so a
+    /// larger value is the `TruncatedTableDirectory` fixture.
+    Builder& declaredNumTables(std::uint16_t n) {
+        declaredNumTables_ = n;
+        return *this;
+    }
+
+    /// An override that replaces the loca slot at index `i`'s stored short entry with `v`. The
+    /// builder re-derives loca contents; this is the simple per-entry perturbation the
+    /// `LocaOutOfBounds` fixture uses.
+    Builder& locaShortEntry(std::size_t i, std::uint16_t v) {
+        locaOverrides_[i] = v;
+        return *this;
+    }
+
+    /// Serialized bytes plus deterministic per-table offsets, so a test can `poke()` into the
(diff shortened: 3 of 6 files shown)


devjean-75 (commented):
## Review

The bounds discipline here is genuinely good, and better than most hand-rolled font parsers: every read funnels through `beU16`/`beI16`/`beU32`, table extents are computed in 64-bit before comparison (`Truetype.cpp:208`, `:287`) so a malformed `numTables` or `offset+length` can't wrap, and `walkDirectory` validates each record's slice before `parse()` ever calls `subspan`. I traced every `*optional` dereference and found no unchecked one that isn't already dominated by a size guard. The error taxonomy is well-factored and the "no FreeType" argument is consistent with the `Spirv.cpp` / `Safetensors.cpp` precedent.

The problems are not memory safety. They're that the parser rejects most of the real fonts it exists to read.

---

### 1. Zero-length glyf records — the spec representation of an empty glyph — are rejected

`Truetype.cpp:386`:

```cpp
const auto span = font.glyf.subspan(start, end - start);
if (span.size() < 10) {
    return err(ParseError::TruncatedGlyph);
}
```

In TrueType, a glyph with no outline (space, and every other blank) is encoded by `loca[i] == loca[i+1]` — a **zero-length** glyf record. That's the mandated encoding, not a corruption. This parser returns `TruncatedGlyph` for all of them.

Verified against `/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf`:

```
numGlyphs = 6253, indexToLocFormat = 1
zero-length glyf records: 63
first ten gids: [1, 2, 3, 98, 768, 2787, ...]
loca[3] == loca[4]  → True     (gid 3 = space)
```

So `parseGlyph()` fails on the space character of a stock font. The test suite doesn't catch this because `emptyGlyph()` (`TruetypeTests.cpp:361-362`) builds a **12-byte** record with `numberOfContours = 0` — a legal but non-canonical encoding — and `text-truetype-empty-glyph-parses` asserts that form. There is no test where `loca[i] == loca[i+1]`. The fix is to return an empty `SimpleGlyph` when `end == start`, before the 10-byte header check.

### 2. The rejection policy is stricter than ADR-010 asks, and no stock font survives it

ADR-010 decision 5 says:

> Unsupported scripts …, **composite glyph substitutions the baker did not pre-bake**, CFF/CFF2 outlines, GPOS positioning, ligatures and hinting **fail the font baker (#160/#161)**…

That is a statement about what the *runtime* must never do, and it explicitly contemplates the baker **pre-baking** composites. The parser instead rejects on the *presence of a table or field in the input file*, which is a much stronger rule. Concretely:

| Input | ADR-010 concern | What this parser does |
|---|---|---|
| `GPOS` / `GSUB` table present | runtime must not apply positioning/substitution | `parse()` fails outright — even though the baker only reads `glyf`/`loca`/`head`/`maxp` and discards these |
| `numberOfContours == -1` | composites the baker *did not pre-bake* | every composite rejected; none are pre-baked |
| `instructionLength > 0` | no on-device hinting | glyph rejected, though the baker need only ignore the `instructions[]` array |

Measured on the eight fonts available on my machine:

```
GPOS,GSUB   DejaVuSans.ttf          GPOS,GSUB   DejaVuSerif.ttf
GPOS,GSUB   DejaVuSans-Bold.ttf     GPOS,GSUB   DejaVuSerif-Bold.ttf
GPOS,GSUB   DejaVuSansMono.ttf      … (8 of 8)
accepted: 0 of 8
```

And per-glyph, *assuming GSUB/GPOS were stripped first*, on DejaVuSans:

```
TruncatedGlyph      (zero-length record) :    63    1.0%
CompositeGlyphRejected                   :  2607   41.7%
HintingRejected                          :  1007   16.1%
parses successfully                      :  2576   41.2%
```

41.7% composites is not an edge case — it's how `é`, `à`, `ü`, and the accented Cyrillic and Greek forms are stored. The PR states "v1 supports … Latin/Cyrillic/Greek LTR", but under this parser you cannot bake `é` from a stock font. S4 (#160) is the wave that feeds a real `.ttf` through this, so it will hit all four rows above on day one.

None of this is a coding error — it's a scope decision. But the PR presents the rejection set as ADR-010's structural half, and ADR-010's wording ("substitutions the baker did not pre-bake") reads as permitting host-side composite resolution. Worth settling explicitly now: either pre-bake composites and ignore hinting/GSUB/GPOS at parse time, or state in the module docs that inputs must be pre-subsetted and stripped, and say which tool does that.

### 3. `parseGlyph()` trusts `Font` invariants that the type doesn't enforce

`Truetype.cpp:380-381`:

```cpp
const std::uint32_t start = font.loca[glyphIndex];
const std::uint32_t end   = font.loca[glyphIndex + 1];
```

Safe for a `Font` produced by `parse()`, which always builds `numGlyphs + 1` entries. But `Font` is an exported aggregate with public fields (`Truetype.cppm:142-154`), so a caller can construct or mutate one with `numGlyphs = 5` and an empty `loca` — and `operator[]` on `std::vector` is UB, not a bounds check. Given the module's stated contract ("it never reads outside the buffer it was given", `Truetype.cppm:25-26`) and that S4 will hold `Font` across calls, a one-line guard is cheap:

```cpp
if (font.loca.size() < static_cast<std::size_t>(font.numGlyphs) + 1u) {
    return err(ParseError::LocaSizeMismatch);
}
```

### 4. `satAdd` guesses where the module promises to refuse

`Truetype.cppm:26-27` states the contract as *"anything it does not understand is a refusal with a specific code rather than a guess"*. `satAdd` (`Truetype.cpp:362-371`) is a guess: a coordinate accumulation that overflows `int16` is clamped to `INT16_MAX`/`INT16_MIN` and parsing continues. Well-formed fonts can't reach it (coordinates are `int16` bounded by the `head` bbox), so it fires **only** on malformed input — precisely the case where the doctrine says to return a code. The output is a plausible-looking but wrong outline that S4 would then bake and byte-commit as evidence. Suggest a `CoordinateOverflow` code instead; the comment's own reasoning ("would already be malformed") argues for rejection, then does the opposite.

Same shape at `Truetype.cpp:461`: a `REPEAT_FLAG` count that overruns `numPoints` is silently truncated by the `flags.size() < numPoints` guard. The comment above claims "an abbreviated repeat or an absent byte fails with `TruncatedGlyphFlags`" — an *over-long* repeat doesn't fail at all.

### Smaller correctness and accuracy items

- **`Truetype.cpp:34-36` comment is wrong.** It says `'OTTO'` "is rejected here only if it carries a CFF table, which the directory walk below spots." OTTO isn't in the accepted sfnt list at `:276`, so it's rejected as `NotATrueTypeOutlineFont` and never reaches the walk. `CffOutlinesRejected` is reachable only for a `0x00010000`/`'true'`/`'typ1'` font that also carries `CFF `/`CFF2`.
- **`flagReservedMask = 0xC0` (`Truetype.cpp:63`) over-rejects.** In the current OpenType spec, bit 6 is `OVERLAP_SIMPLE` (a defined flag with no shaping semantics), not reserved; only bit 7 is. Fonts from modern tooling set it. Mask should be `0x80`. I could not reproduce this on the DejaVu set (0 glyphs use bit 6), so I'm flagging it as a spec point rather than a confirmed failure.
- **`LocaSizeMismatch` doesn't check what its name says.** `Truetype.cpp:233` uses `loca.size() < entryCount * elementSize`, so an over-long `loca` passes, while the enum doc says "byte length does not match numGlyphs+1 entries". Either use `!=` or reword.
- **`unitsPerEm` range `[16, 4096]`** (`Truetype.cpp:321`) is narrower than the spec's `[16, 16384]`. DejaVu's 2048 is fine, but an 8192-upem font is spec-valid and rejected with no ADR backing the narrowing.
- **Dead checks.** `Truetype.cpp:398` (`!numberOfContours`) and `:428` (`!instructionLength`) can't fire — `span.size() >= 10` and the `endPtsEnd + 2` guard already guarantee presence. Harmless, but they read as live bounds checks.
- **Garbled comment** at `Truetype.cpp:229-230`: "a font that author declares one glyph".

### Test coverage

The two table-driven rejection corpora are a good structure and every `ParseError` does get pinned. Gaps:

- No zero-length glyf record case (finding 1) — the one encoding that breaks on real fonts.
- No case for over-long `REPEAT_FLAG`, `satAdd` clamping, or over-long `loca`. All three are paths where the parser currently *accepts* malformed input; the corpora only cover paths where it rejects.
- Everything is synthetic in-memory. That's consistent with the commit message's reasoning and I agree with it for the malformed corpus — but there's no test that any *real* font parses, which is why finding 1 slipped through. One committed, license-clean, subsetted `.ttf` (or a checked-in byte array derived from one) would have caught it.
- Minor convention drift: `SpirvTests.cpp` factors its builders into `SpirvFixtures.hpp`; here ~320 lines of `Builder` live inline in `TruetypeTests.cpp`. A `TruetypeFixtures.hpp` would match the precedent the PR otherwise follows closely.
- No fuzz target — but neither has `Spirv.cpp` or `Safetensors.cpp`, so that's a repo-wide observation, not a regression in this PR. Given all three are untrusted-input boundaries, a shared fuzz harness is worth an issue.

### Non-issues I checked

`import mdux.core.result` in a host tool is fine under ADR-004; `MduXTextBakeLib`'s link line is genuinely unchanged, so the no-SOUP acceptance claim holds. The `@compliance ADR-010` citations resolve — ADR-010 exists on this branch and its base (it arrives via #157), just not yet on `develop`. `Result`/`err` usage matches `Result.cppm`. The `uint8_t` loop counter at `:461` terminates correctly at `repeatCount == 255`.

---

### Summary

Memory safety is solid and the code is unusually well-documented. Two things worth resolving before merge: **zero-length glyf records** (a straightforward bug that breaks the space character of every real font), and the **reject-on-presence policy**, which as written admits no stock font and only 41% of glyphs even after stripping — that one is a scope call, but S4 depends on the answer. The `Font` bounds guard and the `satAdd`-vs-doctrine point are smaller but cheap to fix now while the `ParseError` set is still unpublished.

ambroise-leclerc:
Thanks — the two blocking findings were both real, and the second one changed the design.

**1. Zero-length glyf records (fixed, 16584c1)**

Confirmed exactly as reported. `parseGlyph()` now returns an empty `SimpleGlyph` when `loca[i] == loca[i+1]`, checked before the 10-byte header bound. You were also right about why the corpus missed it: `emptyGlyph()` builds the legal-but-non-canonical 12-byte `numberOfContours == 0` form, and nothing exercised the zero-length one. `zeroLengthGlyphParses` now pins it and asserts the fixture really produced `loca[0] == loca[1]`.

**2. Reject-on-presence (changed to skip-on-presence)**

Settled in your favour. ADR-010 decision 4 forbids *on-device* code that walks a font table or advances a pen by a runtime-computed width; it does not require the host's input file to be free of those bytes. Rejecting on presence was enforcing a stronger rule than the ADR asks for, at the cost of accepting no shipping font.

So `GPOS`, `GSUB`, `GDEF`, `morx` and `mort` now fall through the tag ladder unread — the enforcement is that the walk never slices them, so nothing downstream can bake them — and a glyph's `instructions[]` array is stepped over with a bounds check rather than refused. `GposRejected`, `GsubRejected`, `AppleLayoutRejected` and `HintingRejected` are gone from the enum. `CffOutlinesRejected` stays (there is no `glyf` to read at all) and so does `CompositeGlyphRejected`, which is the signal S4 (#160) acts on when it pre-bakes them — your reading of "substitutions the baker did not pre-bake" as the seam is the one I took.

Measured after the change, running the real parser over the same eight faces you used:

```
accepted: 8 of 8          (was 0 of 8)

DejaVuSans.ttf            upem=2048  numGlyphs=6253  zero-length records=63
    3646 ( 58.3%)  parsed                     (was 41.2%)
    2607 ( 41.7%)  CompositeGlyphRejected
       0          TruncatedGlyph  (was 63)
       0          HintingRejected (was 1007)
    gid 3 (space): parsed as blank
DejaVuMathTeXGyre.ttf     4282 (100.0%) parsed
```

Composites are now the only per-glyph rejection on real input, which is the shape you argued for.

**3. `Font` bounds guard** — added. `parseGlyph()` re-checks `loca.size() >= numGlyphs + 1` and returns `LocaSizeMismatch`; the corpus has a case that mutates a parsed `Font`'s `numGlyphs` to 5 and confirms it.

**4. `satAdd` guesses** — agreed, and the comment's own reasoning was the argument against itself. It is now `checkedAdd` returning `optional`, and the caller emits a new `CoordinateOverflow`. The over-long `REPEAT_FLAG` has its own code too (`GlyphFlagRepeatOverrun`) rather than being silently truncated by the `flags.size() < numPoints` guard.

**Smaller items** — all taken:

- `OTTO`: rather than fix the comment to match the code, I made the code match the comment. `OTTO` now passes the version-word gate so its `CFF ` table yields `CffOutlinesRejected`, which names the problem, instead of `NotATrueTypeOutlineFont`, which only says the container word was unfamiliar. Pinned in the corpus.
- `flagReservedMask` `0xC0` → `0x80`. You flagged this as a spec point you could not reproduce; it is a real over-rejection, and `overlapSimpleFlagParses` now covers bit 6.
- `LocaSizeMismatch`: kept `<` and reworded instead. sfnt pads tables to 4 bytes, so a short-format loca with an even entry count legitimately carries two trailing bytes — `!=` would reject those. `overLongLocaIsAccepted` pins it.
- `unitsPerEm` widened to the spec's `[16, 16384]`.
- Dead checks at `:398`/`:428` removed, with a comment on why the five header reads are safe to dereference.
- Garbled comment at `:229`, the "single zero-point contour state" comment, "filesystem form", "a glyphs", "belows", "tow" — all fixed.

**Test coverage** — six new scenarios (layout tables skipped, zero-length record, hinted glyph parsing to its unhinted twin's outline, OVERLAP_SIMPLE, over-long loca) and four new rejection rows. 398/398 ctest pass.

On the real-font fixture: I verified against the eight stock DejaVu faces with a throwaway harness (the numbers above) but did not commit it — a binary fixture is the thing this suite's header argues against, and the two shapes that actually broke (zero-length records, layout-table presence) now have synthetic scenarios that fail without the fix. `TruetypeFixtures.hpp` and the shared fuzz harness are both fair points; I have left them for a follow-up rather than growing this PR further.

PR #171, opened 2026-08-08
Wave 4 S2 — land the hand-parsed TrueType glyf parser on develop (#158)

ambroise-leclerc:
Merges the Wave 4 epic branch into `develop`. Carries the three commits of S2 (#158) — S1 (#157) already landed separately via #163.

## Why this PR exists as its own step

PR #165 merged S2 into `14-text-schema`, not into `develop`, and it ran **no CI at all**: at the time its base branch matched none of the workflow trigger filters, so the symptom was not a failing check but no check. #164 has since closed that gap (base filters now include `[0-9]+-*`, and `develop` was always covered), so **this is the first CI build the TrueType parser has ever had** — confirmed by checking every run `14-text-schema` has recorded: `cc94b8b`, `3e78e09` and `73823f4` all predate `tools/text/Truetype.cpp`.

`develop` moved 20 commits ahead while this branch sat — all of it CI and Dependabot work, none touching `tools/` or `tests/text/`. `origin/develop` is merged in here (`636436d`, no conflicts) so the PR is up to date and what CI checks is the state that will land.

## What lands

| File | |
|---|---|
| `tools/text/Truetype.cppm` / `.cpp` | module `mdux.tools.truetype` — `parse()` walks the offset table, `head`/`maxp`/`loca`; `parseGlyph()` walks one glyf record. Both return `Result<T, ParseError>`, paralleling `Spirv.cpp` and `Safetensors.cpp`. |
| `tests/text/TruetypeTests.cpp` | 11 SpecLab scenarios plus two table-driven rejection corpora pinning the stable `ParseError` for every rejection path. |
| `tools/CMakeLists.txt` | registers the module as a public module of `MduXTextBakeLib`, so S4's baker (#160) can `import mdux.tools.truetype`. |
| `docs/architecture.md` | records S2 as landed. |

## Scope decision made during review

The parser originally rejected on the *presence* of `GPOS`/`GSUB`/`morx`/`mort` and of per-glyph hinting bytecode. Review (devjean-75) measured that this accepted **0 of 8 stock DejaVu faces**, and 41% of the glyphs of the one it would have accepted with layout tables stripped.

ADR-010 decision 4 forbids *on-device* code that walks a font table; it does not require the host's input file to be free of one. So those tables are now **skipped rather than refused** — the walk never slices them, so nothing downstream can bake them — and hinting `instructions[]` are stepped over with a bounds check. `CffOutlinesRejected` (no `glyf` to read) and `CompositeGlyphRejected` (the seam S4 fills when it pre-bakes composites, per ADR-010's "substitutions the baker did not pre-bake") remain.

Measured after the change, running the real parser over the same eight faces:

```
accepted: 8 of 8          (was 0 of 8)

DejaVuSans.ttf         6253 glyphs, 63 zero-length records
    3646 ( 58.3%)  parsed                      (was 41.2%)
    2607 ( 41.7%)  CompositeGlyphRejected
       0          TruncatedGlyph  (was 63)     ← zero-length glyf bug, fixed
       0          HintingRejected (was 1007)
    gid 3 (space): parses as blank
DejaVuMathTeXGyre.ttf  4282 (100.0%) parsed
```

The zero-length fix matters on its own: `loca[i] == loca[i+1]` is the spec's encoding for a blank glyph and is how every real font stores the space character. It was returning `TruncatedGlyph`.

## Local verification of this merge

Built the merged state with `g++ 16.1.0` using the plain flow #164 documents:

```sh
mkdir build && cd build
CC=gcc-16.1 CXX=g++-16.1 cmake .. -G Ninja
cmake --build .          # 366 targets
ctest -E "InstallTreeConsumer|offscreen_tests|ml.noheap.symbolScan"
```

- `mdux_verify_trust_zones()` clean — "2 governed target(s) clean of Vulkan/windowing dependencies"; `MduXTextBakeLib`'s link line is unchanged, so the no-SOUP claim holds.
- **398 / 398 pass**, including all 20 `text_tools_spec` scenarios.
- Exclusions are the standing three on this container: `InstallTreeConsumer` (flaky here), `offscreen_tests` (no Vulkan ICD), `ml.noheap.symbolScan` (needs `nm`/`dumpbin`). None related to this change.

## Next

S3 (#159) — deterministic rasteriser with coverage AA — branches off `develop` once this lands.

## Summary by CodeRabbit

* **New Features**
  * Added host-only TrueType font parsing for standard `glyf` outlines.
  * Added validation for font metadata, glyph contours, coordinates, flags, hinting data, empty glyphs, and supported `loca` formats.
  * Added clear parsing errors for unsupported or malformed font data, including composite and CFF outlines.

* **Tests**
  * Added comprehensive coverage for valid fonts, layout variations, padding, and malformed glyph records.

* **Documentation**
  * Updated architecture documentation to reflect TrueType parser support and current text-pipeline progress.


ambroise-leclerc:
All four findings were valid. Fixed in `dfe4351`.

## The use-after-free was real, and it invalidated this PR's evidence

CodeRabbit's critical finding is correct in every detail. I verified it rather than reasoning about it — built the suite under `-fsanitize=address` before and after the fix:

```
BEFORE
==3041981==ERROR: AddressSanitizer: heap-use-after-free
  READ of size 1 at 0x76b44d1e9550
    #0 beU16       tools/text/Truetype.cpp:77
    #1 beI16       tools/text/Truetype.cpp:82
    #2 parseGlyph  tools/text/Truetype.cpp:428
    #3 operator()  tests/text/TruetypeTests.cpp:1325
  exit=1

AFTER
  21 scenarios, 0 failed — no ASan diagnostic, exit=0
```

Exactly the path predicted. And the decisive part: **that same pre-fix binary passes with sanitizers off**, which is why this deserved a fix rather than a note — the 398/398 I reported was evidence of nothing for the ten cases that read `font.glyf`. Only `GlyphIndexOutOfRange` and `LocaSizeMismatch` return before touching a span.

`State` now owns the buffer and exposes `parseOwning()`, which every case routes through. The comment on it states the lifetime rule so the pattern does not come back.

## The others

- **Copilot — long-format `loca` unpinned.** Correct: the Builder only ever emitted the short form, so the u32 decode branch had never executed, and that is the branch a production font takes once `glyf` passes 128 KB. The Builder now emits either, and a new `longLocaFormatParses` scenario asserts both encodings of the same glyph layout resolve to identical offsets and produce the same outline.
- **CodeRabbit — unreachable trim block.** Correct: `cursor` starts at `directoryEnd` and only advances, so `cursor < directoryEnd` is never true, and the comment described behaviour no fixture relied on. Removed, with a note on what a truncation fixture actually has to do (the `TruncatedTableDirectory` case trims its own buffer).
- **CodeRabbit — contradictory measurement.** Correct, and it was my sentence. It claimed a presence rule "accepts no shipping font" and then referred to "the one font it did accept". Restated as two separate measurements with the method given (with a presence rule: 0 of 8 faces; with layout tables stripped so per-glyph rules become observable: 41% of DejaVuSans' glyphs), and marked explicitly as a local measurement on one developer machine, not a validated result.

**399/399 ctest pass** (up one: the new long-loca scenario). Both toolchain legs were already green before these fixes; they will re-run on this push.

## Worth a follow-up issue, not this PR

Nothing in CI would have caught this. The repo already has `cmake/Sanitizers.cmake`, but no leg runs ASan, and the three parsers that sit on untrusted input — `Spirv.cpp`, `Safetensors.cpp` and now `Truetype.cpp` — are exactly the code where a lifetime bug reads as a pass. An ASan job over the host-tools suites looks cheap and high-value. Happy to open that issue if you want it.

camille-martin-paris (changes requested):
The production parser looks carefully bounded and the current CI suite is green, but the new long-format `loca` coverage has a correctness gap that should be fixed before merge.

In `tests/text/TruetypeTests.cpp`, the builder computes `stored` as `std::uint16_t(cursor / 2u)` and then emits a long-form entry as `std::uint32_t(stored) * 2u`. At a glyph offset of 131,072 bytes—the first offset short `loca` cannot represent—this wraps to zero. As a result, `longLocaFormatParses` only exercises small offsets and cannot validate the reason the 32-bit decoding path exists.

Please emit the full-width `cursor` directly for long format (while preserving the explicit override behavior), and add a case whose glyph starts at or beyond 131,072 bytes. That test should fail with the current builder and demonstrate that the serialized long `loca` entry and parsed glyph offset do not wrap.

ambroise-leclerc:
@camille-martin-paris you're right, and the gap was mine — the long-format coverage I added couldn't reach the offsets the long format exists for. Fixed in `a796c61`.

The narrowing to `uint16` happened first: `uint16(cursor / 2)` truncates at cursor 131072, so `uint32(stored) * 2` wrote 0. Short `loca` tops out at `0xFFFF * 2 == 131070`, making 131072 exactly the first offset it cannot name — so `longLocaFormatParses` only ever exercised offsets the short form could have carried.

The builder now carries the byte offset at full width and narrows only when writing the short form. Explicit overrides keep their meaning — they are raw short-form storage units, so the offset one denotes is twice its value, which is what the `LocaOutOfBounds` fixture's `locaShortEntry(1, 0xFFFF)` depends on.

**Written before the fix, as you asked.** `longLocaReachesPastShortFormLimit` pads glyph 0 to exactly 131072 bytes so glyph 1 starts at the first unreachable offset:

```
against the unfixed builder
  FAIL  A long-format loca addresses a glyph past the 128 KB the short form can reach
  MduX Text Tools Spec: 22 scenarios, 1 failed

after the builder fix
  PASS  A long-format loca addresses a glyph past the 128 KB the short form can reach
  MduX Text Tools Spec: 22 scenarios, 0 failed
```

One thing worth naming from writing it: **a wrapped offset fails no bounds check.** It parses cleanly and returns the wrong glyph — glyph 1 silently resolves to glyph 0's blank record. So the test asserts on the resolved outline (4 points, corner at 100,100) and not only on `loca[1] == 131072`; the offset assertion alone would pin the encoding while leaving the consequence untested.

400/400 ctest pass, and the suite is still clean under `-fsanitize=address`.

PR #172, opened 2026-08-08
S3 — Deterministic rasteriser with coverage AA (#159)

ambroise-leclerc:
Part of #14. Depends on S2 (#158), which landed via #171.

## What this lands

`mdux.text.raster` — a governed-zone glyph rasteriser that turns outlines into an R8 coverage bitmap.

| File | |
|---|---|
| `include/mdux/text/Raster.cppm` / `src/text/Raster.cpp` | module `mdux.text.raster`, registered on `MduXCore` |
| `tests/text/RasterTests.cpp` | 10 SpecLab scenarios, joined to `text_spec` |
| `docs/architecture.md` | records the new governed module and S3 as landed |

## No floating point at all

#159 permits "integer-only **or** float operations encoded as `u32` bit patterns where determinism requires it". This takes the first option throughout, which makes the cross-toolchain byte-identity claim **a property of the source rather than of the compile flags**. Unlike `mdux.ml.kernels` — which must be float, and therefore depends on `cmake/MduXDeterminism.cmake` continuing to keep contraction away from it — nothing arriving through a preset later can weaken this module.

Verified mechanically rather than asserted:

- the only `double` anywhere in the module is inside a comment;
- `objdump -d` over the compiled object reports **zero** floating-point instructions.

The arithmetic that would otherwise be float: scaling is `(unit * pixelSize * kFixedOne) / unitsPerEm` in `int64` with the multiplications first, so there is one truncation rather than two; curve flattening picks a segment count from an integer curvature proxy and an integer Newton square root; edge/scanline intersection is a single `int64` division, **floored rather than truncated** so negative coordinates round like positive ones instead of biasing coverage toward the glyph's left half.

## Coverage cannot leave [0,255], by construction

The acceptance criterion asks for coverage "in [0,255] with no clamping surprises". The way to have no clamping surprises is to have no clamp: a span is intersected with a pixel's own column before being added, so a pixel gains at most `kFixedOne` per sub-scanline across exactly `kSubScanlines` of them. The accumulator ceiling is `kAccumulatorMax` by construction and the final divide lands in range. **There is no `std::clamp` on a coverage value anywhere in the file** — a change that makes one necessary has broken the invariant rather than found a corner case.

## Why it does not take a `truetype::SimpleGlyph`

`mdux.tools.truetype` is host-tools-zone code, and ADR-004 forbids a governed module from importing one — mechanically checked by `mdux_verify_trust_zones()`. So the input is a governed `Outline` carrying the same TrueType shape (points, per-contour end indices, on/off-curve flags), with S4's baker doing a flat conversion. The indirection is the trust-zone boundary made visible, not an abstraction for its own sake.

Being governed is what #159 asks for: if a device path ever needs to rasterise, it imports *this* module rather than growing a second implementation — ADR-008 decision 1, mirrored to text by ADR-010.

## Tests

Assertions are **derived rather than recorded** wherever the geometry permits, because a test that only says "same as last time" cannot distinguish correct output from consistently wrong output:

- a pixel-aligned square is solid **255** — this is what proves full coverage is reachable rather than saturated;
- a rectangle ending mid-pixel reads exactly **127** in the partial column (`2048 * 255 / 4096`);
- a right triangle reads exactly **127** on every pixel its hypotenuse bisects, solid below, empty above;
- a counter-wound inner contour leaves a hole and a same-wound one does **not**, which is what makes the rule nonzero rather than even-odd — an `o` works only because its counter runs against its bowl;
- quadratic flattening, including TrueType's implied on-curve midpoints;
- a degenerate contour yields an **empty** bitmap, not a 1×1 of zeros, because an atlas packer has to tell "nothing to pack" from "one blank pixel to pack";
- coverage reaches exactly 255 at eight different pixel sizes;
- a seven-case rejection corpus pinning each `RasterError`.

Plus `text-raster-determinism-crossToolchain`, labelled `determinism`: a reference glyph with a straight edge, a diagonal, two curves and a counter, at 37 px/em so nothing lands on a pixel boundary, digested to a frozen SHA-256. I froze it only *after* the derived-value scenarios passed and after inspecting the output as a bitmap — it is a plausible glyph with smooth edges and a clean counter, not merely stable garbage:

```
30x32 origin=(2,33)
|         ..++#######+         |
|     .################.       |
|    #######################   |
|   #######+..      #########+ |
|  ########+        ##########.|
| +#########.+++++++##########.|
| ###########################+ |
|.#########################+   |
|.#####+++...                  |
```

**Worth reporting:** two derived assertions failed on first run, and both were my arithmetic rather than the rasteriser's. 3072 font units at upem 2048 is 1.5 *em*, hence 3 pixels at 2 px/em — not 1.5 pixels; and the triangle's diagonal runs top-left to bottom-right in bitmap coordinates, so the pixel I sampled sat above it. The rasteriser produced exactly the right values in both cases, which is the argument for deriving assertions instead of recording them.

## Acceptance against #159

- **Same glyph byte-identical on MSVC and GCC** — `text-raster-determinism-crossToolchain` is the check; this PR's CI is its first cross-toolchain run.
- **`ctest -L evidence` and `-L determinism` pass on both legs** — locally 367/367 and 4/4.
- **Coverage in [0,255] with no clamping surprises** — structural, see above.
- **Unit tests cover corners, curves and edge cases** — 10 scenarios.

## Local verification

`g++ 16.1.0`, plain flow: **410/410 ctest pass**, `mdux_verify_trust_zones()` reports "2 governed target(s) clean", `mdux_verify_fp_determinism()` reports "1 target(s) free of fast-math options".

## Out of scope

No baker calls this yet. S4 (#160) is the wave that converts a `SimpleGlyph` into an `Outline`, packs the bitmaps into an atlas, and commits a font artifact.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic glyph rasterisation that converts font outlines into coverage bitmaps.
  * Added support for quadratic curves, contour validation, bitmap bounds, and top-first pixel output.
  * Added clear error reporting for invalid outlines, unsupported sizes, overflow, and oversized bitmaps.
  * Degenerate or invisible glyphs now produce empty bitmap results.

* **Documentation**
  * Documented the text rasterisation module and marked its initial coverage milestone as complete.

* **Tests**
  * Added coverage for fills, holes, curves, empty outlines, pixel sizes, error handling, and deterministic output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,6 +195,7 @@ target_sources(MduXCore
             include/mdux/ml/Runtime.cppm
             include/mdux/draw/Draw.cppm
             include/mdux/text/Schema.cppm
+            include/mdux/text/Raster.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
@@ -204,6 +205,7 @@ target_sources(MduXCore
         src/ml/Runtime.cpp
         src/draw/Draw.cpp
         src/text/Schema.cpp
+        src/text/Raster.cpp
         src/governance/Governance.cpp
         src/governance/Justification.cpp
         src/governance/Program.cpp
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,9 @@ We follow the **C++ Core Guidelines** to ensure conformance with modern C++23 ge
 - **Functions/Methods:**
   - Use `lowerCamelCase` (e.g., `processData()`, `getValue()`).
 - **Variables (including const, constexpr, and constinit variables):**
-  - Use `lowerCamelCase` (e.g., `dataBuffer`, `isReady`).
+  - Use `lowerCamelCase` (e.g., `dataBuffer`, `isReady`, `maxSweepWork`).
+  - No Hungarian notation, and **no `k` prefix on constants**: `kMaxSweepWork` is wrong,
+    `maxSweepWork` is right. A constant is a variable and follows the same rule as one.
 - **Namespaces:**
   - Use 'lowercase' (e.g., 'mui', 'backend').
 - **Macros:**
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -53,6 +53,7 @@ are ordinary `PRIVATE` sources.
 | `mdux.governance.compliance` | `include/mdux/governance/Compliance.cppm` | `src/governance/Compliance.cpp` |
 | `mdux.shader.schema` | `include/mdux/shader/Schema.cppm` | `src/shader/Schema.cpp` |
 | `mdux.text.schema` | `include/mdux/text/Schema.cppm` | `src/text/Schema.cpp` |
+| `mdux.text.raster` | `include/mdux/text/Raster.cppm` | `src/text/Raster.cpp` |
 | `mdux.draw` | `include/mdux/draw/Draw.cppm` | `src/draw/Draw.cpp` |
 | `mdux.ml.schema` | `include/mdux/ml/Schema.cppm` | header-only |
 | `mdux.ml.kernels` | `include/mdux/ml/Kernels.cppm` | `src/ml/Kernels.cpp` |
@@ -209,7 +210,7 @@ tracking issue; the issue is authoritative for what remains.
 
 | Planned | Issue | Note |
 |---|---|---|
-| Font and text pipeline | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) | S1 (`mdux.text.schema` + `mdux-textbake` skeleton) landed. S2 (`mdux.tools.truetype` glyf parser, #158) landed; S3–S6 open |
+| Font and text pipeline | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) | S1 (`mdux.text.schema` + `mdux-textbake` skeleton) landed. S2 (`mdux.tools.truetype` glyf parser, #158) landed. S3 (`mdux.text.raster` coverage rasteriser, #159) landed; S4–S6 open |
 | `.medui` compiler | [#15](https://github.com/ambroise-leclerc/MduX/issues/15) | the replacement for the retired HTML/CSS path |
 | Rendered-truth verification | [#16](https://github.com/ambroise-leclerc/MduX/issues/16) | beyond the current pixel test |
 | Content components | [#17](https://github.com/ambroise-leclerc/MduX/issues/17) | `SignalTrace`, `StatusIndicator`, `NumericDisplay` and the rest |
--- a/include/mdux/text/Raster.cppm
+++ b/include/mdux/text/Raster.cppm
@@ -1,0 +1,184 @@
+/**
+ * @file Raster.cppm
+ * @brief Governed-zone glyph rasteriser: outlines to an R8 coverage bitmap, integer arithmetic
+ *        only, so the same glyph produces the same bytes on every supported toolchain.
+ *
+ * @compliance ADR-004 Trust zones in C++ (governed zone: std only, no Vulkan, no windowing)
+ * @compliance ADR-005 Error handling and exceptions policy (Result-returning, noexcept)
+ * @compliance ADR-007 Evidence pipeline doctrine (the bytes this produces get committed)
+ * @compliance ADR-008 Zero-SOUP ML inference (decision 1, mirrored to text by ADR-010: one
+ *             module, not a host implementation and a device one that could drift)
+ * @compliance ADR-010 No on-device text shaping
+ *
+ * Part of MduXCore. The font baker (#160, S4) imports this to fill an atlas; if a device path
+ * ever needs to rasterise, it imports *this* module rather than growing a second implementation.
+ * That is ADR-008 decision 1 applied to text, and it is why this file is governed rather than
+ * sitting in `tools/` next to the parser that feeds it.
+ *
+ * ## Why it does not take a `truetype::SimpleGlyph`
+ *
+ * `mdux.tools.truetype` (#158) is host-tools-zone code. ADR-004 forbids a governed module from
+ * importing one, and the rule is mechanically checked by `mdux_verify_trust_zones()` - so this
+ * module cannot name that type even though it is the obvious input. It takes `Outline` instead:
+ * the same TrueType shape (points, per-contour end indices, on/off-curve flags) expressed in
+ * governed types, with the baker doing a flat conversion. That indirection is the trust-zone
+ * boundary made visible, not an abstraction for its own sake.
+ *
+ * ## The determinism rule
+ *
+ * **There is no floating-point arithmetic in this module.** Not "float used carefully", not
+ * "float with `-ffp-contract=off`" - none. Every coordinate is a fixed-point integer, every
+ * division is integer division, and the flattening of a curve into line segments is a pure
+ * function of integer inputs.
+ *
+ * That is a stronger position than `mdux.ml.kernels` takes, and deliberately so. The kernels have
+ * to be float because the model's arithmetic is float, and they therefore depend on
+ * `cmake/MduXDeterminism.cmake` continuing to keep contraction and fast-math away from them. A
+ * rasteriser has no such obligation: coverage is a ratio of areas, areas are counts of subpixel
+ * units, and counts are integers. Choosing integers means the cross-toolchain byte-identity
+ * claim in issue #159 is a property of the source rather than of the flags it is compiled with,
+ * and no future `-ffast-math` arriving from a preset can weaken it.
+ *
+ * Concretely, the arithmetic that would otherwise be float:
+ *
+ * - **Scaling** from font units to pixels is `(unit * pixelSize * kFixedOne) / unitsPerEm`,
+ *   computed in `std::int64_t` and truncated once.
+ * - **Curve flattening** picks a segment count from an integer curvature proxy and an integer
+ *   square root, then evaluates the Bézier at `t = i/N` with the divisions done last, in
+ *   `std::int64_t`.
+ * - **Edge/scanline intersection** is `x0 + (x1 - x0) * (y - y0) / (y1 - y0)`, again in
+ *   `std::int64_t`, with a floor division so that negative coordinates round the same way as
+ *   positive ones instead of toward zero.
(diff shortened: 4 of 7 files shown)


camille-martin-paris (changes requested):
The determinism structure is strong and every current check passes, but I found three production issues that should be corrected before merge.

1. **Quadratic subdivision changes the curve start between samples.** In `buildEdges()`, `evaluateQuadratic(penX, ..., step, count2)` and its Y equivalent are called inside a loop that assigns `penX = nx` and `penY = ny` after every sample. Starting with the second intermediate sample, the evaluator therefore receives the previous sample as `p0` instead of the original curve start. The final point still reaches the endpoint because its `u` term is zero, so endpoint-based tests can pass while curves with `segmentCount >= 3` are distorted. Preserve `curveStartX/curveStartY` for every evaluation and use `penX/penY` only as the emitted segment's previous endpoint. Add a derived test for at least one known intermediate point on a curve requiring three or more segments.

2. **The public `noexcept` API can terminate on ordinary allocation failure.** `buildEdges()` allocates `ring` and `edges`; `rasterise()` allocates `accumulator`, `crossings`, and `bitmap.coverage`. None of these allocations is handled, so `std::bad_alloc` escapes a `noexcept` function and calls `std::terminate`, despite the Result-returning contract and the hostile-input rationale for the size limits. Add an allocation-failure error and translate all allocations reachable from `rasterise()` to it, or remove `noexcept` and document the throwing contract consistently.

3. **Work is not bounded by the existing bitmap-area check.** The raster loop examines every flattened edge for every one of 16 sub-scanlines in every output row. A legal request can approach 65,536 source points (and substantially more flattened edges) and a bitmap can contain up to 64 million pixels, making the current `height * 16 * edge_count` work an effective denial-of-service path long before allocation limits reject it. Bucket edges by their half-open pixel-row ranges, or introduce and document a conservative total-work bound before entering the sweep. Please add a stress/rejection test that pins the chosen bound.

The existing test indexing issue should also be fixed: after `expect(bitmap.width >= 7 && bitmap.height >= 7)`, `at(bitmap, 1, 4)` still executes because `expect` is non-fatal. Guard subsequent indexing so a raster regression reports a failed assertion rather than reading out of bounds.

ambroise-leclerc:
@camille-martin-paris all three were real, and the first one matters more than a code fix — **the digest this PR originally froze pinned a distorted curve.** Fixed in `358a091`.

## 1. Quadratic subdivision used a moving curve start

Exactly as described. I measured the distortion before touching it, on a 5-segment curve, in fixed-point units:

| step | correct | buggy | delta |
|---|---|---|---|
| 2/5 | (327, 1310) | (356, 1576) | **266 = 1.04 px** |
| 3/5 | (737, 1720) | (794, 1972) | 252 |
| 5/5 | (2048, 2048) | (2048, 2048) | 0 — endpoints agree |

That last row is why every assertion I had passed: the final sample's `u` term is zero, so endpoints always land correctly and only the middle is wrong.

The derived test you asked for turned out to be available exactly where the bug lives. For a quadratic, `B(1/2) = (p0 + 2c + p1)/4`; with `p0.x=1024, c.x=2048, p1.x=1024` that is 1536 font units — **exactly 6.0 pixels** at 8 px/em over a 2048 em. So a correct rasteriser produces a bitmap 6 pixels wide, and the buggy one produced 7. That curve takes 6 segments at this size, comfortably past the ≥3 threshold.

Digest re-frozen: **30x32 → 29x32**, one pixel narrower because the curve no longer bulges outward. I re-inspected the corrected glyph as a bitmap before freezing rather than just re-recording it.

## 2. `noexcept` could terminate on allocation failure

Correct, and it contradicted this header's own promise that a corrupt outline yields a diagnostic rather than a `bad_alloc`. The body is now `rasteriseImpl()`, which may throw, behind a `noexcept` entry point returning the new `RasterError::AllocationFailed`. I catch `std::length_error` alongside `bad_alloc` — `resize()` raises it past `max_size()`, the same condition under another name — with a defensive `catch (...)` behind both.

## 3. Sweep work was unbounded

Edges are now bucketed by the pixel rows they span, stored as **CSR** rather than a vector-of-vectors, so a tall glyph does not pay `height` small allocations. Cost becomes proportional to the geometry rather than to `height × 16 × edgeCount`.

Two explicit bounds back it, both reported as `OutlineTooComplex` and both checked **before anything proportional to them is allocated**: `kMaxEdges` on the flattened count, `kMaxSweepWork` on total edge-row visits.

`text-raster-sweep-work-bound` pins it with a comb of 1200 full-height strokes at the maximum pixel size (~4.9M edge-row visits against a 4194304 budget). It also asserts in the same scenario that an ordinary **4096×4096** square at that same size still rasterises, in 0.66s — a bound that rejected legitimate work would be worse than the problem it solves.

## The rest

- Indexing after the non-fatal `expect()` is now inside the guarded branch (you and CodeRabbit).
- The `Crossing` doc had nonzero winding **backwards** — it claimed a same-wound counter makes a hole, which is the case that stays filled. Corrected to match the two scenarios that pin it (Copilot).
- Scenario name and the stray "parser" reference in `kFixedShift`'s doc, both fixed (Copilot).

**411/411 ctest**, `-L determinism` 4/4, `-L evidence` 368/368, clean under `-fsanitize=address` (I re-ran that specifically because the CSR bucketing adds new bounds-driven indexing), and the compiled object still contains **zero** floating-point instructions.

One note for the record: the two bugs that survived my own testing were both invisible to endpoint- and extent-based assertions. The scenarios that would have caught them are the ones asserting a value derived from the geometry — which is the argument for writing more of those, not for adding another digest.

ambroise-leclerc:
Both CodeRabbit findings addressed in `08bddcc`, plus a style violation @ambroise-leclerc caught.

## Coverage fill was unbounded — fixed by removing the cost, not capping it

The finding is correct: `maxSweepWork` counts edge-row visits, which a solid rectangle barely spends (four edges, two vertical). What it spends is *fill*, and that was `height × subScanlines × spanWidth` — bounded by nothing above it.

Rather than add a `maxCoverageWork` budget, I changed the accumulation so the cost cannot arise. Each span is now recorded as **range-adds on a per-row difference buffer**, resolved by one prefix pass at the end of the row. A span is O(1) regardless of width, and total fill becomes **O(bitmap area)** — which `maxBitmapPixels` already bounds. That leaves no new budget to tune and no second limit that could drift out of agreement with the first.

`text-raster-wide-fill-bounded` pins it: 16384 × 512, only 1024 edge-row visits so the sweep budget never notices it, but **134,217,728** per-pixel writes under the old scheme. The whole 33-scenario suite now runs in **0.27s**. It asserts the coverage as well as completing, since a fast wrong answer would otherwise pass.

**The frozen digest is unchanged**, which is the evidence that this refactor is behaviour-preserving rather than just faster.

## Fixture figures — all three corrections hold

I checked the arithmetic rather than taking it on trust, and you're right on every point. The comb is one *closed* contour, so each stroke's return segment also spans the full height, and I had the bitmap dimensions transposed:

```
edge-row visits : 2400 * 4096 = 9,830,400   (budget 4,194,304 — exceeded)
bitmap          : 2398 * 4096 = 9,822,208   (limit 67,108,864 — inside)
```

Behaviour unaffected, as you noted — only my stated numbers were wrong. Corrected, with the two figures written out so the next reader can check them.

## Constants no longer carry a `k` prefix

`CONTRIBUTING.md` already required `lowerCamelCase` for "variables (including const, constexpr, and constinit variables)", so every `kFixedOne` and `kMaxSweepWork` here was a violation from the start — my error, not an ambiguity in the rules. Renamed throughout: eight exported constants, two implementation-internal, three in tests. Digest unchanged, confirming it's cosmetic.

I added the specific prohibition next to the existing rule in `CONTRIBUTING.md`. The rule already covered it, but naming the failure mode makes it harder to repeat.

**Out of scope, flagged not fixed:** `k`-prefixed identifiers remain in about two dozen other files that predate this branch — `src/governance/Compliance.cpp` (19), `src/governance/Governance.cpp` (15), `tests/tools/CliTests.cpp` (13) and others. That's a mechanical repo-wide commit, not something to bury in a rasteriser PR. Happy to do it separately if you want it.

412/412 ctest, `-L determinism` 4/4, clean under `-fsanitize=address`, zero floating-point instructions in the object, docs-lint clean.

ambroise-leclerc:
One nitpick this round, and I'm declining it — with the evidence, so it's easy to overrule me if I've read the convention wrong.

**Finding:** the `///` comments on `frozenDigest` and `determinismPixelSize` should begin with `@brief`.

**Why I'm not applying it:**

1. **The codebase says otherwise, unanimously.** `///` line comments in the governed sources:

   ```
   without @brief : 391
   with @brief    :   0
   ```

   Adding it to exactly these two would make them the only ones in the tree that have it.

2. **The cited guideline doesn't reach them.** There's no `@brief` rule in `.coderabbit.yaml`, so this is inferred from `CONTRIBUTING.md` — whose Doxygen section scopes the requirement to *file-level documentation*, *class documentation* and *method documentation*, and then says "Private members: generally no documentation needed unless complex." These two are file-local `constexpr` constants inside an anonymous namespace in a test file: not public API, not a class, not a method. The file-level block at the top of `RasterTests.cpp` does carry `@brief`, which is the part the rule actually covers.

3. **The proposed diff wouldn't work.** It renders as ``/// `@brief` The digest…`` — backticked, so Doxygen would treat it as literal text rather than a command. Applying it verbatim would produce a comment that looks compliant and isn't.

Happy to reverse this if the intent is that *every* `///` gets `@brief` — but that's a 391-line sweep across the repo rather than two lines here, and worth being a deliberate decision rather than a side effect of this PR.

---

**Status:** all substantive findings from all three rounds are resolved —

| Round | Finding | |
|---|---|---|
| Copilot | winding described backwards, scenario name, "parser" slip | fixed |
| CodeRabbit | `noexcept` vs allocation | fixed (`AllocationFailed`) |
| CodeRabbit | unbounded sweep | fixed (CSR row buckets + two bounds) |
| CodeRabbit | unbounded coverage fill | fixed (per-row deltas, O(area)) |
| CodeRabbit | fixture figures wrong | fixed, arithmetic verified |
| @camille-martin-paris | moving curve start | fixed, derived apex test |
| @camille-martin-paris | non-fatal `expect()` then indexing | fixed |
| @ambroise-leclerc | `k` prefix on constants | fixed, rule documented |

412/412 ctest, both toolchain legs green, clean under `-fsanitize=address`, zero floating-point instructions in the object. The only thing outstanding is @camille-martin-paris's CHANGES_REQUESTED, which needs their re-look rather than another commit from me.

PR #173, opened 2026-08-08
S4 prerequisite (1/2) — cmap and hmtx in mdux.tools.truetype (#160)

ambroise-leclerc:
Part of #14, prerequisite for S4 (#160). **First of two stacked PRs.**

> ### Reviewers: what is deliberately not here
>
> **This PR adds no atlas packing, no baking, no committed font, and no recipe changes.** All of that is in the second PR of the stack, which targets *this* branch:
>
> | | |
> |---|---|
> | **1/2 — this PR** | `cmap` + `hmtx` in `mdux.tools.truetype`. Table decoding only. |
> | **2/2 — follows** | Shelf atlas packer, `mdux-textbake` completion, `TXT` diagnostics, `mdux_bake_artifact()` registration, the committed DejaVu Sans source asset and `generated/font/<id>/`, SOUP register entry. |
>
> So: no `AtlasPacker`, no `atlas.bin`, no `package.json` for a font, and `TextBake.cpp` is untouched. If you are looking for the thing that *uses* `glyphForCodePoint()`, it is in 2/2 — nothing in this branch calls it outside the tests, by design.
>
> The split exists because S3's 600-line change attracted four real defects in review; folding a table decoder into a baker would have made both harder to see.

## Why this is not in #160's text

S4 asks for glyphs packed into an atlas and a committed font package. Neither was expressible: the parser walked `head`, `maxp`, `loca` and `glyf` only, so there was

- **no code-point-to-glyph mapping** — a recipe naming a charset had nothing to resolve it against, and
- **no advance widths at all** — though ADR-010 defines the baked unit as "one glyph baked at a known atlas slot **with a known advance**" and lists "glyph advances" among a font package's contents.

The epic's chain (S2 → S3 → S4) doesn't capture this. I found it while planning S4 and stopped rather than discover it mid-baker.

## `cmap`

Both Unicode formats a real font carries, normalised into one sorted `CmapRange` list:

- **format 4**, the BMP workhorse — four parallel arrays plus an `idRangeOffset` that is a byte offset measured *from its own slot*, a self-relative pointer into the trailing glyph array. That indirection is the fiddliest arithmetic in the format; it is walked exactly once here so its bounds check lives in one place rather than at every lookup.
- **format 12**, the only one of the two that reaches past the BMP.

Preference is (3,10) format 12 → (3,1) format 4 → Unicode-platform. A font carrying only a Macintosh (1,0) table is **refused**, not decoded: that table maps bytes through a legacy codepage, and guessing which codepage is exactly the inference this parser exists not to make.

An unmapped code point returns `nullopt`, **not glyph 0**. Returning `.notdef` would let a baker fill an atlas with tofu boxes and report a successful bake.

## `hmtx`

Sliced, not decoded — `metricsFor()` reads one glyph's pair on demand, the way `parseGlyph()` reads one `glyf` record. A 6253-glyph font shouldn't pay for 6253 metric structs when a recipe asks for ninety-five.

Glyphs at or past `numberOfHMetrics` inherit the last advance and carry only a bearing. That's the format's compression for a trailing run of equal-width glyphs, not an error.

## Verified against a real font, not only fixtures

DejaVuSans 2.37 through this parser:

```
parsed: numGlyphs=6253 upem=2048 cmapRanges=281 numberOfHMetrics=6238
ASCII 0x20-0x7E: mapped=95/95  outlines ok=95  composites=0
  'A'     gid=36 advance=1401 lsb=16
  'space' gid=3  advance=651  lsb=0
  '0'     gid=19 advance=1303 lsb=135
  'W'     gid=58 advance=2025 lsb=68
```

Two things fall out that matter for 2/2. **Printable ASCII is entirely composite-free in DejaVuSans**, so the first baked artifact needs no composite pre-baking — which is why 2/2 can ship without touching `CompositeGlyphRejected`. And `numberOfHMetrics` (6238) is below `numGlyphs` (6253), so a real font exercises the inherited-advance path rather than just the fixture.

## Tests

The `Builder` now emits `cmap`, `hhea` and `hmtx` by default, so every pre-existing scenario keeps working, with controls for the new rejection paths.

- **`text-truetype-cmap-resolves`** builds the same mapping in *both* encodings and asserts they agree code point by code point, plus that points either side of the run and an unmapped CJK point resolve to nothing. Formats 4 and 12 share no code, so their agreement is the assertion that carries weight.
- **`text-truetype-hmtx-metrics`** uses a six-glyph font with four metric pairs: the first four advances differ, the last two inherit the fourth.
- **Seven new rejection rows** — `MissingCharacterMap`, `MissingHorizontalMetrics` (hhea and hmtx separately), `TruncatedHhea`, `TruncatedHmtx`, `UnsupportedMetricCount` at both ends.

`cmap` and `hmtx` are now **required** tables with their own codes rather than folding into `MissingRequiredTable`, so an author learns which capability is missing.

24 scenarios, **414/414 ctest**, clean under `-fsanitize=address`.

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving Unicode characters to glyphs from TrueType character maps in formats 4 and 12.
  * Added retrieval of glyph advance widths and side bearings, including shared metric handling.
* **Bug Fixes**
  * Improved validation and error reporting for missing, truncated, oversized, unsupported, or malformed character-map and horizontal-metrics data.
  * Added safeguards for invalid metric counts and out-of-range glyph or character lookups.

### DIFF
--- a/tests/text/TruetypeTests.cpp
+++ b/tests/text/TruetypeTests.cpp
@@ -55,6 +55,10 @@ constexpr std::uint16_t firstAdvance         = 500;
 constexpr std::uint16_t advanceStep          = 10;
 constexpr std::int16_t  sideBearing          = 7;
 
+/// The glyph id the truncated-subtable fixture's filler decodes to. In range, so a decoder that
+/// reads past the declared length produces a plausible wrong answer rather than an obvious one.
+constexpr std::uint16_t strayGlyphInFiller = 5;
+
 // ---------------------------------------------------------------------------
 // An assembler for in-memory TrueType files, plus helpers for the glyf records each scenario
 // uses. A real `.ttf` would be the same bytes in a different file; building them in code keeps
@@ -130,6 +134,21 @@ public:
     }
     Builder& dropHmtx() {
         dropHmtx_ = true;
+        return *this;
+    }
+
+    /// Emits the format 4 segment through `idRangeOffset` - the self-relative indirection into a
+    /// trailing glyph array - instead of the affine `code + idDelta` form. Real fonts use both,
+    /// and the two take completely different paths through the decoder.
+    ///
+    /// `truncateGlyphArray` declares a subtable `length` that stops before the glyph array the
+    /// segment needs, and appends filler bytes after it standing in for whatever subtable would
+    /// follow in a real font. A decoder that slices the subtable to the end of `cmap` rather than
+    /// to its declared length reads that filler as glyph ids and returns *wrong mappings* with no
+    /// diagnostic; one that clips correctly refuses with TruncatedCmap.
+    Builder& cmapIndirect(bool truncateGlyphArray = false) {
+        cmapIndirect_          = true;
+        cmapTruncateGlyphArray_ = truncateGlyphArray;
         return *this;
     }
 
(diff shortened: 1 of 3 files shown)


ambroise-leclerc:
Stack pointer: **2/2 is [#174](https://github.com/ambroise-leclerc/MduX/pull/174)** (`mdux.tools.atlaspacker`), which targets this branch rather than `develop`.

If you are reviewing this PR and wondering *what calls* `glyphForCodePoint()` or `metricsFor()` — nothing does yet, by design. #174 adds the packer, and the baker that consumes both is a further PR I have flagged for @ambroise-leclerc's decision at the bottom of #174.

Merge order is #173 → #174.

ambroise-leclerc:
All three addressed in `25f154c`. One of them produced **silently wrong mappings**, which is the worst failure mode this parser has.

## Subtable sliced to the end of `cmap` instead of its own length (@Copilot)

Correct, and the consequence is worse than a bounds problem. `decodeCmap()` used `cmap.subspan(*best)`, which runs to the end of the whole table. Format 4's `idRangeOffset` is a self-relative pointer into a glyph array that the **subtable's own length** is what terminates — so an overlong slice lets that indirection read into whatever subtable follows.

Every read stays inside `cmap`, so no bounds check fires. The decoder just returns someone else's glyph ids and reports success. A font with several cmap subtables — the normal case — would draw the wrong characters.

Both formats now clip to their declared length, which they store differently: format 4 as a `uint16` at offset 2, format 12 as a `uint32` at offset 4.

The regression test asserts the **wrong answer**, not merely the absence of a crash. Its fixture declares a length covering four glyph-array entries, needs twenty-six, and fills the rest with an in-range id so a decoder reading past produces a confident wrong mapping:

```
before  FAIL ... it parsed, and U+004B resolved to 5 - read out of the next subtable
                 (the correct glyph is 11)
after   PASS ... refused with TruncatedCmap
```

## `segCountX2` parity (@Copilot)

Right. An odd value isn't a slightly-wrong count — it means the four parallel arrays aren't where the layout arithmetic computes them to be, so every later offset addresses the wrong field and the table is **misread rather than rejected**. Now refused as `TruncatedCmap`.

## The wrapping bounds check (@coderabbitai)

Right on both counts, including that nothing was reachable: `*offset + 2u` wraps in 32-bit, but `beU16()` would still have refused the read. Widened to 64-bit anyway — a bounds check that doesn't hold on its own terms is one nobody can rely on later.

## The test gap this exposed

Worth naming: the Builder only ever emitted **affine** format-4 segments, so the `idRangeOffset` indirection — the fiddliest arithmetic in the format, and the code the first finding is about — had **no test at all**. That's why a bug in it survived my own review. The Builder now emits both forms, and `text-truetype-cmap-indirect` asserts the indirect segment maps the same run as the affine one.

Re-validated against DejaVuSans after the clipping change, since that's the one that could plausibly have broken a real font: 281 ranges, ASCII 95/95 mapped with 95 outlines, identical glyph ids (U+0041→36, U+004B→46, U+0030→19, U+007E→97).

26 scenarios, **416/416 ctest**, clean under `-fsanitize=address`.

PR #174, opened 2026-08-08
S4 (2/3) — deterministic shelf atlas packer (#160)

ambroise-leclerc:
Part of #14, S4 (#160). **Second of three stacked PRs.**

> ### Stack
>
> | | | |
> |---|---|---|
> | **1/3** | [#173](https://github.com/ambroise-leclerc/MduX/pull/173) — **merged** | `cmap` + `hmtx` in `mdux.tools.truetype`: code-point→glyph mapping and advance widths |
> | **2/3** | this PR | `mdux.tools.atlaspacker` — deterministic shelf packing |
> | **3/3** | follows, targets this branch | `mdux-textbake` completion, recipe font fields, `TXT` diagnostics, `mdux_bake_artifact()`, the committed DejaVu Sans asset, `generated/font/<id>/`, SOUP register entry |
>
> Now that 1/3 has merged, this PR targets `develop` and its diff is exactly the packer: five files, nothing else.
>
> **What is deliberately absent, and where it went:**
>
> - **Anything that calls this packer.** Nothing outside the tests invokes `pack()` — the baker is the caller and it is in 3/3. That absence is the most likely thing to read as an oversight, so: it is not one.
> - **`cmap`/`hmtx`, code-point resolution, advance widths** → #173, merged.
> - **`TextBake.cpp`, the recipe, `atlas.bin`, any committed font binary** → 3/3.

## Host-tools, not governed

`mdux.text.raster` (#159) is governed because a device path could conceivably need to rasterise, and ADR-008 decision 1 says that would have to be *that* module rather than a second one. Packing has no such future: an atlas is chosen once at build time and a device consumes the slot rectangles the baker recorded. Nothing on a device ever packs, so a governed packer would be a promise with no caller — and every governed module is one a reviewer has to reason about for device behaviour.

## Shelf, not skyline — and why that is the right trade here

Skyline packing wastes less area. It is not used, deliberately.

This sheet is **committed and byte-compared across toolchains on every CI run** (ADR-007). A packer whose output depends on a heuristic's tie-breaking is one whose diff a reviewer cannot predict from the recipe. Shelf placement's output follows from the input in a way you can check by hand, and the atlas is not the size constraint in any realistic font package.

The sort key is **(height, width, id)**, not height alone. Height alone would leave equal-height glyphs free to swap depending on whether `std::sort` happened to be stable — making the committed bytes a property of the standard library implementation. The triple is a total order, so placement is a pure function of the glyph set.

Each candidate sheet size **re-packs from scratch** rather than growing incrementally, so a sheet packed at 512 wide is identical whether or not 256 was tried first.

One pixel of padding between glyphs: a sampler filtering at a slot edge reads the neighbouring texel, and without a gap that texel belongs to a different glyph. That matters once #162 samples `CoverageR8` with linear filtering.

**The search is bounded to sheets at least as wide as they are tall.** A set that would fit 64×128 gets 128×128, so the result is the smallest area among wide-or-square candidates rather than the smallest possible — stated explicitly after review, because the docs originally claimed a minimum the search does not look for.

## Tests — properties, not a digest

A digest would pin every property at once and tell you nothing about which one broke, so each is asserted directly:

- **power-of-two edges, everything inside the sheet, and no two glyphs overlapping.** Overlap is checked pairwise rather than assumed: a packer that stacked two glyphs would produce an atlas that looks plausible, packs *efficiently*, and renders one of them as garbage.
- **the same glyph set packs identically in three input orders** (forward, reversed, pairwise-swapped). This is what the (height, width, id) sort exists for, and what makes a committed atlas a function of the charset rather than of how the recipe happened to list it.
- **fifty blank glyphs do not change the sheet**, but each still gets a slot — a space must be recorded as blank, not as absent, or the package reads as "this character was not baked".
- **six rejection rows**: `NoGlyphs`, `DuplicateGlyphId`, `GlyphTooLarge` (no sheet size can help, so it is distinct from the budget code), `AtlasBudgetExceeded`, and two near-`UINT32_MAX` extents added after review — the original guard used addition and wrapped, so the very largest glyph packed successfully instead of being refused.

30 scenarios in `text_tools_spec`, **420/420 ctest**, clean under `-fsanitize=address`.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic glyph atlas packing for text rendering.
  * Supports power-of-two atlas layouts, glyph padding, blank glyphs, occupancy reporting, and stable slot ordering.
  * Provides clear handling for duplicate IDs, oversized glyphs, allocation limits, and invalid packing requests.

* **Tests**
  * Added coverage for atlas dimensions, containment, overlap prevention, deterministic layouts, blank glyphs, and error reporting.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -588,6 +588,7 @@ add_executable(text_tools_spec
     text/TextToolsSpecMain.cpp
     text/TextBakeTests.cpp
     text/TruetypeTests.cpp
+    text/AtlasPackerTests.cpp
 )
 
 target_link_libraries(text_tools_spec PRIVATE MduX::TextBakeLib speclab::speclab)
--- a/tests/text/AtlasPackerTests.cpp
+++ b/tests/text/AtlasPackerTests.cpp
@@ -1,0 +1,279 @@
+/**
+ * @file AtlasPackerTests.cpp
+ * @brief BDD scenarios for the host-only shelf atlas packer (issue #160).
+ *
+ * @compliance ADR-004 Trust zones in C++ (host-tools zone)
+ * @compliance ADR-007 Evidence pipeline doctrine
+ *
+ * The properties asserted here are the ones a committed artifact depends on, and each is checked
+ * directly rather than through a digest: dimensions are powers of two, no two glyphs overlap, no
+ * glyph escapes the sheet, and the same input produces the same layout however it was ordered on
+ * the way in. A digest would pin all four at once and tell you nothing about which had broken.
+ *
+ * Overlap in particular is worth testing explicitly rather than trusting: a packer that placed two
+ * glyphs on top of each other would produce an atlas that looks plausible, packs efficiently, and
+ * renders one of the two as garbage.
+ */
+
+import std;
+import speclab;
+import mdux.core.result;
+import mdux.tools.atlaspacker;
+
+#include "../framework/SpecLabBridge.hpp"
+
+namespace {
+
+namespace ap = mdux::tools::atlas;
+using ap::PackError;
+
+/// A spread of sizes with the awkward cases a real font produces: a tall narrow 'l', a wide 'W',
+/// square digits, and a blank space. Deliberately not uniform - a packer only shows its shelf
+/// behaviour when heights differ.
+[[nodiscard]] std::vector<ap::GlyphExtent> mixedGlyphs(std::uint32_t count = 40) {
+    std::vector<ap::GlyphExtent> extents;
+    extents.reserve(count);
+    for (std::uint32_t i = 0; i < count; ++i) {
+        // A deterministic spread, not a random one: the layout is committed, so the fixture that
+        // produces it has to be reproducible from its own source.
+        const std::uint32_t w = 4 + (i * 7) % 23;
+        const std::uint32_t h = 5 + (i * 11) % 17;
+        extents.push_back(ap::GlyphExtent{.id = i, .width = w, .height = h});
+    }
+    extents.push_back(ap::GlyphExtent{.id = count, .width = 0, .height = 0});  // the space
+    return extents;
+}
+
+[[nodiscard]] bool isPowerOfTwo(std::uint32_t v) noexcept {
+    return v != 0 && (v & (v - 1)) == 0;
+}
+
+/// Reports the first overlapping pair, if any, as a readable message.
+[[nodiscard]] std::optional<std::string> findOverlap(const ap::AtlasLayout& layout) {
+    for (std::size_t a = 0; a < layout.slots.size(); ++a) {
+        const auto& lhs = layout.slots[a];
+        if (lhs.width == 0 || lhs.height == 0) {
+            continue;
+        }
+        for (std::size_t b = a + 1; b < layout.slots.size(); ++b) {
+            const auto& rhs = layout.slots[b];
+            if (rhs.width == 0 || rhs.height == 0) {
+                continue;
+            }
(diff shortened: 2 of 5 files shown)


ambroise-leclerc:
All three addressed in `22d3452`. This branch also picked up #173's fixes via a merge, so its diff stays reviewable on its own.

## The size guard could be bypassed by exactly the inputs it exists to catch (@Copilot)

Correct. `extent.width + glyphPadding > maximumAtlasEdge` is uint32 arithmetic — at a width near `0xFFFFFFFF` the addition wraps to a small number, the comparison passes, and `placeAt()` then computes a wrapped `needWidth` from the same expression.

Now `extent.width > maximumAtlasEdge - glyphPadding`, which cannot wrap because the constant exceeds the padding. I added a `static_assert` pinning that relationship rather than leaving it as a fact about two constants somebody could later change independently.

Demonstrated both ways with a `0xFFFFFFFF`-wide glyph in the corpus:

```
before  FAIL - a glyph whose width is near UINT32_MAX: packing succeeded unexpectedly
after   PASS - refused with GlyphTooLarge
```

Both axes are covered, since the two comparisons are separate expressions and fixing one wouldn't imply the other.

## The docs claimed a minimum the search doesn't look for (@Copilot, both sites)

Also correct. The inner loop is bounded by `height <= width`, so sheets taller than they are wide are never considered — a set that would fit 64×128 gets 128×128. Saying "smallest power-of-two sheet" overstated that.

The bound is deliberate for a texture atlas, so I fixed the claim rather than widening the search. The interface now says the result is the smallest area **among wide-or-square candidates**, and the loop comment carries the 64×128 example so the cost is concrete rather than implied.

**420/420 ctest**, 30 scenarios, clean under `-fsanitize=address`.

ambroise-leclerc:
Stack pointer: **3/3 is [#176](https://github.com/ambroise-leclerc/MduX/pull/176)** — the baker, the recipe, the committed DejaVu Sans asset and `generated/font/dejavu-ui/`. It targets this branch.

That is where `pack()` finally gets a caller, so if you are reviewing this PR and wondering what consumes it, #176 is the answer.

Merge order: #174 → #176.

ambroise-leclerc:
Both addressed in `795cc11`.

## The sheet is a first fit, not a minimum (@coderabbitai)

You're right, and this is the second time I got the wording wrong. Last round I changed "smallest power-of-two sheet" to "smallest area among wide-or-square candidates" — still false, for exactly the reason you give: candidates are enumerated width-major, so **area is not monotonic along the search**. 256×256 (65536 px) is reached before 512×64 (32768 px), and a set fitting both takes the larger.

I kept the search and fixed the claim, rather than reordering by ascending area. "Squarish" is the property worth having in a texture atlas and "minimal area" is not — a 512×64 strip wastes no pixels but is a worse shape to sample from and to fit under a device's texture limit. What ADR-007 needs is that the choice be a *pure function of the glyph set*, which a first fit in a fixed order already is.

So the docs now stop claiming minimality at all. The module doc and the loop comment both give the enumeration order and your 256×256-over-512×64 example, and `pack()` reads "the first that holds them in a width-major enumeration, which is not the same as the smallest". I noted in the comment that this is the third wording, since the first two both claimed a minimum of some kind and neither was true.

## The catch block had its rationale backwards (@Copilot)

Correct — it said "`pack()` is not noexcept" when it is, in both the interface and the definition. And that is precisely *why* the catch exists: the vectors can throw `bad_alloc`, and an exception escaping a noexcept function calls `std::terminate`. Corrected to say so, matching what `raster::rasterise()` documents at its own noexcept boundary.

No behaviour change; 30 scenarios still pass.

**Note on the branch:** GitHub rebased this branch when #173 merged, so my local copy had diverged with the same content under different commit ids. I realigned onto the remote and replayed only the new commit rather than force-pushing my copy over yours — the history you see is GitHub's rebase plus one doc commit on top.

PR #176, opened 2026-08-08
S4 (3/3) — font baker and the first committed font package (#160)

ambroise-leclerc:
Part of #14, S4 (#160). **Third of three stacked PRs — targets #174, not `develop`.**

> ### Reviewers: read this first
>
> | | | |
> |---|---|---|
> | **1/3** | [#173](https://github.com/ambroise-leclerc/MduX/pull/173) — **merged** | `cmap` + `hmtx`: code-point→glyph mapping and advance widths |
> | **2/3** | [#174](https://github.com/ambroise-leclerc/MduX/pull/174) | `mdux.tools.atlaspacker` — deterministic shelf packing |
> | **3/3** | this PR | the baker, the recipe, `TXT` diagnostics, the committed font and artifact |
>
> **This PR's base is #174's branch**, so its diff is only the baker and the artifact. Review #173 and #174 first, or the calls into `glyphForCodePoint()`, `metricsFor()` and `pack()` will look like they came from nowhere.
>
> **If something looks missing, it is probably in 1/3 or 2/3:** the parser's table decoding, the packer's shelf algorithm and their tests all live there. This PR adds no parsing and no packing — it wires them together and commits what falls out.

## The pipeline, end to end

Read the `.ttf` → resolve the charset through `cmap` → read each advance from `hmtx` → parse and rasterise each outline → pack → blit into one R8 sheet → emit `package.json`, `report.json`, `atlas.bin`.

A recipe carrying `[charset]` is a **font** recipe; one without it is the text recipe S1 already handled. They share `parseRecipe()`, `run()`, `write()` and `verify()` because they share everything around the edges and differ only in what fills the sidecar — so `mdux-textbake` stays one tool with one CLI, which is what #160 asks for.

From the committed artifact:

```
atlas 128x128 = 16384 bytes, occupancy 54%, 95 glyphs
' '  gid=  3 slot=(  0,  0)  0x 0 adv= 651
'A'  gid= 36 slot=( 13, 46) 11x12 adv=1401 lsb= 16
'W'  gid= 58 slot=( 90, 32) 16x12 adv=2025 lsb= 68
```

And the coverage is legible as letterforms, not merely non-zero — which is the check that the whole chain works rather than each link separately:

```
'A' (11x12)          'g' (9x13)           '5' (8x13)
|    -*-    |        |  -#@#:++|          |.****** |
|   :@+@.   |        | ##   .@*|          |:@.     |
|  -@. .@-  |        | :@*:-%%*|          |:@***-  |
| +@@@@@@@= |        |  -%@%-#*|          |      @*|
|*@       @*|        |  @@%@@- |          |*@@@@%: |
```

## The space nearly broke it

The first bake failed on U+0020 with `EmptyOutline`. That is the rasteriser being **right**: an empty outline span is a caller mistake everywhere except here, where it is the font correctly saying the character draws nothing.

Relaxing `rasterise()` would have made a genuinely malformed call indistinguishable from a space, so the baker recognises blanks instead and gives them a zero-area slot that still carries an advance. The space appears in the package rather than being omitted — a missing entry would read as "not baked" rather than "blank".

Found by *running* the tool, not by reading it. That is the argument for baking a real font in this child rather than a fixture.

## The committed font, and a licence correction

**DejaVu Sans, unmodified, 742 KB.** Chosen after Roboto turned out not to be Apache-2.0 any more — `google/fonts` lists it under `ofl/roboto` with `license: "OFL"`, `googlefonts/roboto-3-classic` reports OFL-1.1, and Google Fonts ships only a variable build. Recording it as Apache-2.0 would have put a false licence claim in the SOUP register.

DejaVu's font files are Bitstream Vera; only the Debian *packaging* is GPL-2+, and that is not redistributed here. `LICENSE-DejaVu.txt` carries the font terms with that stanza **deliberately omitted** rather than copied across where it could be mistaken for a term on the `.ttf`. The naming restriction binds modified versions; the font is not modified, so it is not engaged. Registered in `docs/governance/soup-register.toml`.

**Size, flagged rather than buried:** 742 KB against a 13 KB largest-existing committed asset. That is what fonts cost, but it is a permanent step change in the repo's binary weight.

**Printable ASCII only**, and that is load-bearing: all 95 code points are simple outlines in DejaVu Sans — measured, 95/95 with zero composites — so the charset needs no composite pre-baking, which the parser refuses and #160 does not add.

## Diagnostics

Seven new `TXT` codes rather than one "bake failed", because an author needs to know whether the charset, the font, the size or the packer is at fault: `TXT011` malformed charset, `TXT012` unreadable font, `TXT013` unparsable font, `TXT014` charset not in font, `TXT015` outline rejected (where composites land), `TXT016` raster failure, `TXT017` packing failure.

A code point the font cannot draw is **fatal, not skipped**: silently omitting it would produce a package whose charset is not the one the recipe asked for, and the `.medui` compiler (#15) would then trust it.

## Verification

- `mdux_bake_artifact()` registered, so `ctest -L evidence` byte-compares the committed artifact. **Confirmed it bites**: flipping one byte in `atlas.bin` makes `evidence.font.dejavu-ui` fail; restoring it passes.
- Two new scenarios — the end-to-end *shape* a digest cannot describe (95 glyphs, power-of-two sheet, exactly one byte per texel, the space blank but advancing), and a rejection corpus pinning `TXT011` and `TXT014`.
- **423/423 ctest**, 32 scenarios in `text_tools_spec`, docs-lint clean.

## Against #160's acceptance criteria

- ✅ `mdux-bake-update` produces the reviewed artifacts
- ✅ atlas dimensions are power-of-two, and the packer rejects an over-budget set with a stable code (#174)
- ⏳ `ctest -L evidence` byte-identical on MSVC and GCC — green locally on GCC; **this PR's CI is the first MSVC run**
- ✅ no normal build writes a tracked file

## Summary by CodeRabbit

* **New Features**
  * Added support for baking TrueType fonts into UI-ready font packages.
  * Added the DejaVu Sans UI font with printable ASCII coverage, 16px rendering, and packed atlas output.
  * Added validation and diagnostics for invalid font sizes, missing glyphs, charset mismatches, and unsupported ranges.

* **Documentation**
  * Documented the font pipeline, roadmap progress, licensing, provenance, and redistribution requirements.

* **Tests**
  * Added end-to-end coverage for successful font baking, atlas generation, metadata, blank spaces, and error handling.


ambroise-leclerc:
Both addressed in `75bfc1d`. One was a silent misconfiguration.

## `pixelSize` was narrowed before it was checked (@Copilot)

Correct, and the consequence is worse than a bad diagnostic. `static_cast<std::uint32_t>(...asInteger())` ran *before* the `1..maxPixelSize` range check, so a value outside uint32 wrapped into a valid one and baked — `4294967312` (2³² + 16) and `-4294967280` both arrive as **16**.

The artifact is then produced at a size nobody wrote, and it looks deliberate. Verified both ways with 2³² + 16 in the corpus:

```
narrowing first  FAIL - the bake succeeded unexpectedly, and no diagnostic was emitted at all
checking first   PASS - refused with TXT018
```

The check now runs against the `int64` the TOML parser actually produced, and the narrowing happens after it. A negative size is covered separately, since the two wrap through different paths.

## An invalid size reported the charset's code (@Copilot)

Also right. It emitted `TXT011` (`recipeCharsetMalformed`), so an author who mistyped a size was sent to look at `[charset]` — which undermines the whole reason the font pipeline has seven codes instead of one "bake failed". Added `TXT018`, and corrected the comment above the code list, which described a separation the code did not implement.

**The committed artifact is byte-identical after the change**, which is the check that this touched validation and not the bake.

423/423 ctest, 380 evidence tests, 32 scenarios.

Branch note: this branch is merged up from #174's rebased head, so it carries the packer doc corrections too.

ambroise-leclerc:
@coderabbitai review

This PR's base was retargeted from `160-atlas-packer-and-font-baker` to `develop` when #174 merged, so the recorded "reviews are disabled for this base branch" result predates the change. The diff is now the baker and the committed artifact only — the packer landed with #174.

camille-martin-paris (changes requested):
Found three recipe-validation issues that should be addressed before merge:

1. **Overlapping charset ranges produce duplicate code-point records (P1).** In `tools/text/TextBake.cpp:238-247`, each range is validated independently, but ranges are never checked against one another. For ranges such as A-Z and A-F, the baker emits duplicate package entries for the overlapping code points, assigns them separate atlas slots, and leaves downstream lookup ambiguous. Please reject overlapping ranges with TXT011, including duplicate single-code-point ranges.

2. **Surrogate code points are accepted as Unicode characters (P2).** The validation at `tools/text/TextBake.cpp:239` accepts every integer through 0x10FFFF, including U+D800-U+DFFF. These are reserved UTF-16 surrogate values rather than Unicode scalar values and should not appear as package code points. Please reject any range that intersects the surrogate block (or explicitly split it if that is an intended supported behavior).

3. **Empty locale identifiers bypass the existing no-locale validation (P2).** In `tools/text/TextBake.cpp:190-205`, only an empty locale vector is rejected. A recipe with `ids = [""]` succeeds and emits `"locales": [""]`, although the equivalent text-recipe locale must be non-empty. Please validate every locale entry; rejecting duplicates would also prevent redundant approval entries.

ambroise-leclerc:
All six addressed in `a537c94` — three from @camille-martin-paris, three from @coderabbitai. Two were licensing or security rather than validation.

## The licence file was missing a required notice (@coderabbitai)

This is the one I'm most glad was caught. My extracted file **claimed** "with the Arev additions" while the Debian machine-readable copyright it was extracted from contains **no Arev notice at all** — `grep -ci "tavmjong\|arev"` on it returns **0**. So it asserted a licence component whose notice it did not carry, and both the Bitstream and Arev licences require their notice to accompany redistributed copies.

Tidying the file to drop the GPL-2+ packaging stanza was the right instinct applied the wrong way: I hand-extracted instead of using the upstream file. Replaced with the canonical upstream `LICENSE`, verbatim and whole — 21 Bitstream mentions, 7 Arev, 8 Tavmjong Bah, and **zero** GPL text, which was the original concern. Shipping it whole removes the judgement call about what may be dropped.

## The font path could leave the repository (@coderabbitai)

Correct, and worse than it looks. `root / spec.source` is not containment: `std::filesystem::operator/` **replaces** the left operand when the right is absolute, so `source = "/etc/shadow"` read that file and ignored `root` entirely.

Absolute paths are now refused, and the joined path is normalised and checked with `lexically_relative()` so `a/../../b` fails rather than passing a string-prefix test. Not a live attack surface — the recipe is repository content — but a baker that will one day run over a recipe somebody else wrote shouldn't have "the recipe is trusted" as its only defence.

## Overlapping charset ranges (both of you)

Right. `A-Z` plus `A-F` baked the overlap twice: two atlas slots and two `glyphs` entries for one code point, leaving a consumer that indexes by code point with an ambiguous package. Refused with `TXT011`, naming both ranges. The check runs on a *sorted copy* so `spec.charset` keeps the recipe's order, which is what `package.json` is emitted in.

## Surrogates were accepted as characters (@camille-martin-paris)

Also right — U+D800..U+DFFF passed the `<= 0x10FFFF` bound. They're UTF-16 surrogate code *points*, not scalar values, so a package listing one describes a glyph for something that cannot appear in text. Refused rather than silently split; splitting would quietly bake a charset other than the one written.

## Empty and duplicate locale ids (@camille-martin-paris)

Only an empty *list* was rejected, so `ids = [""]` emitted `"locales": [""]` — which reads downstream as a locale rather than as the mistake it is, while the equivalent text-recipe locale has always had to be non-empty. Every entry is checked now, and duplicates too, since a second identical approval records nothing.

## Two slot lookups that disagreed (@coderabbitai)

`composeAtlas()` and `fontPackageJson()` each rebuilt the same map and diverged on a missing id — one skipped, the other called `.at()` and **threw out of a baker whose contract is to return diagnostics**.

Built once in `runFontBake()` and validated there: every glyph has exactly one slot, the slot matches its bitmap's extent, and it lies inside the sheet. Both call sites now share one guarantee instead of each inventing a policy. The blit loops are driven by the bitmap's extent rather than the slot's, since that's the loop indexing `bitmap.coverage`.

---

Six new rejection rows cover overlap, surrogates, both locale shapes and both path escapes.

**The committed artifact is byte-identical throughout**, which is the check that all of this touched validation and not the bake. 423/423 ctest, 380 evidence tests, docs-lint clean.

ambroise-leclerc:
Three more addressed in `710e113`, including one I'd missed from your previous round.

## The licence file contained text of ours (@coderabbitai)

Fair, and precisely put. The upstream *license text* was verbatim; the **file** was not, because the first fourteen lines were mine and the header claimed otherwise.

`LICENSE-DejaVu.txt` is now the upstream file **byte for byte** — `diff` against the fetched original reports no difference — with no header, note or edit. Provenance moved to `recipes/font/dejavu-ui/README.md`, where it cannot be mistaken for a licence term.

That README records both earlier mistakes, because both looked tidier than they were: first an extraction that dropped the Arev notice while claiming it, then the upstream text under a header that made the file no longer verbatim. The rule that avoids both is the one now applied — **a licence file contains the licence and nothing else.**

## The module header still described S4 as future work (@coderabbitai, missed last round)

You flagged this at `TextBake.cpp:770` in the previous review and I didn't act on it — my oversight. Both blocks said the first artifact would be committed by S4 "once the parser and rasteriser are available", and that a recipe with no `[runs]` table was the only shape `run()` produces. This PR *is* S4.

Rewritten to describe the two dispatch paths, and `run()`'s doc now states what it dispatches on, what `root` confines, and which font-specific refusals return nullopt.

## The evidence check earned its keep

Worth reporting, since it's the mechanism this whole PR exists to establish. Editing the recipe's provenance comment changed its digest, which `report.json` records — so `evidence.font.dejavu-ui` **failed** on a one-byte difference at `report.json` byte 852 until I re-baked:

```
Evidence mismatch for font/dejavu-ui: the baked artifact is not byte-identical
  report.json: first differs at byte 852 - baked 0x66, committed 0x39
```

Re-baked, and **only the recipe digest moved** — `package.json` and `atlas.bin` are unchanged, which is the check that a comment edit did not reach the bake.

## Declining one, again

The `@brief` nitpick on `tests/text/TextBakeTests.cpp:559` is on a file-local `constexpr std::string_view` inside an anonymous namespace. I declined the same finding on #172 with the repo-wide count; the local one is just as lopsided — **this file has 13 `///` comments and 0 with `@brief`**, several predating my changes. `CONTRIBUTING.md`'s Doxygen section scopes the requirement to file-level, class and method documentation, and adds "Private members: generally no documentation needed unless complex."

Happy to reverse if the intent is that every `///` carries `@brief`, but that's a repo-wide sweep rather than one line here.

423/423 ctest, 380 evidence tests, docs-lint clean.

PR #177, opened 2026-08-08
S5 — mdux.font.schema and restricted-charset validation (#161)

ambroise-leclerc:
Part of #14. Depends on S4 (#160), which landed as #173 → #174 → #176. Single PR this time — the schema and its adoption are one change, not a stack.

## One definition, not two

S4 emitted the font package's JSON **by hand**, because the baker landed before this module existed. Adopting the governed type does not change what a package *is* — it removes the second, informal definition.

That is ADR-008 decision 1 mirrored to text by ADR-010, and the same reason `mdux.text.schema` exists for text packages: a baker producing one shape and a runtime expecting another is the Wave 2 lesson. `write()` validates on the way out and `parse()` validates on the way in, so **there is no state in which a caller holds a `FontPackage` that has not been checked**.

The atlas is byte-identical after the change — the sheet did not move, the package gained its declared fields.

## The restricted charset

The declared set of code points a package can draw, carried as **ranges** rather than a code-point list: `U+0020..U+007E` is checkable by eye in a way ninety-five integers is not, and membership becomes a binary search.

`permits()` is the test the `.medui` compiler (#15) will apply to every code point a dynamic-text format could produce — the thing that turns "no shaping on device" from a slogan into a compile error. `validate()` refuses a package whose charset names a code point it has **no glyph for**, so the table cannot promise more than the atlas delivers. That matters because the runtime has no fallback, and having one would mean mapping code points on device.

## Tabular figures — a safety rule, not a typographic preference

`validate()` requires every decimal digit present to share one advance width.

A numeric field on a medical device — a heart rate, a dose, an elapsed time — is redrawn as its value changes. If `1` is narrower than `8`, the digits shift horizontally on every update and the field **jitters**. A clinician reading a changing number at a glance is precisely the reader that motion hurts most, and the defect is *invisible in a static screenshot* — so it has to be caught where the font is baked rather than where it is looked at.

A font with proportional digits is not broken; it is unsuitable for this use, and the baker says so with `TXT019` — its own code, because the fix is a different font rather than a different recipe.

I checked this against the committed font before designing around it: **DejaVu Sans' ten digits all advance 1303 font units**, so the rule holds on the artifact this repository already ships.

## Tests — `font_spec`

Its own executable rather than cases inside `text_spec`: a font package and a text package are different shapes with different rules, and a failure should name which schema rejected the input. It links `MduX::Core` only, which is standing evidence that a font package describes an atlas and a charset without naming a Vulkan handle.

- **The round trip is the scenario that matters.** Write, parse, compare field by field — including signed bearings and a negative kerning adjustment — then re-write and compare **bytes**. A writer and a reader can only agree wrongly if they share code, which is exactly the property this module is for. Asserting that each side independently produces what a test author expected would prove much less.
- **Eighteen rejection rows**, one per rule, each breaking a valid fixture in exactly one way. The fixture's own validity is asserted first, or every row would pass for the wrong reason.
- **The committed `dejavu-ui` package parses**, permits both ends of its charset, refuses an accented letter outside it, and has a space that is blank but still advances. A schema that only ever validated its own fixtures would not catch a baker writing a field under a different name.

## Against #161's acceptance criteria

| | |
|---|---|
| a dynamic-text format referencing an unbaked code point fails the baker with a stable code | the table and `permits()` are here; the **compiler-side** rejection is #15's, which this unblocks |
| tabular-figure mismatches fail the baker | yes — `TXT019`, with `SchemaError::TabularFigureMismatch` underneath |
| `font_spec` schema tests pass on MSVC and GCC | green locally on GCC; this PR's CI is the first MSVC run |
| the governed-zone check still passes | yes — "2 governed target(s) clean of Vulkan/windowing dependencies" |

**427/427 ctest**, 384 evidence tests, docs-lint and schema-drift clean.

## Out of scope, deliberately

`kerning` is emitted **empty rather than omitted**, so a consumer reads "this package bakes no kerning" instead of having to distinguish an absent member from an empty one. Populating it is future work — and ADR-010's rule is that whatever is not in that list does not apply, because no runtime lookup can find it.

## Summary by CodeRabbit

* **New Features**
  * Added a standardized font package schema covering metadata, glyphs, character ranges, atlas metrics, and kerning.
  * Added validation, parsing, serialization, glyph lookup, charset checks, and kerning lookup.
  * Updated text baking to generate validated font packages.

* **Bug Fixes**
  * Improved diagnostics for invalid tabular-figure data and other font schema errors.

* **Documentation**
  * Updated the architecture roadmap to record the completed font schema milestone.

* **Tests**
  * Added coverage for serialization round trips, validation, charset boundaries, metrics, and the bundled font package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,7 @@ target_sources(MduXCore
             include/mdux/ml/Kernels.cppm
             include/mdux/ml/Runtime.cppm
             include/mdux/draw/Draw.cppm
+            include/mdux/font/Schema.cppm
             include/mdux/text/Schema.cppm
             include/mdux/text/Raster.cppm
     PRIVATE
@@ -204,6 +205,7 @@ target_sources(MduXCore
         src/ml/Kernels.cpp
         src/ml/Runtime.cpp
         src/draw/Draw.cpp
+        src/font/Schema.cpp
         src/text/Schema.cpp
         src/text/Raster.cpp
         src/governance/Governance.cpp
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -54,6 +54,7 @@ are ordinary `PRIVATE` sources.
 | `mdux.shader.schema` | `include/mdux/shader/Schema.cppm` | `src/shader/Schema.cpp` |
 | `mdux.text.schema` | `include/mdux/text/Schema.cppm` | `src/text/Schema.cpp` |
 | `mdux.text.raster` | `include/mdux/text/Raster.cppm` | `src/text/Raster.cpp` |
+| `mdux.font.schema` | `include/mdux/font/Schema.cppm` | `src/font/Schema.cpp` |
 | `mdux.draw` | `include/mdux/draw/Draw.cppm` | `src/draw/Draw.cpp` |
 | `mdux.ml.schema` | `include/mdux/ml/Schema.cppm` | header-only |
 | `mdux.ml.kernels` | `include/mdux/ml/Kernels.cppm` | `src/ml/Kernels.cpp` |
@@ -210,7 +211,7 @@ tracking issue; the issue is authoritative for what remains.
 
 | Planned | Issue | Note |
 |---|---|---|
-| Font and text pipeline | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) | S1 (`mdux.text.schema` + `mdux-textbake` skeleton) landed. S2 (`mdux.tools.truetype`, #158) landed. S3 (`mdux.text.raster`, #159) landed. S4 (atlas packer + font baker, #160) landed with the first committed font package; S5–S6 open |
+| Font and text pipeline | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) | S1–S4 landed (`mdux.text.schema`, `mdux.tools.truetype`, `mdux.text.raster`, atlas packer + font baker with the first committed font package). S5 (`mdux.font.schema` + restricted charset, #161) landed; S6 open |
 | `.medui` compiler | [#15](https://github.com/ambroise-leclerc/MduX/issues/15) | the replacement for the retired HTML/CSS path |
 | Rendered-truth verification | [#16](https://github.com/ambroise-leclerc/MduX/issues/16) | beyond the current pixel test |
 | Content components | [#17](https://github.com/ambroise-leclerc/MduX/issues/17) | `SignalTrace`, `StatusIndicator`, `NumericDisplay` and the rest |
--- a/generated/font/dejavu-ui/package.json
+++ b/generated/font/dejavu-ui/package.json
@@ -1150,11 +1150,18 @@
     }
   ],
   "id": "dejavu-ui",
+  "kerning": [],
   "kind": "font",
   "locales": [
     "en-US"
   ],
   "pixelSize": 16,
+  "restrictedCharset": [
+    {
+      "first": 32,
+      "last": 126
+    }
+  ],
   "schemaVersion": 1,
   "unitsPerEm": 2048
 }
--- a/generated/font/dejavu-ui/report.json
+++ b/generated/font/dejavu-ui/report.json
@@ -26,7 +26,7 @@
   "outputs": [
     {
       "path": "package.json",
-      "sha256": "61c719a6572d6474148943f57f9588312f08f3302fb5d9e5ed6409c4f6506051"
+      "sha256": "34bff16611756b163fc7fbedcfa4935d614d23e4b35c705ccab75f1814f28873"
     },
     {
       "path": "atlas.bin",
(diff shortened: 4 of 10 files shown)


camille-martin-paris (changes requested):
Blocking findings:

1. **P1 — Numeric fields are narrowed before range validation.** In `src/font/Schema.cpp` around lines 370–484, parsed `int64_t` values are directly cast to smaller signed/unsigned destination types. For example, `unitsPerEm: 65537` becomes `1`, while negative values assigned to unsigned fields wrap. A malformed package can therefore validate as different data than the JSON contains. Please range-check every parsed integer before conversion, including enforcing the Unicode scalar ceiling of `U+10FFFF`.

2. **P1 — The declared atlas digest is never validated.** `FontPackage::validate()` accepts an empty or malformed `atlas.sha256`, despite the public contract requiring exactly 64 lowercase hexadecimal characters. This allows a package to pass validation without a usable sidecar integrity value. Please validate the digest and cover empty, incorrectly sized, uppercase, and non-hex inputs.

3. **P2 — `atlas.path` still permits special and drive-relative paths.** The current check only rejects slash characters, so `"."`, `".."`, and Windows drive-relative forms such as `"C:atlas.bin"` are accepted as bare filenames. Please validate that the value is exactly one ordinary filename component on supported platforms.

Secondary correctness issue: wrong-type arrays/objects and grouped integer failures around lines 381–481 are frequently collapsed to `MissingMember` rather than preserving `WrongType`, which makes malformed input diagnostics misleading.

CI is green, but the first two findings directly undermine the guarantee that successfully parsed packages have been fully validated.

ambroise-leclerc:
All ten addressed in `7c8e91f`. Two of them undermined the guarantee this module exists to make.

## Every parsed integer was narrowed before it was checked (@camille-martin-paris P1, @coderabbitai, @Copilot)

Correct, and worth naming plainly: **this is the same defect Copilot found on `pixelSize` in #176, repeated across every field in `parse()`.** I fixed the instance there and not the pattern.

`unitsPerEm: 65537` became 1, `pixelSize: -1` became 4294967295, and both then satisfied `validate()` — the package parsed successfully as *different data than the JSON contained*, which is precisely what "it parsed, so it was validated" is supposed to rule out.

Fixed with one `requireIntIn<T>()` that range-checks before narrowing, applied uniformly, rather than eleven individual guards that would drift apart. Code points go through `requireCodePoint()`, which also enforces the **U+10FFFF ceiling** you both asked for — `char32_t` admits four billion values of which 0x110000 are characters.

## The atlas digest was never validated (@camille-martin-paris P1, @coderabbitai)

Right. The field's whole purpose is letting a runtime check the sidecar it was handed belongs to the package, and `write()` could emit one with no usable value. Now exactly 64 lowercase hex, with rows for empty, wrong length, uppercase and non-hex — uppercase matters because two spellings of one digest compare unequal as strings.

## The path check was too narrow (@camille-martin-paris P2)

Also right: banning `/` and `\` let `"."`, `".."` and `"C:atlas.bin"` through, each resolving somewhere other than beside `package.json` on at least one supported platform. Checked positively now.

## `validate()` could hang (@Copilot)

Genuine, and the nastiest of the set. The charset walk incremented a `char32_t` under `point <= range.last`, so a range ending at the type's maximum wrapped to 0 and looped forever. The scalar ceiling makes it unreachable, and the loop now terminates regardless — the end test is inside the body, so the increment can never step past it.

## `WrongType` collapsed into `MissingMember` (@camille-martin-paris, @coderabbitai)

`requireArray()` returned nullptr for both absent and wrong-type, so `"glyphs": {}` reported a missing member that is right there. Returns a `Result` now, and the atlas' four integers are reported individually rather than as one grouped failure.

## Tests

`font-schema-integer-range` is the scenario for the first finding — four documents, each with one field replaced by an out-of-range value, all refused rather than narrowed. Plus eight more `validate()` rows (three path shapes, four digest shapes, a code point above U+10FFFF), `occupancyPercent` added to the round-trip comparison, and the committed-package scenario now checks **every** printable-ASCII point rather than the two endpoints — a package missing an interior one could otherwise have 95 glyphs and pass.

**The committed artifact is byte-identical throughout**: this hardened the reader, not the writer.

428/428 ctest, 385 evidence tests, docs-lint clean.

---

## On `@brief` — third time, so let me settle it rather than decline again

@coderabbitai is right that `CONTRIBUTING.md` requires `@brief` for *class documentation*, and four of my `///` comments are on structs. But the repo does both: `TextRun` in `mdux.text.schema` and `FileRecord` in `mdux.evidence.report` are one-line `///` with no `@brief`, while longer struct docs use `/** @brief … */`. Mine follow that split — short one-liners `///`, the long `FontPackage` block `/** @brief */`.

Repo-wide the count is **391 `///` comments without `@brief`, 0 with**.

@ambroise-leclerc — this is a convention call rather than a per-PR one, and it has now come up on three consecutive PRs. Either reading is defensible; I would rather apply yours consistently than keep declining. If you want `@brief` on every `///`, say so and I will do the sweep as its own commit.

PR #178, opened 2026-08-08
feat(text): coverage draw path and text pixel tests (#162)

### LINKED ISSUE #162: S6 — Coverage draw path and text pixel tests

ambroise-leclerc:
Closes #162. Last child of epic #14 — with this merged, the font and text pipeline is complete: a `.ttf` in `recipes/` becomes a committed package, and a baked run from that package reaches a frame.

## What this adds

**`mdux.text.draw`** (governed, `MduXCore`) — decodes the six-byte v1 run record and records each glyph as a `CoverageR8` rectangle.

It is replay, not layout. A record already carries the glyph and the position the baker computed; this module looks the glyph up in the package's table and emits a rectangle where it was told to. There is no pen advancing by a computed width, no code-point-to-glyph mapping, no fallback. The test for ADR-010 is whether a different input could produce a different *layout* — it cannot, because positions are inputs rather than outputs. Worth stating explicitly, because "draws text" sounds like the thing the ADR forbids.

**`UiRenderer::createWithCoverageAtlas()`** (adapter) — uploads a baked sheet in place of the built-in 2×2 default. The renderer previously had no way to replace its atlas contents at all, which is why this is here rather than in a later issue.

**`tests/render/TextPixelTests.cpp`** — four scenarios under lavapipe, `-L pixel`.

## The finding that made this module necessary

`DrawList::addRect()` takes its `uv` as a `core::Rect`, which is **integer**, and copies those integers straight into the vertex's `float u`/`v`. The fragment shader samples with `texture(uAtlas, fragUv)` — a `sampler2D`, so **normalised** coordinates.

An integer rectangle can therefore only ever express 0 and 1. A real atlas slot such as (13, 46, 11×12) would arrive as `u = 13.0` and sample far outside the sheet. Every glyph would have been wrong, and silently — a sampler clamps rather than fails.

So `mdux.draw` gains a `UvRect` (four floats) and an `addRect` overload taking it; the integer overload now delegates by widening, so nothing that exists changes behaviour. Normalising a texel slot against the sheet extent is exactly what `addGlyphRect()` is for, and it is the only supported way to record a `CoverageR8` rectangle.

## Why the pixel tests have no goldens

The atlas is synthetic: an 8×8 R8 sheet with a 4×4 block at coverage 255 beside a 4×4 block at coverage 128. That makes every expected pixel **derivable** from the blend equation — `255 * coverage/255` — rather than recorded from a run that happened to look right. A failing test names a wrong number instead of a changed image, and nobody is ever tempted to re-bless a golden.

Three of the four scenarios are negative, because a pixel test that only asserts the good case proves very little:

| scenario | what it would catch |
|---|---|
| a run renders at its recorded bounds with its recorded coverage | the happy path, including that 128 blends to 128 and not 255 |
| a run moved by one pixel no longer matches | a comparison too loose to see placement |
| a run drawn from the wrong atlas slot no longer matches | a uv error that lands *inside* the sheet |
| a blank glyph occupies no pixels but does not fail the run | the space — must draw nothing without being an error |

## Governed unit tests

`tests/text/DrawTests.cpp` covers the three things a rendered frame can only show indirectly, where a failure would otherwise appear as "a glyph is in the wrong place":

- **little-endian decode**, including a negative `x`. The byte order is a contract, not an implementation detail: the sidecar is committed bytes compared across toolchains, so a decode inheriting the host's order would move glyphs on one CI leg and not the other. A five-byte record is refused.
- **the baseline inversion** — `bitmapOriginY` is measured *up* from the baseline while the draw path's y axis points down. Getting this backwards renders text that looks correct in isolation and sits at the wrong height beside anything else.
- **the uv division** — slot (4,0) 4×4 on an 8×8 sheet gives u 0.5→1.0, v 0.0→0.5. Checkable by eye.

Plus the two refusals: a partial run records *nothing* before failing (a run one byte short would otherwise draw every glyph but the last and look merely truncated), and a glyph index past the package's table is refused rather than drawing a plausible wrong character. The index is into the *package's* glyph table, not the font's `glyf` table — those differ, since the package holds only the baked charset.

## Verification

- `434/434` tests pass — 33 → 36 in `text_spec`, 26 → 30 in `offscreen_tests`
- `mdux_verify_trust_zones: 2 governed target(s) clean of Vulkan/windowing dependencies` — `mdux.text.draw` is in `MduXCore` and imports `std` only
- labels intact: `evidence` 5, `determinism` 4, `pixel` 1, `noheap` 5, `regulatory` 4
- docs: `mdux.text.draw` recorded in `architecture.md`; `#14` marked closed in `roadmap.md` and the test count refreshed

## Summary by CodeRabbit

- **New Features**
  - Added text rendering from baked glyph runs, including placement, coverage drawing, normalized texture coordinates, and validation.
  - Added coverage-based font atlas support and fractional UV rectangles.
  - Added draw-list rollback support for recovering from failed text rendering.

- **Bug Fixes**
  - Improved validation for malformed records, invalid glyphs, empty atlases, and mismatched atlas data.

- **Tests**
  - Added unit and pixel-rendering coverage for text rendering scenarios.

- **Documentation**
  - Updated architecture, roadmap, and contribution guidance.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,7 +1,9 @@
 name: "MduX CodeQL config"
 
-# ninja-gcc (binaryDir build-gcc) sets MDUX_BUILD_TESTS/MDUX_BUILD_EXAMPLES=ON, so CPM.cmake
-# fetches Catch2 and glfw into build-gcc/_deps/ and CodeQL would otherwise extract and alert on
-# that third-party source too - longer runs, noise that is not this repository's code to fix.
+# This job extracts without building (see codeql.yml), so there is no build-gcc/_deps/ tree of
+# fetched Catch2 and glfw sources to exclude - CPM.cmake never runs. The entry is kept, scoped to
+# any build directory, so a future traced build cannot silently start reporting alerts against
+# third-party source that is not this repository's to fix.
 paths-ignore:
-  - build-gcc/_deps/**
+  - '**/_deps/**'
+  - 'build*/**'
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -82,6 +82,13 @@ jobs:
         ln -sf /opt/cmake-${CMAKE_VERSION}-linux-x86_64/bin/ctest /usr/bin/ctest
         cmake --version
 
+    # The container runs as root while the checkout is owned by the runner user, so every git
+    # command the CodeQL action issues fails with "detected dubious ownership" and it falls back to
+    # the commit SHA from the environment. On a pull_request event that is the merge commit rather
+    # than the head, so alerts would be attributed to a commit that exists in no branch.
+    - name: Trust the workspace checkout
+      run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
     - name: Initialize CodeQL
       uses: github/codeql-action/init@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
       with:
@@ -96,8 +103,31 @@ jobs:
     # options change.
     - name: Configure and build
       run: |
+        # CodeQL sets LD_PRELOAD to the `${LIB}_${PLATFORM}_trace.so` form, whose ld.so tokens this
+        # container's dynamic linker does not expand: it reports "cannot be preloaded ... ignored"
+        # and then every compiler invocation runs untraced. The job still passed, because an empty
+        # database analyses cleanly - 397 compiled translation units were producing a 7.76 KiB
+        # source archive and 20 KiB of relations. Pointing LD_PRELOAD at the concrete 64-bit tracer
+        # the same environment already exports is what makes the extraction real.
+        export LD_PRELOAD="${SEMMLE_PRELOAD_libtrace64:-$LD_PRELOAD}"
         cmake --preset ninja-gcc
         cmake --build --preset ninja-gcc
+
+    # A CodeQL run that extracted nothing reports success, which is the worst of both worlds: a
+    # green check and no analysis. Nothing upstream fails on it, so the check has to be here, and
+    # it has to be a count rather than a warning scan - the tracer's own failure was already
+    # printed as an ERROR line in a passing job for months.
+    - name: Verify the tracer actually saw the code
+      run: |
+        archive="${RUNNER_TEMP}/codeql_databases/cpp/src"
+        extracted=$(find "$archive" \( -name '*.cpp' -o -name '*.cppm' -o -name '*.hpp' \) 2>/dev/null | wc -l)
+        echo "CodeQL extracted ${extracted} C++ source files"
+        # The tree holds over a thousand; a working extraction reaches the hundreds, and a broken
+        # one lands in single digits. 200 separates those without tracking the exact figure.
+        if [ "${extracted}" -lt 200 ]; then
+          echo "::error::CodeQL extracted only ${extracted} source files - the build tracer is not attached, so the analysis below would be vacuous"
+          exit 1
+        fi
 
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,7 +20,6 @@ jobs:
   analyze:
     name: Analyze (C++)
     runs-on: ubuntu-latest
-    container: gcc:16
     permissions:
       security-events: write
       actions: read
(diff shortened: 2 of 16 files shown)


ambroise-leclerc:
Addressed. Nine of eleven fixed, one declined with evidence, one already correct.

## The two substantive ones

**`recordRun()` left rectangles behind on failure** (CodeRabbit, Copilot). Correct, and my header was making a claim the code did not honour: the length check catches a truncated sidecar before anything is recorded, but `GlyphIndexOutOfRange`, `EmptyAtlas` and `ListRejected` can all fire on the fourth record of four. "A partially decoded run does not reach a frame" was true only if every caller discarded the list.

It is now true of the list itself. `DrawList` gains an opaque `Marker` with `mark()`/`rollback()`, and `recordRun()` restores the list to entry state on any failure — leaving untouched whatever the caller recorded *before* the run.

I did not take the suggested scratch-list-and-merge: `DrawList` is a view over caller-provided storage, so a scratch copy needs its own buffers, which a governed no-heap module cannot conjure. The marker is four integers and a clip rect, no allocation. One subtlety worth naming — restoring `commandCount` alone is not enough, because `addRect()` extends the last command's `indexCount` in place when the clip has not changed, so a rolled-back list would keep a command claiming indices that were no longer there. The marker carries that fourth number.

Verified rather than asserted: with `rollback()` temporarily neutered, the new test fails on two expectations. Re-enabled, 37/37.

**The `SampledRgba` footgun** (Copilot). Agreed it is real, and I went further than the docs fix suggested. An R8 image sampled as RGBA returns `(coverage, 0, 0, 1)` — a plausible picture in the wrong colours, which is exactly the class of failure that survives review. `record()` now refuses any frame containing a `SampledRgba` vertex when the atlas is R8, with `RenderError::SampledRgbaWithCoverageAtlas`. The scan is per-frame because the mode is per-vertex and the list is rebuilt each frame; it is a pass over the same vertices the upload already memcpys.

Writing the test for it found a bug in that very change: `atlasIsCoverageOnly_` was missing from the hand-written move assignment, and since `create()` returns by value the flag was set on a renderer that was immediately moved from — so the guard silently never fired. The comment already sitting in that function warns about precisely this ("create() returns by value, so *every* renderer is a moved one — this list is not an edge case"). Caught by the test, not by review, which is the argument for the test.

## The rest

| Finding | Done |
|---|---|
| `decodeRecord()` accepts oversize spans (Copilot) | Now requires exactly one record — passing the whole run returned the first glyph and looked like it worked. New `RecordSizeWrong` code and a test for the two-record case |
| `GlyphPlacement::glyphIndex` collides with `GlyphRecord::glyphIndex` (Copilot) | Renamed to `packageIndex`. They are genuinely different numbers for the same glyph, and one name for both is how you draw the wrong character |
| `@param origin` names no parameter (CodeRabbit) | Split into `originX`/`originY` |
| `AtlasExtentMismatch` text mentions only the byte count (CodeRabbit) | Now names the zero-extent case too. The enumerator's own doc comment already did — it was only `describe()` that was short |
| Duplicated v1 record encoder (CodeRabbit, nitpick) | Agreed — the layout is a format contract, and two encoders could drift into agreeing with a decoder that was also wrong. Now `tests/framework/RunRecords.hpp`, shared by both targets |
| `renderRun()` collapses five failures to `nullopt` (CodeRabbit) | Now carries which step failed and why. Also fixed a latent trap in doing so: `REQUIRE_MESSAGE` takes its message by value, so calling `.error()` at the assertion would read the error of an `expected` holding pixels — on the *passing* path |
| Docs contradict themselves (CodeRabbit) | Fair, and mine: marking #14 closed in two places left a "Complete" row inside architecture.md's **Planned, not built** table, and a roadmap still opening Wave 4 with #14 while blocking #15 on it. Wave state, the epic footer, `#15`'s heading and the Tests row are consistent again |

## Declined

**`@brief` on `///` comments.** This is the fourth time across four PRs, so rather than decline again I measured it and wrote the answer down. The tree has **485** `///` comments and **zero** carry `@brief`; `/** */` blocks do, 149 of them. My new code follows both halves of that exactly. A `///` comment *is* the brief — Doxygen already treats it as one — so the tag adds a word meaning "this is a comment".

The rule is now in `CONTRIBUTING.md` next to the `k`-prefix rule, so the question has an answer that outlives this thread.

One thing I noticed but did not change, because it is a convention call rather than a review item: `CONTRIBUTING.md` says file-level docs should use `@brief` only, **no `@file`** — but 82 files use `@file` and 54 do not, including every file in this PR. Worth settling separately; I did not want to silently rewrite a written rule to match what the tree does.

## Verification

- **435/435** pass (was 434) — `text_spec` 36 → 37, `offscreen_tests` 30 → 31
- `mdux_verify_trust_zones: 2 governed target(s) clean` — `DrawList::Marker` is std-only
- labels unchanged: `evidence` 5, `determinism` 4, `pixel` 1, `noheap` 5, `regulatory` 4

ambroise-leclerc:
## CodeQL was passing green on an empty database

Those log lines are not cosmetic. The `LD_PRELOAD ... trace.so cannot be preloaded: ignored` line is CodeQL's build tracer failing to attach, so no compiler invocation was ever observed — and the job passes anyway, because an empty database analyses cleanly.

The numbers from the last run before this branch touched it:

```text
397 translation units compiled
Finished writing database (relations: 20.30 KiB; string pool: 2.06 MiB)
Finished zipping source archive (7.76 KiB)
```

7.76 KiB of source archive for 397 TUs of C++23. Every CodeQL run on this repository has been reporting success while examining **two files**.

### What I tried, in order

**1. Fix the tracer.** CodeQL sets `LD_PRELOAD` to the `${LIB}_${PLATFORM}_trace.so` form, whose ld.so tokens the `gcc:16` container's linker does not expand; the same environment already exports the concrete `lib64trace.so`, so I pointed `LD_PRELOAD` at that. The error message went away, the tracer loaded cleanly — and the database still held two files. So the tracer was not the thing to keep fixing.

**2. A guard, first.** Before changing the approach I added a step that counts extracted sources and fails, because nothing upstream fails on a vacuous run — that is precisely how this went unnoticed. It is what made the problem legible:

```text
CodeQL extracted 2 C++ source files
##[error]CodeQL extracted only 2 source files - the build tracer is not attached,
         so the analysis below would be vacuous
```

First honest failure this workflow has produced.

**3. `build-mode: none`.** Buildless extraction reads the sources directly and needs no tracer. With no build to trace, the `gcc:16` container, the pinned CMake download and the whole dependency install go with it — the job is now checkout, extract, analyse.

```text
                       before        after
source archive       7.76 KiB    506.45 KiB
relations           20.30 KiB      1.29 MiB
files extracted             2           153
```

153 against 139 tracked C++ files, so that is the whole tree.

**4. Then I got the guard's own threshold wrong**, and it failed a working run. I had written the floor as a literal 200, from a file count that included untracked build trees. It now counts `git ls-files` at run time and requires 80%, so there is no number for me to have miscalibrated and none to rot as the tree grows.

**Result: the first real CodeQL analysis of this repository is green with zero C++ alerts.** The eight open code-scanning alerts are all Scorecard's, from `security-analysis.yml`, and predate this.

### The cost, stated plainly

Buildless extraction resolves fewer includes than a working traced build would, so this is a real reduction against *a traced build that worked* — but that is not the alternative on offer; the alternative was two files. The compile is not lost either: `linux-gcc16-build.yml` and `windows-build.yml` both build this tree on every PR, so nothing in this job was the only thing proving it compiles.

Also fixed: the container ran as root against a runner-owned checkout, so every git command the action issued failed with `dubious ownership` and it fell back to the environment's commit SHA — the merge commit on a `pull_request` event, which exists in no branch, so alerts were being attributed to a commit in no branch. Dropping the container removes it.

---

## The four new review comments

**`DrawList::Marker` could be forged or borrowed** (CodeRabbit, Major) — correct, and it was my doc comment that was wrong: I wrote "the counters only ever move backwards here, so the worst case is a list shorter than the caller expected", which holds only for a marker from the same list. `rollback()` writes through the command count, so a marker naming more commands than the storage holds is an out-of-bounds write, not a wrong picture.

The fields are now private with the owning list carried alongside, and `rollback()` returns `DrawError::WrongList` for a foreign, default-constructed or stale marker. The owner is compared, never dereferenced, so a marker outliving its list is a mismatch rather than a dangling read. New test covers all three cases, including that a rejected marker leaves the list unmoved.

**Both `renderAndRead()` results discarded** — right, a readback failure would have passed that test for the wrong reason. Both are now required to succeed.

**Fence language on the wave diagram** — tagged `text`.

**`@brief` on `RunRecords.hpp`** — declining, consistent with the rule added to `CONTRIBUTING.md` in the previous commit: 485 `///` comments in the tree, none with `@brief`; `/** */` blocks do carry it, and that file's own file-level block does.

436 tests pass.

PR #181, opened 2026-08-09
ci: run ASan and UBSan over the test suite (#179)

### LINKED ISSUE #179: Run the sanitizers in CI
## Why

ASan found a real use-after-free in this repository during #171 — every `buildFont` lambda in `TruetypeTests.cpp` parsed from a `serialize()` temporary, and `tt::Font` holds non-owning spans. The ordinary build was green. It took an explicitly instrumented run to see it, that run was manual, and nothing repeats it.

`cmake/Sanitizers.cmake` already exists and already exposes `ENABLE_SANITIZER_ADDRESS`, `_LEAK`, `_UNDEFINED_BEHAVIOR`, `_THREAD` and `_MEMORY`, wired through `enable_sanitizers(MduX_options)`. Nothing in `.github/workflows/` sets any of them, so the machinery has never run in CI.

This matters more here than in most projects. The tree is full of hand-rolled binary parsing — `mdux.tools.truetype` walks `glyf`/`loca`/`cmap`/`hmtx` over non-owning spans, `mdux.text.raster` indexes edge buckets by arithmetic, `mdux.text.draw` decodes byte records — and span lifetime bugs of exactly the #171 shape are the failure mode that class of code has. A governed target that must not allocate is also one where a stray write has nowhere to be caught.

## Scope

- A workflow running the test suite under ASan + UBSan (they compose; TSan and MSan do not compose with them and are out of scope here).
- `-fno-sanitize-recover=all`, so UBSan fails the job rather than printing and continuing. A sanitizer that only warns is a sanitizer nobody reads.
- Exclude what cannot be usefully instrumented: `offscreen_tests` runs against lavapipe, whose driver allocations are not our code. Decide whether to run it with suppressions or skip it, and say which in the workflow.
- Not on every PR if it is slow — a schedule plus `workflow_dispatch` is enough to catch regressions, and the point is that it runs at all.

## Done when

- A green sanitizer run exists on `develop`.
- A deliberately reintroduced use-after-free of the #171 shape fails it. Demonstrated, not assumed — the whole reason this issue exists is that a green build proved nothing.

ambroise-leclerc:
Closes #179.

`cmake/Sanitizers.cmake` has exposed `ENABLE_SANITIZER_ADDRESS` and friends for a long time, and nothing in `.github/workflows/` has ever set one. The machinery was there; nobody had pulled the lever.

## It found a use-after-free on the first run

Not a hypothetical. `tests/shader/SchemaTests.cpp` built a `ShaderPackage` as a local in its `Given` step and kept the pointers `find()` returned into `package.modules`:

```cpp
.Given("the reference package", [state] {
    const ShaderPackage package = validPackage();   // dies at the closing brace
    state->found = package.find("ui.vert");         // pointer into package.modules
    state->miss  = package.find("ui.frag");
})
```

The `Then` step then dereferences `state->found->stage`. The suite had been green throughout, because the freed bytes still happened to hold the right stage value.

```text
ERROR: AddressSanitizer: heap-use-after-free on address 0x76b5b03e0060
READ of size 1 at 0x76b5b03e0060 thread T0
    #0 tests/shader/SchemaTests.cpp:940
freed by thread T0 here:
    #7 mdux::shader::ShaderPackage::~ShaderPackage() include/mdux/shader/Schema.cppm:170
    #8 tests/shader/SchemaTests.cpp:930
```

Same shape as the #171 finding — a view outliving what it views — in a different suite, found the same way. The package is now owned by the shared state.

That is also the acceptance criterion #179 asked for ("a deliberately reintroduced use-after-free fails it"), satisfied by a real one rather than a planted one.

## Two build-level changes were needed

**`-Wframe-larger-than=4096` is suspended under ASan.** The sanitizer wraps every stack object in redzones, which inflated `src/font/Schema.cpp` to 7648 bytes and `src/governance/Compliance.cpp` to 5040 — for reasons that have nothing to do with the source. Left on, the sanitizer build simply cannot compile. Raising the number for everyone would have retired a real guard to accommodate a build nobody ships, so it is scoped with a generator expression instead; ordinary builds keep it, and that is where a genuine stack regression would appear. Verified: 62 occurrences of the flag remain in the ordinary `build.ninja`.

**`-fno-sanitize-recover=all`.** UBSan's default is to print and continue. In CI that is a job that reports a violation and then goes green, which is a finding nobody reads.

## The one exclusion, and why it is not "skip the hard suite"

`offscreen_tests` runs against lavapipe. The Vulkan ICD leaks 512 bytes once, on a thread it starts itself, via `pthread_once`:

```text
Direct leak of 512 byte(s) in 1 object(s) allocated from:
    #0 realloc
    #1 <unknown module>
    #3 __pthread_once_slow
    #7 asan_thread_start
```

No MduX frame anywhere in it, and the frames resolve to `<unknown module>`, so it cannot be suppressed by name either. Leak detection is therefore off **for that suite only**. Use-after-free, buffer overflow and use-after-scope stay on — those are precisely the checks that matter for code rendering through spans into mapped memory, and dropping the suite entirely would have been the worse trade.

## Verification

| | |
|---|---|
| under ASan + UBSan, leaks on | 435/435 |
| `offscreen_tests` under ASan, leaks off | 31/31 |
| ordinary build, unchanged | 436/436 |

Runs on PRs to the usual branch set, on push to `main`/`develop`, weekly, and on demand. On PRs because that is the point at which this bug class is cheapest to fix — both instances found so far were introduced by a passing PR.

## Summary by CodeRabbit

- **Tests**
  - Added automated sanitizer coverage for memory errors and undefined behavior.
  - Added scheduled and on-demand validation for general functionality and Vulkan rendering.
  - Rendering checks now fail when the graphics path is not exercised.
  - Improved shader package test stability and pointer-lifetime validation.
  - Corrected test reporting so rendering failures cannot be masked.

- **Bug Fixes**
  - Sanitizer-detected issues now stop validation immediately.
  - Preserved frame-size checks for standard builds while supporting sanitizer-enabled validation.

### DIFF
--- a/.github/workflows/linux-gcc16-build.yml
+++ b/.github/workflows/linux-gcc16-build.yml
@@ -124,6 +124,10 @@ jobs:
         # covered by .gitignore's `*.log` - so the "Verify no source-tree writes" step below could
         # not see it, and, more to the point, could not see any other .log a build step wrote
         # either. Keeping this file outside the tree keeps that check honest.
+        # Same reason as the sanitizer leg: `run:` gets `bash -e`, not `bash -eo pipefail`, so the
+        # `| tee` below would otherwise report tee's status and let a failing pixel run pass. This
+        # predates #179 and was found while reviewing the copy of this step made for it.
+        set -o pipefail
         PIXEL_LOG="$RUNNER_TEMP/pixel.log"
         # -V rather than --output-on-failure: a passing run must still print, because two of the
         # things this step verifies are only visible in the output of tests that succeeded.
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -111,7 +111,9 @@ jobs:
       run: |
         export DISPLAY=:99
         Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-        ctest --test-dir build-asan -E offscreen_tests --output-on-failure
+        # --no-tests=error: a filter that matched nothing would otherwise be a pass. The whole
+        # point of this job is that it ran.
+        ctest --test-dir build-asan -E offscreen_tests --output-on-failure --no-tests=error
 
     # offscreen_tests runs against lavapipe, and the Vulkan ICD leaks 512 bytes once, on a thread
     # it starts itself, through pthread_once - a stack with no MduX frame in it at all. That is the
@@ -129,4 +131,15 @@ jobs:
       run: |
         export DISPLAY=:99
         Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-        ctest --test-dir build-asan -R offscreen_tests --output-on-failure
+        # offscreen_tests exits 77 with no loader, no ICD or no device, and tests/CMakeLists.txt
+        # sets SKIP_RETURN_CODE 77 so CTest reports that as Skipped - which does not fail a run.
+        # Without the check below, a runner that lost its ICD would turn the sanitized half of the
+        # render path into a silent no-op and this job would stay green, which is the same shape as
+        # the CodeQL database that analysed two files. linux-gcc16-build.yml guards its pixel step
+        # the same way and for the same reason.
+        LOG="$RUNNER_TEMP/asan-pixel.log"
+        ctest --test-dir build-asan -R offscreen_tests --output-on-failure --no-tests=error 2>&1 | tee "$LOG"
+        if grep -q "Skipped" "$LOG"; then
+          echo "::error::offscreen_tests was skipped - no Vulkan device, so nothing in the render path was sanitized"
+          exit 1
+        fi
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,132 @@
+name: Sanitizers
+
+# AddressSanitizer and UndefinedBehaviorSanitizer over the test suite (issue #179).
+#
+# This exists because a green ordinary build proves less than it looks like it does here. ASan
+# found a heap-use-after-free in TruetypeTests.cpp during #171 - a test parsing from a temporary
+# while `tt::Font` held non-owning spans - and found a second one of exactly the same shape in
+# SchemaTests.cpp the first time this workflow was written. Both suites had been passing for
+# months. The tree is full of hand-rolled binary parsing over spans, which is the code where that
+# bug lives, so the instrumented run is not a formality.
+
+on:
+  push:
+    branches: [ main, develop ]
+  pull_request:
+    # Matches the PR's *base*, not its head - see linux-gcc16-build.yml for why the list reads
+    # this way and what `[0-9]+-*` is.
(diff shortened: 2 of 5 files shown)


ambroise-leclerc:
CI is green, including the new job: **ASan + UBSan (GCC 16) — 3m5s**, which is faster than the ordinary Linux leg (3m20s). That removes the reason #179 suggested keeping it off PRs, so it runs on them.

Verified in CI rather than only locally: 435/435 with leak detection on, then `offscreen_tests` separately with it off, reported `Passed` rather than `Skipped`.

Added one thing after opening this: the render step now fails if that suite was **skipped**. `offscreen_tests` exits 77 with no Vulkan device and `tests/CMakeLists.txt` sets `SKIP_RETURN_CODE 77`, so a runner that lost its ICD would have turned the sanitized half of the render path into a silent no-op while the job stayed green — the same shape as the CodeQL database that analysed two files. `linux-gcc16-build.yml` already guards its pixel step this way; this now matches. `--no-tests=error` on both invocations too, so a filter matching nothing is a failure rather than a pass.

camille-martin-paris (changes requested):
Blocking finding:

**P1 — Preserve the rendered-test failure status through `tee`.** In `.github/workflows/sanitizers.yml` around line 141, the implicit Actions shell does not guarantee `pipefail`. If `ctest` fails but `tee` succeeds, the step continues and can remain green, defeating the sanitizer gate for `offscreen_tests`.

Please enable `set -o pipefail` before the pipeline or explicitly select a Bash shell configuration that enables it, while preserving the log capture and skipped-test check.

All current checks pass, including the sanitizer job, but that successful run does not exercise this failure-propagation path.

ambroise-leclerc:
Fixed, and the finding was correct in a way that stings — it is the same silent-pass shape I spent this branch hunting, in my own new code.

Demonstrated rather than taken on trust:

```console
$ bash -e -c 'false | tee /dev/null; echo $?'
0
$ bash -e -c 'set -o pipefail; false | tee /dev/null; echo $?'
1
```

A `run:` block gets `bash -e`, **not** `bash -eo pipefail`, so `ctest ... | tee "$LOG"` reported tee's status and a failing test run would have passed the step. The skip-guard I added to catch a vacuous run would itself have been unable to fail on a real one.

## It was not only my copy

`linux-gcc16-build.yml`'s "Verify rendered output" step — which is where I copied the `| tee` pattern from — has the identical defect and predates this branch. Its pixel-test failures have been maskable the whole time, on the leg that exists specifically to prove the pixel tests are not silently skipped.

Fixed in the same commit. It is technically outside this PR's subject, but it is one line, it is the same defect, it was found only because of this review, and leaving a known CI-masking bug in place to keep a diff tidy is the wrong trade. Flagging it explicitly so it is a decision rather than something slipped in.

Grepped the rest of `.github/workflows/` for the pattern: those two were the only pipelines through `tee`; `clang-build.yml` already sets `set -euo pipefail` itself.

PR #182, opened 2026-08-09
docs: require @file, and add it to the 56 files that lacked it (#180)

### LINKED ISSUE #180: Settle the @file convention: the written rule contradicts the tree
## The contradiction

`CONTRIBUTING.md` says:

> **File-level documentation:** Use `@brief` only, no `@file` or `@author` tags.

The tree does the opposite: **82** files carry `@file`, **56** do not. Review bots cite `CONTRIBUTING.md` as authoritative, so the rule surfaces repeatedly on unrelated PRs.

## Why the rule is the wrong half to keep

Removing `@file` does not merely relocate documentation — it **misattributes** it. Measured with Doxygen 1.15 on `tests/framework/RunRecords.hpp` from this tree:

| | file compound `briefdescription` | where the text goes |
|---|---|---|
| with `@file` | "The v1 glyph-run record encoder, shared by…" | `RunRecords_8hpp.xml` |
| without `@file` | **EMPTY** | `namespacemdux_1_1spec.xml` |

So `namespace mdux::spec` ends up documented as "The v1 glyph-run record encoder, shared by the governed and rendered text suites." That is wrong documentation rather than missing documentation, and it is silent — Doxygen emits no warning, it just attaches the block to the next entity it finds.

For `.cppm` module interfaces the block lands on the module compound instead, which is defensible. For plain headers and `.cpp` files it lands on a namespace, which is not.

Doxygen requires `\file` to document a file. The rule as written asks for something the tool does not support.

## Scope

- Correct `CONTRIBUTING.md`: file-level blocks **must** open with `@file <basename>`, followed by `@brief`.
- Add the missing tag to the 56 files that lack it. All 56 already carry a leading `/** */` block, so this is inserting one line into an existing comment, not writing new documentation.
- Leave `@author` prohibited — that part of the rule is about attribution churn and is unrelated.

## Done when

- No tracked `.cpp`/`.cppm`/`.hpp`/`.h` file has a file-level block without `@file`.
- `CONTRIBUTING.md` states the rule the tree follows, with the misattribution as the reason, so it is not re-litigated.

ambroise-leclerc:
Closes #180.

`CONTRIBUTING.md` said file-level blocks should use "`@brief` only, no `@file`". The tree said otherwise — **82** files carried the tag against **56** that did not — and since review bots cite `CONTRIBUTING.md` as authoritative, the contradiction kept surfacing on unrelated PRs.

## Why the rule was the wrong half to keep

Removing `@file` does not relocate documentation. It **misattributes** it. Measured on a real file from this tree with Doxygen 1.15:

| | file compound brief | where the sentence goes |
|---|---|---|
| with `@file` | "The v1 glyph-run record encoder, shared by…" | `RunRecords_8hpp.xml` |
| without `@file` | **EMPTY** | `namespacemdux_1_1spec.xml` |

So `namespace mdux::spec` ends up documented as *"The v1 glyph-run record encoder, shared by the governed and rendered text suites."* Doxygen emits no warning — it just attaches the block to the next entity it finds. That is wrong documentation rather than missing documentation, which is the worse of the two failure modes, and `@file` is the tag Doxygen provides precisely to prevent it.

For `.cppm` module interfaces the block lands on the module compound instead, which is defensible. For plain headers and `.cpp` files it lands on a namespace, which is not.

## The change

- `CONTRIBUTING.md` now requires `@file <basename>` first, then `@brief`, with the measurement above as the stated reason so it is not re-litigated. `@author` stays prohibited — that half of the rule is about attribution churn and is unrelated.
- One line inserted into each of the 56 files. **All 56 already had a leading `/** */` block**, so nothing here writes or invents documentation; it tags what was already there. 56 files changed, 56 insertions, 0 deletions.

## Verification

- Every tracked `.cpp`/`.cppm`/`.hpp`/`.h` file now carries the tag — the missing count is 0.
- Confirmed the tag does what it is for: `include/mdux/draw/Draw.cppm` now reports "Governed-zone fixed-budget draw types: what a frame is, before Vulkan sees it" on its **file** page, where before the change that page had no brief at all.
- 436 tests pass.

Reviewer note: the diff is 56 identical one-line insertions, so it reads faster from `git show --stat` than file by file.

## Summary by CodeRabbit

* **Documentation**
  * Standardized file-level Doxygen comments by requiring an `@file` annotation with each filename and extension before `@brief`.
  * Added file-identification annotations across source, module, example, test, tool, and header documentation.
  * Updated contribution guidelines to reflect the revised documentation standard.

* **Behavior**
  * No functional behavior, public declarations, or API changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,19 @@ We follow the **C++ Core Guidelines** to ensure conformance with modern C++23 ge
 
 We use **Doxygen** syntax for code documentation. Follow these guidelines:
 
-- **File-level documentation:** Use `@brief` only, no `@file` or `@author` tags.
+- **File-level documentation:** open the block with `@file <basename>`, then `@brief`. Still no
+  `@author` tags.
+
+  `@file` is what makes Doxygen attach the block to the *file*. Without it the block is not
+  ignored — it silently becomes the documentation of whatever comes next. Measured on
+  `tests/framework/RunRecords.hpp` with Doxygen 1.15: with the tag, the file page carries "The v1
+  glyph-run record encoder, shared by…"; without it, the file page's brief is empty and that
+  sentence turns up as the documentation for `namespace mdux::spec`. Wrong documentation, with no
+  warning, rather than missing documentation.
+
+  This rule previously said the opposite. It was never what the tree did — 82 files used `@file`
+  against 56 that did not — and review bots cite this document, so the contradiction kept
+  resurfacing on unrelated pull requests. #180 settled it in favour of the half that works.
 - **Class documentation:** Include `@brief` with detailed description and usage examples.
 - **Method documentation:**
   - Use compact notation `/** @brief Description */` for simple one-line descriptions.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,8 +35,18 @@ We follow the **C++ Core Guidelines** to ensure conformance with modern C++23 ge
 
 We use **Doxygen** syntax for code documentation. Follow these guidelines:
 
-- **File-level documentation:** open the block with `@file <basename>`, then `@brief`. Still no
-  `@author` tags.
+- **File-level documentation:** open the block with `@file`, naming the file **with its
+  extension and no path**, then `@brief`. Still no `@author` tags.
+
+  ```cpp
+  /**
+   * @file VulkanRenderer.cppm
+   * @brief One sentence on what this file is.
+   */
+  ```
+
+  The extension matters: `Draw.cppm` and `Draw.cpp` are different files with the same stem, and
+  this tree has several such pairs.
 
   `@file` is what makes Doxygen attach the block to the *file*. Without it the block is not
   ignored — it silently becomes the documentation of whatever comes next. Measured on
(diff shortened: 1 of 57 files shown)


camille-martin-paris (approved):
No blocking findings. I verified that all 56 modified C++ files use an `@file` value matching the complete filename, and all checks pass.

Non-blocking wording suggestion: use `@file <filename>` instead of `@file <basename>` in `CONTRIBUTING.md`. “Basename” is sometimes understood to exclude the extension, while the intended convention includes it. A short example would make the rule unambiguous.

ambroise-leclerc:
Good catch — fixed, and it would have mattered.

`@file` now reads "naming the file **with its extension and no path**", with a worked example. "basename" reads as "without extension" to about half of the people who meet it, which would have reintroduced precisely the churn this PR exists to end.

The extension is load-bearing rather than pedantic here: **31 stems in this tree name more than one file** — `Draw.cppm`/`Draw.cpp`, `AtlasPacker.cppm`/`AtlasPacker.cpp`, `Compliance.cppm`/`Compliance.cpp` and so on — so a stem alone does not identify a file at all.

Also verified while here that all **139** tags match their own filename exactly, with no path prefix and no typo: zero mismatches.

Issue #184, opened 2026-08-09
Relicensing consent: EUPL-1.2 and alternative commercial terms


ambroise-leclerc:
Compliatory intends to make this repository available under a dual-licensing model:

1. the European Union Public Licence 1.2 (`EUPL-1.2`); or
2. separate commercial terms offered by the project steward.

The audit of the default branch identified original, non-merge commits authored by:

- @camille-martin-paris
- @0xfritz-dev
- @mucas-benoit
- @amelie-roussel-afnor
- @AM-L

### Consent requested

If you agree, please reply to this issue with exactly **`I agree`** from the GitHub account associated with your contribution.

By doing so, you confirm that, to the best of your knowledge, you are entitled to license your contributions and grant **Ambroise Leclerc**, as project steward, a perpetual, worldwide, non-exclusive, royalty-free licence to use, reproduce, modify, distribute, sublicense and relicense your contributions already present in this repository under:

- EUPL-1.2; and
- alternative commercial terms.

This is not a transfer of ownership. Your contributions will remain available under EUPL-1.2, and your attribution will remain in the Git history. The grant also includes a royalty-free patent licence for patent claims you control that are necessarily infringed by your contributions alone or in combination with this repository.

If you do not agree, or believe the audit incorrectly attributes a contribution to you, please say so here. The affected contribution will then be reviewed for removal or independent rewriting before relicensing.

This request concerns contributions already present on the default branch at the date this issue was opened. A separate contributor agreement will govern future invited contributions.

> Note: this issue records contributor consent; it is not legal advice. The project steward may have the final agreement reviewed before completing the relicensing.

camille-martin-paris:
I agree

0xfritz-dev:
I agree

mucas-benoit:
I agree

amelie-roussel-afnor:
I agree

AM-L:
I agree

ambroise-leclerc:
All identified contributors explicitly agreed. The EUPL-1.2 and commercial dual-licensing change has been merged in #185.

Issue #116, opened 2026-08-01
S10 — Enforce governed-zone source and exception policy


ambroise-leclerc:
Part of #11. Implements the enforcement still missing after ADR-004/ADR-005 and the `MduXCore` split.

## Work

- Compile governed implementation code with exceptions and RTTI disabled (`-fno-exceptions -fno-rtti` or the verified MSVC equivalents).
- Add a governed-source lint covering module interfaces and implementations; reject `throw`, `try`, `catch`, throwing `std::expected::value()`, allocation APIs, forbidden platform/filesystem/time/random APIs, unsafe casts, and `std::fma` where determinism requires separate operations.
- Keep the existing link-graph trust-zone verification and `.clang-tidy` profile as independent defenses.
- Add negative configure/build/lint fixtures proving each forbidden dependency or construct fails for the intended reason.
- Ensure the rule does not scan generated shader C arrays as hand-authored governed source.

## Acceptance

- `MduXCore` builds and tests on MSVC and both active GCC legs with exceptions/RTTI disabled.
- Every negative fixture fails and every existing governed module passes.
- ADR-004/ADR-005 and agent guidance describe only mechanisms CI actually enforces.

PR #187, opened 2026-08-11
Move mdux.text.raster to the host-tools zone [#116]

ambroise-leclerc:
## Summary

First of three for #116. The rasteriser was governed on a speculative argument: a device path might one day need to rasterise, and ADR-008 decision 1 would then require it to be *this* module rather than a second one. The argument was hypothetical; the cost was not.

`rasterise()` allocates, and `std::vector` reports failure by throwing, so `Raster.cpp` carries a `try`/`catch` at its `noexcept` boundary to turn an out-of-memory condition into a diagnostic instead of `std::terminate`. That block is well-reasoned and **stays exactly as written** — but it is incompatible with the governed zone's no-throwing rule, which the next PR in this chain starts enforcing. A real constraint outranks a hypothetical consumer, so the zone moves rather than the code.

The module runs once per glyph at build time and never on a device, which its own header comment already said. Its only consumers are `mdux-textbake` and its tests.

- `include/mdux/text/Raster.cppm` → `tools/text/Raster.cppm`
- `src/text/Raster.cpp` → `tools/text/Raster.cpp`
- registered on `MduXTextBakeLib`, dropped from `MduXCore`
- `RasterTests.cpp` moves from `text_spec` to `text_tools_spec`
- the stale "raster is governed" rationales in `AtlasPacker.cppm` and `TextBake.cpp` are corrected rather than left to mislead

The test move is not bookkeeping. `text_spec` links `MduX::Core` alone, and `tests/CMakeLists.txt` states that this link line **is** its standing evidence that governed text code reaches nothing but `std`. Leaving raster cases there would force `MduX::TextBakeLib` onto the target and destroy the very thing it demonstrates.

The module name `mdux.text.raster` is unchanged, so no `import` statement anywhere moves.

### Breaking change

This removes `mdux.text.raster` from the **installed surface** — `MduXTextBakeLib` is not in the install/export set — which affects any 0.5.0 consumer that imported it. A build-time baker has no business in a device consumer's include tree, so the removal is intended. `InstallTreeConsumer` passes, confirming nothing that remains depends on it.

Part of #116.

Invitation / CLA issue: #184

## Dependency

- **Base branch:** `develop`
- **Merge order:** mergeable now. It is the predecessor of #188 and #189; both must wait on it.

## Shared registries touched

- [x] `CMakeLists.txt` (root) — `MduXCore` loses two sources
- [x] `FILE_SET CXX_MODULES` lists — `Raster.cppm` moves between targets
- [x] `tools/CMakeLists.txt` — `MduXTextBakeLib` gains both files
- [x] `tests/CMakeLists.txt` — `RasterTests.cpp` moves suite

## Rebase state

- **Rebased onto:** `e96dc35` (current `develop`)

## Verification

Clean configure and build from an empty build directory, GCC 16.1.0 / CMake 4.2.3 / Ninja.

- [x] `cmake --preset ninja-gcc && cmake --build build-gcc` — clean; `mdux_verify_trust_zones()` passes at configure
- [x] `ctest --test-dir build-gcc` — 436/436 passed, the same count as before the move
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK (78 files, 0 findings)
- [x] Other:
  - `ctest -R "text_spec|text_tools_spec"` — raster cases now run under `text_tools_spec` (44 cases, was 41); `text_spec` is down to 26 and still links `MduX::Core` alone
  - `InstallTreeConsumer` — passed (7.32s)
  - installed to a scratch prefix and confirmed `mdux.text.raster` is absent from the install tree, and `Schema.cppm`/`Draw.cppm` are still present
  - `evidence` label — 5/5, so the committed `generated/font/dejavu-ui/` atlas still byte-compares with the rasteriser in its new home
  - `mdux-evidence-lint` — OK (38 files)

Not run locally: the MSVC leg. CI covers it.

## Post-merge gate

## Regulatory impact

**Trust-zone boundary change**, so ADR-004 applies. A module moves from the governed zone to host tools; the governed zone shrinks, which is the safe direction. No governed module imports the moved one — verified by grep across `include/` and `src/` — and `mdux_verify_trust_zones()` passes.

The determinism properties are unchanged: the rasteriser is integer-only, and its committed output is still byte-compared in CI on both toolchain legs. The ADR-004 text recording this placement rule lands with #189, so that the ADR describes what merged rather than what was intended.

## Summary by CodeRabbit

* **Refactor**
  * Moved text glyph rasterization into the host-only text baking workflow.
  * Included rasterization with atlas generation instead of the core runtime.
  * Clarified that rasterization and atlas packing occur at build time, while devices consume generated atlas data.

* **Documentation**
  * Updated architecture, trust-boundary, and text pipeline documentation to reflect the revised component classification.

* **Tests**
  * Reclassified rasterization tests under the text tools test suite.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,12 +51,20 @@ Makefile
 *.user
 *.sln.docstates
 
-# Documentation files (large PDFs, proprietary standards)
-inputs/Documentation/*.pdf
-inputs/Documentation/*.epub
-inputs/Documentation/ISO\ 13485/
-inputs/Documentation/ISO\ 14971\ The\ Definitive\ Guide/
-inputs/Documentation/Medical\ Device\ Cybersecurity\ for\ Engineers\ and\ Manufacturers/
+# Local reference material - never committed.
+#
+# The whole directory, not selected paths inside it. Five specific sub-paths were listed here
+# until issue #116, which left inputs/CMake/ and any newly-added folder under
+# inputs/Documentation/ untracked-but-not-ignored - so a `git add -A` would have committed them.
+# That is the exact class of content #7 and #23 spent a history rewrite removing, and ADR-006
+# forbids outright. A directory-level rule cannot be defeated by adding a file to it.
+#
+# This lives in the *first* PR of the #116 stack rather than the last, because the narrow rule
+# actually caught someone out while that stack was in review: a `git add -A` on this branch staged
+# `inputs/CMake` as a gitlink. It was a one-line SHA rather than the 431 MB behind it, and it was
+# amended out before anyone pulled, but a guard that lands after the thing it guards against is
+# not a guard.
+inputs/
 
 # Temporary files
 *.log
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,6 @@ target_sources(MduXCore
             include/mdux/font/Schema.cppm
             include/mdux/text/Draw.cppm
             include/mdux/text/Schema.cppm
-            include/mdux/text/Raster.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
@@ -209,7 +208,6 @@ target_sources(MduXCore
         src/font/Schema.cpp
         src/text/Draw.cpp
         src/text/Schema.cpp
-        src/text/Raster.cpp
         src/governance/Governance.cpp
         src/governance/Justification.cpp
         src/governance/Program.cpp
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -53,7 +53,6 @@ are ordinary `PRIVATE` sources.
 | `mdux.governance.compliance` | `include/mdux/governance/Compliance.cppm` | `src/governance/Compliance.cpp` |
 | `mdux.shader.schema` | `include/mdux/shader/Schema.cppm` | `src/shader/Schema.cpp` |
 | `mdux.text.schema` | `include/mdux/text/Schema.cppm` | `src/text/Schema.cpp` |
-| `mdux.text.raster` | `include/mdux/text/Raster.cppm` | `src/text/Raster.cpp` |
 | `mdux.font.schema` | `include/mdux/font/Schema.cppm` | `src/font/Schema.cpp` |
 | `mdux.text.draw` | `include/mdux/text/Draw.cppm` | `src/text/Draw.cpp` |
 | `mdux.draw` | `include/mdux/draw/Draw.cppm` | `src/draw/Draw.cpp` |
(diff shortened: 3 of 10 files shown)


0xfritz-dev (dismissed):
Approve. The CMake registration moves cleanly in both directions (FILE_SET and PRIVATE sources), RasterTests.cpp moves to text_tools_spec with the link-line rationale documented, and the stale governed-rationale comments in AtlasPacker.cppm/TextBake.cpp are corrected rather than deleted. CI green on every leg.

One doc nit for when #189 lands: three places here (docs/architecture.md, the tools/CMakeLists.txt comment, and Raster.cppm) say the module could not stay in a target *compiled with* -fno-exceptions. #189 records that MduXCore cannot be compiled that way at all (import std vs the shared std BMI), so the accurate conflict is the no-throw rule, not the compile flag. Worth a one-line alignment so the two PRs don't contradict each other after merge.

ambroise-leclerc:
Both review findings addressed.

**Copilot — the `-fno-exceptions` comment (suppressed comment on `Raster.cpp:585`).** This was the most valuable finding across the four PRs, and it was right: the comment stated as present-tense fact that `MduXCore` is compiled with `-fno-exceptions`. It is not, this PR does not make it so, and #188 subsequently established that it *cannot* be — `import std` and `-fno-exceptions` are mutually exclusive on GCC, because the std BMI records its dialect and CMake synthesises one shared std target for the whole build.

So the comment was wrong when written and would have stayed wrong. Worse, it was wrong in exactly the way issue #116 exists to eliminate: a mechanism asserted in the present tense that nothing implements.

The real justification for the move is the no-throw rule itself, which does not depend on any compile flag and *is* enforced as of #116 — `mdux-governed-lint` rejects `try`/`catch` in governed source on every toolchain, and `governed.noThrow.symbolScan` rejects the resulting `__cxa_throw` in the object on GCC/Clang. Corrected in all five places the flag was cited: `Raster.cpp`, `Raster.cppm`, `AtlasPacker.cppm`, `tools/CMakeLists.txt` and `docs/architecture.md`.

**@coderabbitai — the stale `Outline` doxygen.** Also correct. The type's own comment still said it exists because a governed module may not import a host-tools one — true until this PR, stale the moment it merged, and the public documentation was giving two different reasons for the same type. It now says the restatement was forced until #116 and is kept by choice, so the module stays movable back into `MduXCore` if a device path is ever built allocation-free. That matches the module overview instead of contradicting it.

**Also added to this PR:** the `.gitignore` tightening to a directory-level `inputs/` rule, moved here from the last PR of the stack. While addressing this review, a `git add -A` staged `inputs/CMake` as a gitlink — a one-line SHA rather than the 431 MB behind it, amended out before anyone pulled, but a guard that lands after the thing it guards against is not a guard.

436/436 on GCC 16; documentation lint clean.

camille-martin-paris (approved):

PR #188, opened 2026-08-11
Enforce ADR-005's no-throwing rule on the emitted objects [#116]

ambroise-leclerc:
## Summary

Second of three for #116. ADR-005 says the governed zone does not throw. Nothing checked it — the claim was made in the present tense at `ADR-005:41`, the lint said to reject `throw`/`try`/`catch` had never been written, and `bugprone-exception-escape` (the check ADR-005 *opens by citing* as the reason a `noexcept` violation was invisible) was still suppressed in `.clang-tidy`.

### The intended mechanism is not available

ADR-005 specified `-fno-exceptions` on `MduXCore`. That cannot be done while the governed zone uses `import std`. Verified directly rather than inferred:

- GCC records the language dialect in every module BMI. A governed module compiled with the flag is rejected against the shared `std` BMI with `language dialect differs 'C++23', expected 'C++23/no-exceptions/no-rtti'`. **Each of `-fno-exceptions` and `-fno-rtti` fails this way on its own**, not only in combination.
- CMake synthesises exactly one `__CMAKE__CXX23@synth_*` std target for the whole build, shared by the adapter and host-tools targets, which must keep exceptions. Adding the flags to `MduXCore` did not produce a second one.
- A separately built governed-dialect `std` BMI *does* compile and work when driven by hand, so the constraint is the **sharing**, not the flags. Hand-rolling one would displace the most fragile part of this build.

So the rule is enforced where it can be checked properly — in the objects.

### What landed instead

`MduXNoHeapScan.cmake` grows a profile selector, and `governed.noThrow.symbolScan` runs the new `governed-throw` profile over all 29 `MduXCore` objects. This is the stronger of the two available checks: `-fno-exceptions` stops you *writing* a throw, while the symbol scan sees a throw arriving through a std facility the source never spells out.

The allocator symbols do **not** generalise from the ML profile and are excluded — `mdux.evidence.json` and `mdux.governance` allocate by design.

### What it tolerates, and says so

The scan forbids a literal throw (`__cxa_throw`), and no governed object references it. It deliberately tolerates libstdc++'s throw *helpers*: eight objects reference `__throw_length_error`, `__throw_logic_error` or `__throw_out_of_range_fmt`, all from inside `std::string`, `std::vector` and `std::string_view::substr` — `Json.cpp`'s parser and `Governance.cpp`'s id splitting are both of that shape, and in both the caller's invariant makes the throw unreachable.

Forbidding those means banning those types from the governed zone, which is a larger decision than this issue and needs its own ADR. Until then the scan **prints all 14 references on every run** rather than implying they do not exist.

Part of #116.

Invitation / CLA issue: #184

## Dependency

- **Base branch:** `116-move-raster-to-host-tools` (stacked, so this PR shows one issue's diff)
- **Predecessor PR:** #187
- **Merge order:** **must not merge until #187 has merged.** The scan fails on `develop` as it stands today — `src/text/Raster.cpp` is governed and contains `try`/`catch`, which is precisely what #187 resolves.

## Shared registries touched

- [x] `tests/CMakeLists.txt` — a second scan registration, the negative fixture object library and its test
- [x] None of the others

## Rebase state

- **Rebased onto:** `913fe84` (predecessor head), itself on `e96dc35` (current `develop`)

## Verification

- [x] `cmake --preset ninja-gcc && cmake --build build-gcc` — clean
- [x] `ctest --test-dir build-gcc` — 438/438 passed (436 + the two new `governed` tests)
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK
- [x] Other:
  - `ctest -L governed` — both pass; the scan reports `29 governed object(s) contain no throw expression` and lists all 14 tolerated helper references
  - `ctest -L noheap` — 5/5, the existing ML scan unaffected by the profile split
  - the negative fixture was confirmed to fail for the *right reason*, matching `Symbol scan violation (profile 'governed-throw')` on `ThrowFixture.cpp.o: U __cxa_throw`

Not run locally: the MSVC leg. It is the least-exercised part of this PR — `dumpbin /SYMBOLS` rather than `nm -u -C`, and `_CxxThrowException` rather than `__cxa_throw` — so the Windows check on this PR is the gate, not a formality.

## Post-merge gate

## Regulatory impact

**Adds a compliance-relevant check, and narrows a claim.** ADR-005's "no throwing" assertion becomes mechanically verified for literal throws, which it was not before.

It also establishes that the governed zone **cannot** currently be built with `-fno-exceptions`, contradicting a consequence ADR-005 lists as positive. That is a reduction in what the project may claim, not an increase. The ADR text recording both the new mechanism and the retracted claim lands with #189 — deliberately last, so the documents describe what merged.

`bugprone-exception-escape` is re-enabled with a comment stating plainly that it is *available* rather than *enforced*: the only job running clang-tidy does so with `|| true` against a compile database it cannot parse for module interface units.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,11 +2,22 @@
 # Based on C++ Core Guidelines
 # Enforces coding standards from CONTRIBUTING.md
 
+# bugprone-exception-escape is deliberately NOT disabled. It was, until issue #116: ADR-005 opens
+# by citing that suppression as the reason a `noexcept` violation in governed code was invisible to
+# static analysis, and listed re-enabling it as a follow-up gated on MduXCore existing. MduXCore has
+# existed since Wave 1.
+#
+# Scope, stated honestly: this check is *available* rather than *enforced*. The only CI job running
+# clang-tidy is security-analysis.yml, which invokes it with `|| true` against a GCC-generated
+# compile_commands.json that clang-tidy cannot parse for module interface units - both limitations
+# are documented in that workflow. So re-enabling this catches a noexcept violation for a developer
+# running clang-tidy locally, and will catch one in CI when the Clang leg is restored (#48), but it
+# is not what stops one landing today. That is governed.noThrow.symbolScan, which reads the emitted
+# objects and runs on every ctest.
 Checks: >
   -*,
   bugprone-*,
   -bugprone-easily-swappable-parameters,
-  -bugprone-exception-escape,
   cert-*,
   -cert-dcl21-cpp,
   -cert-dcl50-cpp,
--- a/.github/workflows/linux-gcc16-build.yml
+++ b/.github/workflows/linux-gcc16-build.yml
@@ -108,6 +108,15 @@ jobs:
     - name: Verify ML determinism (GCC 16)
       run: ctest --preset ninja-gcc -L determinism --output-on-failure
 
+    # Governed-zone no-throw, at the object level (ADR-005, issue #116). Broken out with -V rather
+    # than folded into the full run above, because a passing ctest prints none of a test's output -
+    # and this scan's output is half its purpose. It reports the throw-helper references it
+    # deliberately tolerates, and the ADR justifies tolerating them on the grounds that they stay
+    # "visible and counted". A number nobody can see in a job log is not counted. -V is what makes
+    # the claim true, and is the same reason the pixel step below uses it.
+    - name: Verify governed-zone no-throw (GCC 16)
+      run: ctest --preset ninja-gcc -L governed -V --no-tests=error
+
     # Rendered-truth verification (issue #125), moved here when the GCC 15 leg was retired.
     # Without it a broken ICD turns the only test that renders anything into a silent no-op:
     # offscreen_tests exits 77 with no device and CTest reports Skipped, which does not fail a run.
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -115,6 +115,16 @@ jobs:
       run: ctest --preset ninja-msvc -L determinism
       shell: cmd
 
+    # Governed-zone no-throw (ADR-005, issue #116). Informational on this toolchain, and run
+    # anyway: the MSVC STL inlines its own throw sites, so _CxxThrowException in a governed object
+    # is the same symbol whether it came from a hand-written throw or a std::string growth path.
+    # The scan therefore forbids nothing here and prints what it found, which is worth having in
+    # the log - it is the only place the MSVC-side references are visible at all. mdux-governed-lint
+    # is what actually gates the rule on Windows. -V because a passing ctest prints no test output.
+    - name: Verify governed-zone no-throw (MSVC, informational)
+      run: ctest --preset ninja-msvc -L governed -V --no-tests=error
+      shell: cmd
+
     - name: Verify no source-tree writes
       shell: pwsh
       run: |
--- a/.gitignore
+++ b/.gitignore
@@ -51,12 +51,20 @@ Makefile
 *.user
 *.sln.docstates
 
-# Documentation files (large PDFs, proprietary standards)
-inputs/Documentation/*.pdf
-inputs/Documentation/*.epub
-inputs/Documentation/ISO\ 13485/
-inputs/Documentation/ISO\ 14971\ The\ Definitive\ Guide/
-inputs/Documentation/Medical\ Device\ Cybersecurity\ for\ Engineers\ and\ Manufacturers/
+# Local reference material - never committed.
+#
+# The whole directory, not selected paths inside it. Five specific sub-paths were listed here
+# until issue #116, which left inputs/CMake/ and any newly-added folder under
+# inputs/Documentation/ untracked-but-not-ignored - so a `git add -A` would have committed them.
+# That is the exact class of content #7 and #23 spent a history rewrite removing, and ADR-006
+# forbids outright. A directory-level rule cannot be defeated by adding a file to it.
+#
+# This lives in the *first* PR of the #116 stack rather than the last, because the narrow rule
+# actually caught someone out while that stack was in review: a `git add -A` on this branch staged
+# `inputs/CMake` as a gitlink. It was a one-line SHA rather than the 431 MB behind it, and it was
+# amended out before anyone pulled, but a guard that lands after the thing it guards against is
+# not a guard.
+inputs/
 
 # Temporary files
 *.log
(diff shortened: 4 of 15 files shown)


ambroise-leclerc:
## MSVC leg failed, and it was right to

The Windows check on the first push reported nine governed objects referencing `_CxxThrowException`:

```
Json.cpp.obj, Report.cpp.obj, Schema.cpp.obj (x3),
Governance.cpp.obj, Justification.cpp.obj, Program.cpp.obj, Compliance.cpp.obj
```

`mdux-governed-lint` (PR #189) independently confirms none of those sources contains a `throw`. The explanation is in which objects are *absent*: `Digest`, `Kernels`, `Runtime` and `Draw` are not listed, and the nine that are match almost exactly the GCC set that references `__throw_length_error`.

**The MSVC STL inlines its own throw sites.** A `std::string` growth path emits `_CxxThrowException` straight into the caller's object, where libstdc++ emits a call to an out-of-line `__throw_*` helper. So the distinction this profile rested on — forbid the literal throw, tolerate the std helper — is a property of libstdc++'s code generation, not a portable one. On MSVC the two are the same symbol.

Forbidding it there fails the build on correct code. Tolerating it forbids nothing. Either way the symbol cannot carry the claim.

### What changed

`1f3890e` makes the profile toolchain-aware. Under `dumpbin` it forbids nothing, reports `_CxxThrowException` and the `_X*` helpers, and its closing message says **no verdict is claimed** rather than "N objects contain no throw expression" — asserting the latter while checking nothing is precisely the shape of unearned claim this issue exists to remove.

The negative fixture is not registered on MSVC either, and the configure step says so rather than skipping quietly. A negative test needs something to be forbidden before it can prove the rejection works.

**The rule is still enforced on Windows.** `mdux-governed-lint` is toolchain-independent and rejects `throw` in governed source everywhere. What is GCC-specific is only the object-level corroboration — and #189's ADR text now says exactly that, rather than describing the scan's verdict as universal.

### Verified

- GCC unchanged: 438/438, negative fixture still passing
- the `dumpbin` branch exercised through a shim emitting MSVC-shaped symbol output — confirmed it reports, exits 0, and prints the no-verdict message

This also corrects something the PR description above claims. The scan is described there as "the stronger of the two available checks"; that holds on GCC/Clang and not on MSVC, where the source lint is the only gate. #189 carries the corrected wording in ADR-005.

0xfritz-dev (changes requested):
Request changes: the Windows (MSVC) leg fails on `governed.noThrow.symbolScan` — 9 governed objects (Json, Report, Schema x3, Governance, Justification, Program, Compliance) reference `_CxxThrowException`.

Root cause is structural, not a real violation: MSVC's STL throws inline through `_CxxThrowException` — the *same symbol* the compiler emits for a literal `throw`. The design here works on GCC because libstdc++ routes container/string throws through distinct `__throw_*` helpers the scan can tolerate; MSVC has no equivalent seam, so 'forbid literal throws, tolerate std helpers' is not implementable on MSVC by symbol scan. The GCC leg confirms no source contains a literal throw (no `__cxa_throw`), and #189's GOV001 source lint confirms it cross-platform.

Fix options, in decreasing order of preference:
1. Demote `_CxxThrowException` to `reported_symbols` when the tool is `dumpbin`, printing it every run exactly like the libstdc++ helpers, and document that the object-level layer is only discriminating on Itanium-ABI toolchains (source lint GOV001 provides the MSVC-side enforcement).
2. Skip the test on MSVC with a deliberate, non-silent warning analogous to the missing-tool case.

Whichever lands, the ADR-005 'What is enforced' wording in #189 ('two layers, both required to pass') must reflect it — as written it describes a layer that fails on one of the two supported platforms.

Minor: `tests/governed/ThrowFixture.cpp:18` says the negative test 'carries `WILL_FAIL`', but tests/CMakeLists.txt:811 deliberately uses `PASS_REGULAR_EXPRESSION` and explains why `WILL_FAIL` must not be added — stale docstring, please fix.

Notably good: the negative-fixture design (matching the message rather than the exit status), the print-every-run treatment of tolerated helpers, and the profile selector's refuse-to-default behaviour.

ambroise-leclerc:
@0xfritz-dev — thank you, this was a good catch and the diagnosis was exactly right. Both points addressed.

**1. MSVC blocker — fixed via your option 1.**

`_CxxThrowException` is demoted to `reported_symbols` when the tool is `dumpbin`, printed every run exactly like the libstdc++ helpers. Your framing of the root cause matches what I found independently before your review landed: the nine objects (`Json`, `Report`, `Schema`×3, `Governance`, `Justification`, `Program`, `Compliance`) are almost exactly the GCC set that references `__throw_length_error`, and the absentees (`Digest`, `Kernels`, `Runtime`, `Draw`) are the allocation-free ones. MSVC has no seam because it inlines the throw site; libstdc++ has one because it calls out.

Two things beyond option 1 as stated:

- The closing message no longer says "N objects contain no throw expression" under `dumpbin`. It says no verdict is claimed and points at the source lint. Reporting the symbol while still printing a verdict the scan did not establish would have been the same defect one layer up.
- The negative fixture is not registered on MSVC either, with a non-silent configure message — your option 2's treatment, applied to the one test that genuinely has nothing to prove there. A negative test needs something to be forbidden.

**2. ADR-005 wording — corrected in #189.**

You were right that "two layers, both required to pass" described a layer failing on one of two supported platforms. The layer table now carries a **"gates on"** column (`every toolchain` / `GCC/Clang; reports only on MSVC` / `nothing yet`), with a line under it saying only the first is a gate everywhere and that the no-throw claim therefore rests on the source lint. "What is therefore claimed" gained a third not-established entry: on MSVC the property is single-sourced. `ADR-004`'s compliance-required wording now reads "verified at source level on every toolchain and additionally in the emitted objects on GCC/Clang", and `docs/architecture.md` matches.

**3. Stale `WILL_FAIL` docstring — fixed.** Copilot flagged it independently. It now describes `PASS_REGULAR_EXPRESSION`, says why `WILL_FAIL` must not be added, and records the MSVC non-registration.

Also fixed from Copilot's review on this PR: the empty-list `FATAL_ERROR` claimed the objects should be part of `MduXCore`, which would have misdirected anyone debugging the negative fixture's list generation; and the header listed three tolerated symbols where the script tracks four.

**One follow-up I have not done here**, since it is beyond this issue: MSVC now has no independent object-level check at all. Getting one means attributing `_CxxThrowException` references to their enclosing function via `dumpbin /DISASM` and filtering std-internal frames. Happy to file it if you think it is worth having rather than leaving the source lint as the single Windows source.

Verified: 438/438 on GCC 16 with the negative fixture passing; the `dumpbin` branch exercised through a shim emitting MSVC-shaped symbol output; and the real Windows leg now green, with the 22 tolerated references visible in the job log — a separate fix, since a passing `ctest` prints no test output and the ADR's "visible and counted" justification did not hold as shipped.

camille-martin-paris (approved):

0xfritz-dev (approved):

Issue #11, opened 2026-07-26
Epic: Foundations and trust-zone skeleton


ambroise-leclerc:
The load-bearing foundations for downstream renderer, compiler, and ML work. Most original children landed before or during Wave 2; this epic now tracks only the enforcement and compiler gaps that remain.

## Children

- [x] #40 — ADR: trust zones in C++
- [x] #41 — ADR: error handling and exceptions policy
- [x] #42 — Delete dead code from examples/
- [x] #43 — .gitignore evidence/CMake negations and clean-tree CI
- [x] #44 — MduXTest and per-case discovery
- [x] #45 — Linux/Clang/debug CMake presets
- [x] #46 — MduXCore and link-graph trust-zone verification
- [x] #47 — Install/export and install-tree consumer
- [x] #48 — Resolve renderer GCC ICE/exclusion and restore Clang CI
- [x] #116 — Enforce governed-zone source and exception policy
- [x] #117 — Stack-safe PR integration and post-merge policy

## Wave 2 RETEX

Canonical types and schemas must land before their consumers. A stacked successor is rebased and re-reviewed after its predecessor merges, and the post-merge develop workflow gates the next dependent merge.

**Blocked by:** nothing.
**Blocks:** nothing downstream. The evidence kernel (#12), renderer (#13), ML (#18), documentation (#10), and font/text (#14) are all complete or unblocked. Downstream work proceeds without waiting on the two remaining enforcement children (#116, #117).
**Status:** foundations delivered in Wave 1. The epic stays open for the enforcement gaps (#116, #117) folded back in from the post-v0.2.0 trust-zone work; closing those closes this epic.

## Done when

MduXCore cannot reach Vulkan/windowing, is compiled and linted under the governed no-exception policy (per #116), the stack-safe PR and post-merge policy of #117 is encoded in repository guidance and the PR template, install-tree consumption works, MSVC/GCC build in CI, and repository delivery rules prevent the Wave 2 stacked-merge failure mode. The Clang CI leg stays disabled until #48's remaining Clang work lands; no documentation claims coverage that does not exist.

ambroise-leclerc:
All eleven children closed. The two enforcement gaps folded back in after v0.2.0 landed ahead of Wave 5:

[...]

PR #189, opened 2026-08-11
Add mdux-governed-lint and reconcile ADR-004/005 with what runs [#116]

### LINKED ISSUE #116: S10 — Enforce governed-zone source and exception policy

ambroise-leclerc:
## Summary

Last of three for #116, and the one that closes it.

`ADR-005:54` said "the governed-zone source lint also rejects `throw`, `try`, and `catch`". No such lint had ever been written. `ADR-004` item 2 described a "mandatory CI banned-include lint" that was equally absent. This is both of them, plus the ADR rewrite that makes the documents describe only mechanisms CI runs.

### The lint

Nine rules over `MduXCore`'s sources: `throw` (GOV001), `try`/`catch` (GOV002), throwing `.value()` (GOV003), raw allocation (GOV004), filesystem and console (GOV005), clocks and randomness (GOV006), unsafe casts (GOV007), `std::fma` (GOV008), and banned platform/graphics/OS headers (GOV009). Codes are stable once published, in the shared diagnostic envelope from #118.

GOV009 closes the gap ADR-004 item 1 names explicitly: denying `MduXCore` Vulkan's include *directories* does not make a system-installed `<vulkan/vulkan.h>` unreachable, because the compiler still finds it in a default search path. Only a source check does.

Two design points carry the weight:

**The file list is parsed out of `target_sources(MduXCore ...)`, not globbed.** A glob over `include/mdux/**` also collects the adapter modules under `render/` and `vulkansc/`, which are permitted to throw and would be reported as violations — and it would miss #116's requirement that generated shader C arrays are not scanned as hand-authored governed source. Those are emitted into the build tree, so deriving from the source list excludes them by construction rather than by an exclusion pattern someone must maintain. A module added to `MduXCore` is covered the moment it is registered, which matters for #15. If the block cannot be parsed the lint exits non-zero rather than scanning nothing.

**Matching happens on the source with comments and string literals removed.** The governed tree is heavily commented and those comments discuss the very constructs banned here — `Kernels.cppm` says "Never `std::fma`", `Draw.cpp` says "a new command is needed", `Runtime.cppm` says "throw away exactly the information the incident report needs". A line-based matcher reports all three. Stripping preserves offsets, so findings still carry a real line and column.

The tree passes with two suppressions, both in `src/ml/Runtime.cpp`, both the `reinterpret_cast` that reads a caller-supplied weights blob as floats — the mechanism ADR-008 chose, guarded by an explicit alignment check on the line above. The marker is per-line and shows up in a diff.

### The ADR reconciliation

`ADR-005` gains a **"What is enforced"** section listing the three layers and naming the file that implements each. It opens with the `import std` / `-fno-exceptions` constraint from #188, and states two consequences plainly: the "usable in `-fno-exceptions` builds" consequence is marked **unmet**, and it is unreachable for any project consuming `MduXCore` through `import std`.

The section ends by saying what is claimed and what is not — governed code contains no throw expression, and *can* still reach libstdc++'s throwing helpers through ordinary `std::string` and `std::vector` use. The risk register gains that item with **"Mitigation: none today"**, because the alternative was to leave a reader with an impression the tree does not support.

`ADR-004` item 2 now names the lint. A new subsection records the zone-placement rule #187 established: a speculative future device consumer does not justify governed placement against a present constraint. The "claim, stated exactly" paragraph gains a no-throw clause and an `-fno-exceptions` exclusion.

### Also here

- `docs/architecture.md`: the trust-zone claim, the `governed` test label, the negative fixture.
- `docs/roadmap.md`: the header table said 7 delivered and 3 waves while the footer said 8 and four; nine are delivered now that #11 closes. The header also still named a commit SHA that `f4e9900` reported replacing with the release tag.
- `.gitignore`: `inputs/` as a directory rule. Five specific sub-paths were listed, leaving `inputs/CMake/` and any new folder under `inputs/Documentation/` untracked but **not ignored** — so `git add -A` would have committed exactly the material #7 and #23 purged from history and ADR-006 forbids.

Closes #116.

Invitation / CLA issue: #184

## Dependency

- **Base branch:** `116-governed-throw-enforcement` (stacked)
- **Predecessor PR:** #188, itself stacked on #187
- **Merge order:** **must not merge until #188 has merged**, which must not merge until #187 has. The ADR text here describes the mechanism #188 adds, so merging it first would document something that does not exist — the failure mode this whole issue is about.

## Shared registries touched

- [x] Generated indexes — none changed, but `generate_ai_reference.py --check` was run to confirm
- [x] None of the others

## Rebase state

- **Rebased onto:** `6b4268e` (predecessor head), on `913fe84`, on `e96dc35` (current `develop`)

## Verification

- [x] `cmake --preset ninja-gcc && cmake --build build-gcc` — clean, 0 errors
- [x] `ctest --test-dir build-gcc` — 438/438 passed
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK (78 files, 0 findings)
- [x] Other:
  - `python3 tools/governed-lint/mdux_governed_lint.py` — OK (29 files checked, 0 findings); the 29 files match the 29 objects the symbol scan reads
  - `python3 -m unittest discover -s tools/governed-lint` — 31 tests, OK
  - end-to-end regression check: a `throw` appended to `src/draw/Draw.cpp` was reported at the correct file, line and column with exit status 1, then reverted
  - `mdux-evidence-lint` OK; `check_schema_type_drift.py` OK; `generate_ai_reference.py --check` up to date
  - confirmed `git status` no longer shows `inputs/`

Not run locally: the MSVC leg. The lint is toolchain-independent Python, so the Windows exposure here is inherited from #188 rather than added.

### A note on the unit tests

They are weighted toward the cases where the lint must stay **quiet**. A source lint an author routes around by rewording documentation is worse than no lint. Three of those fixtures quote block-comment interiors and needed their `/* */` delimiters to mean what they claim — without them the fixture really is code, and it failed correctly.

## Post-merge gate

## Regulatory impact

**This is a compliance-documentation correction as much as a feature.** Two ADRs asserted enforcement that did not exist; both now describe only what CI runs, and both gained explicit statements of what they cannot claim:

- `MduXCore` cannot be built with `-fno-exceptions` (ADR-005, previously listed as an achieved positive consequence).
- Governed code can reach libstdc++'s throwing helpers (ADR-005 risk register, mitigation: none today).

`ADR-004`'s "claim, stated exactly" paragraph — the wording `sdf-documents` and `mdux-regulated-change` both require for compliance contexts — was updated in both directions, and now carries a note that every clause names a CI mechanism and every exclusion names something checked and found not to hold.

Closing #116 and #117 closes epic #11, so no enforcement gap carries into Wave 5.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,11 +2,22 @@
 # Based on C++ Core Guidelines
 # Enforces coding standards from CONTRIBUTING.md
 
+# bugprone-exception-escape is deliberately NOT disabled. It was, until issue #116: ADR-005 opens
+# by citing that suppression as the reason a `noexcept` violation in governed code was invisible to
+# static analysis, and listed re-enabling it as a follow-up gated on MduXCore existing. MduXCore has
+# existed since Wave 1.
+#
+# Scope, stated honestly: this check is *available* rather than *enforced*. The only CI job running
+# clang-tidy is security-analysis.yml, which invokes it with `|| true` against a GCC-generated
+# compile_commands.json that clang-tidy cannot parse for module interface units - both limitations
+# are documented in that workflow. So re-enabling this catches a noexcept violation for a developer
+# running clang-tidy locally, and will catch one in CI when the Clang leg is restored (#48), but it
+# is not what stops one landing today. That is governed.noThrow.symbolScan, which reads the emitted
+# objects and runs on every ctest.
 Checks: >
   -*,
   bugprone-*,
   -bugprone-easily-swappable-parameters,
-  -bugprone-exception-escape,
   cert-*,
   -cert-dcl21-cpp,
   -cert-dcl50-cpp,
--- a/.github/workflows/governed-lint.yml
+++ b/.github/workflows/governed-lint.yml
@@ -1,0 +1,47 @@
+name: Governed Source Lint
+
+on:
+  push:
+    branches: [ master, develop ]
+  pull_request:
+    # Matches the PR's *base*, not its head: a base branch missing from this list means that PR
+    # reports no checks at all, silently - which is worse than reporting failures. `[0-9]+-*` is
+    # the scheme GitHub's "create a branch for this issue" button generates (issue number, dash,
+    # slugified title) and this project's standard - see AGENTS.md "Branch naming". It is also
+    # what gets a stacked PR checked, i.e. one targeting its predecessor rather than develop so a
+    # reviewer sees one issue's diff instead of the cumulative one. `feat/**` predates the
+    # convention and is kept for branches already in flight.
+    branches: [ master, develop, '[0-9]+-*', 'feat/**' ]
+
+permissions:
+  contents: read
+
+jobs:
+  governed-lint:
+    name: Governed Source Lint
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # No step here performs an authenticated Git operation, so the checkout token has no
+      # reason to survive into .git/config where any later build or test step could read it
+      # (OpenSSF Scorecard's Token-Permissions check, zizmor's `artipacked`).
+      with:
+        persist-credentials: false
+
+    # ADR-005's governed-zone source rules (issue #116): no throw/try/catch, no throwing
+    # `.value()`, no raw allocation, no filesystem/console/clock/randomness, no unsafe cast, no
+    # std::fma. Needs no C++ toolchain - the lint parses MduXCore's source list out of
+    # CMakeLists.txt rather than configuring the project - so it runs as its own fast job.
+    #
+    # This is one of two layers. The other, governed.noThrow.symbolScan, reads the emitted objects
+    # and therefore needs a build; it runs inside the toolchain legs as a ctest. Neither subsumes
+    # the other: this sees a std::filesystem call that never throws, and the scan sees a throw
+    # arriving through a std facility the source never spells out.
+    - name: Run mdux-governed-lint
+      env:
+        PYTHONDONTWRITEBYTECODE: "1"
+      run: |
+        python3 -m unittest discover -s tools/governed-lint -p "test_*.py"
+        python3 tools/governed-lint/mdux_governed_lint.py
--- a/.github/workflows/linux-gcc16-build.yml
+++ b/.github/workflows/linux-gcc16-build.yml
@@ -108,6 +108,15 @@ jobs:
     - name: Verify ML determinism (GCC 16)
       run: ctest --preset ninja-gcc -L determinism --output-on-failure
 
+    # Governed-zone no-throw, at the object level (ADR-005, issue #116). Broken out with -V rather
+    # than folded into the full run above, because a passing ctest prints none of a test's output -
+    # and this scan's output is half its purpose. It reports the throw-helper references it
+    # deliberately tolerates, and the ADR justifies tolerating them on the grounds that they stay
+    # "visible and counted". A number nobody can see in a job log is not counted. -V is what makes
+    # the claim true, and is the same reason the pixel step below uses it.
+    - name: Verify governed-zone no-throw (GCC 16)
+      run: ctest --preset ninja-gcc -L governed -V --no-tests=error
+
     # Rendered-truth verification (issue #125), moved here when the GCC 15 leg was retired.
     # Without it a broken ICD turns the only test that renders anything into a silent no-op:
     # offscreen_tests exits 77 with no device and CTest reports Skipped, which does not fail a run.
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -115,6 +115,16 @@ jobs:
       run: ctest --preset ninja-msvc -L determinism
       shell: cmd
 
+    # Governed-zone no-throw (ADR-005, issue #116). Informational on this toolchain, and run
+    # anyway: the MSVC STL inlines its own throw sites, so _CxxThrowException in a governed object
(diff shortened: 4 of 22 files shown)


0xfritz-dev (changes requested):
Request changes — blocked by #188's MSVC failure, plus three findings of its own. I checked out this branch and ran the tool: the lint passes (29 files, 0 findings), its 31 unit tests pass, `--format json` validates against `docs/governance/schemas/diagnostic.schema.json`, and positive findings (GOV001/GOV004) fire with correct line/column. The strip-comments-and-literals approach and the parsed-not-globbed source list are both sound, and the ADR rewrites' honesty (marking the -fno-exceptions consequence *unmet*, 'Mitigation: none today' in the risk register) is exactly right.

1. **Blocker (inherited):** the Windows leg fails on `governed.noThrow.symbolScan` — see the #188 review. ADR-005's 'two layers, both required to pass' and architecture.md must describe whatever MSVC resolution lands there.

2. **Bug in the tool's own CLI:** `check_file` calls `path.relative_to(root)`, so the advertised `[paths...]` interface crashes with a raw `ValueError` traceback for any file outside the repository root (reproduced: `python3 tools/governed-lint/mdux_governed_lint.py /tmp/bad.cpp`). CI never passes paths, so the job is unaffected, but the tool's documented interface should not traceback — use `os.path.relpath` or emit a GOV000-style diagnostic.

3. **Roadmap coordination risk:** the roadmap edit declares '#11 — Done / all eleven closed', which requires both #116 (this stack) *and* #117 (#186) merged. #186 targets `develop` independently of this stack and nothing enforces the order. If this merges before #186, the roadmap claims a closure that has not happened — either merge #186 first or reword.

4. **Rule gaps worth closing before the codes go 'stable once published':** GOV004 misses `delete[]` (`\bdelete\s` requires trailing whitespace) and placement spellings; GOV005 misses `snprintf` and bare C `system(`; GOV009 is a blocklist, so `linux/`, `termios.h`, `io.h` etc. are uncovered. The `delete[]` one is a two-character fix; the rest are acceptable if the blocklist framing stays documented.

The `.gitignore` change to a directory-level `inputs/` rule is a correct tightening (verified nothing under `inputs/` is tracked today).

ambroise-leclerc:
@0xfritz-dev — thank you for checking out the branch and actually running the tool; the schema validation and the line/column spot-check are more than I could have asked for. All four addressed.

**1. Inherited MSVC blocker — fixed in #188** (`_CxxThrowException` demoted to reported under `dumpbin`, per your preferred option). The ADR-005 wording you flagged is corrected here: the layer table now has a **"gates on"** column, and only `mdux-governed-lint` reads `every toolchain`. Details in my reply on #188.

**2. CLI crash — fixed, and it was real.** Reproduced exactly as you described:

```
ValueError: '/tmp/…/bad.cpp' is not in the subpath of '/home/…/MduX'
```

`display_path()` now falls back to `os.path.relpath`, and to the absolute path where even that is inexpressible (different drive on Windows). Two tests cover it, including one asserting `check_file` *reports* rather than crashes for a file outside the root.

Copilot separately suggested resolving relative arguments against `root` instead of `cwd`. I did not take that: resolving against the working directory is what every CLI does and what `mdux-evidence-lint` already does, so `mdux_governed_lint.py Draw.cpp` from inside `src/draw/` should mean the file the shell completed. The crash it was worried about is fixed at the reporting end, which is where it belonged. A comment records the choice so it does not get "fixed" later.

**3. Roadmap coordination — reworded rather than assumed away.** You are right that nothing orders #186 against this stack. Rather than drop the status, the entry now names both PRs and tells the reader what to conclude if #186 is still open:

> If you are reading this and #186 is still open, the "Done" above is ahead of the tree — which is precisely the kind of unearned status `#116` existed to remove, so say so rather than working around it.

A roadmap quietly containing an unearned status *about the issue that exists to remove unearned statuses* would have been a poor way to close it. That said, merging #186 first is still the better outcome and I would suggest it.

**4. Rule gaps — all three closed.** Verified each before fixing:

| Input | Before | After |
|---|---|---|
| `delete[] p;` | **missed** | GOV004 |
| `std::snprintf(...)` | **missed** | GOV005 |
| `system("rm")` | **missed** | GOV005 |

`delete` now matches `\bdelete\s*(?:\[\s*\]|\s)`; the printf family, `getenv` and `system` match with or without a `std::` qualifier. Four new unit tests pin the spellings that were slipping through — 35 tests total.

GOV009 stays a blocklist and now says so in place, with the names you suggested added (`linux/`, `termios.h`, `io.h`, plus others). The comment explains why an allowlist is not worth it here — governed code reaches std through `import std` and carries exactly one `#include` today — and points at ADR-004 item 3's link-graph assertion as the real backstop rather than leaving this rule to carry weight it cannot.

Verified after the changes: 438/438 on GCC 16, lint clean over 29 files, 35 unit tests, JSON envelope still validates, docs lint clean.

0xfritz-dev (approved):

Issue #198, opened 2026-08-11
S9 — CMake integration and the mdux-meduic host tool


ambroise-leclerc:
Part of #15. Blocked by S8; blocks S12.

## Work

- `mdux-meduic`, the host-only compiler executable, split library-and-executable like every other baker so tests drive the calls rather than spawn a process.
- A `mdux_compile_screen()` CMake function registering the compile, following `mdux_bake_artifact()`'s arrangement: build-tree output, an `-update` target that stages into `generated/`, and a test that byte-compares the two.
- Registered in `tools/CMakeLists.txt` in the host-tools zone, deliberately not declared governed.

## Acceptance

- A normal build never writes into the source tree; `mdux-bake-update` remains the only path that does.
- The evidence test appears under the `evidence` label and passes on both toolchain legs.

## Binding constraints from ADR-012 (#190, PR #202)

Three properties of `mdux_bake_artifact()` that this issue has to design around rather than
discover. All verified in `cmake/MduXBake.cmake` on `develop`.

**`KIND` and `ID` must match `^[a-z0-9][a-z0-9-]*$`** — a `FATAL_ERROR` at line 96 enforces it. A
screen's identity in the grammar is CamelCase (`Screen NeuroSense500`), so this issue must decide
and document the mapping from screen name to artifact id, and the compiler must apply it
consistently — the id appears in the directory path, the test name `evidence.screen.<id>`, and the
emitted C++ identifier.

**Every `OUTPUTS` entry is a hard `add_custom_command` OUTPUT** (line 143). A baker that decides not
to write `goldens.json` — for a screen with no safety-critical and no explicitly-positioned nodes —
breaks the build rather than producing a smaller artifact. Either always write the file, empty
array included, or omit it from `OUTPUTS` conditionally. Prefer the former: an always-present file
makes "this screen has no goldens" an assertion rather than an absence.

**`toolVersion` stays the project's semantic version.** ADR-007 decision 5 removed `toolGitSha`
deliberately, and the reasoning binds here: build-dependent provenance in a byte-compared artifact
fails `evidence.screen.<id>` on one leg the moment the legs differ. Do not record the host compiler,
its version, or anything else that varies per machine.

**Locale multiplicity is unsettled** — see #197. Whether `generated/screen/<id>/` holds one package
for all approved locales or one per locale changes this issue's file layout and `OUTPUTS` list.
Settle it jointly with #197 before the first artifact is committed.

[...]

Issue #197, opened 2026-08-11
S8 — Canonical package and C++ emitters


ambroise-leclerc:
Part of #15. Blocked by S5 and S7; blocks S9 and S10.

## Work

- A canonical JSON package for a compiled screen, written through `mdux.evidence.json` so float encoding and key ordering follow ADR-007 rather than a second convention.
- A C++ emitter producing a `constexpr`-constructible screen, in the two forms `cmake/MduXShaderEmit.cmake` established: a `.cppm` module interface and a `.hpp` fallback. This is what lets a device build link no host-tools module and run no parser at startup — the end state #153 is pursuing for ML packages.
- `static_assert(package.validate().has_value())` in the generated source, so a malformed screen is a build error rather than a startup failure.

## Acceptance

- A test compiles both emitted forms in one binary and asserts they describe the same screen, as `shader_spec` does for shaders.
- The schema module is imported by both the device runtime and the host compiler — one definition, not two that can drift.

## Binding constraints from ADR-011/ADR-012 (#190, PR #202)

**One emitter precedent exists, not three.** `mdux_emit_shader_package()` is the tree's only
`.cppm`+`.hpp` emitter. Model and font packages have none, and `EcgClassifierExample` still links
`MduX::MlBakeLib` to parse `package.json` at startup (`examples/CMakeLists.txt:122`) — the state
#153 exists to leave. Follow the shader arrangement; do not assume a second reference exists.

**`constexpr`-compatible types throughout.** The generated source carries
`static_assert(screen.validate().has_value())`, which requires `mdux.medui.schema` to be header-only
and fully `constexpr`. `mdux.ml.schema` is the model — `tests/ml/SchemaTests.cpp:126` already
`static_assert`s a reference package validating at compile time. `TextPackage` is explicitly **not**
the model: it owns `std::string` and `std::vector` and cannot appear in a `static_assert` at all.

**Open question this issue must settle: locale multiplicity.** Baking glyph runs makes *which
locale* a property of the artifact. ADR-012 does not decide whether one compiled screen carries
every approved locale or whether there is one package per locale, and the recipe names locales
plural while `TextPackage` took the one-locale-per-package shape. Settle it here with #198, because
the file layout under `generated/screen/<id>/` depends on the answer.

> **Settled — see the decision comment below.** This paragraph reasons from a premise #203 removed:
> the compiled screen no longer carries glyph runs, so *which locale* is not a property of the
[...]

Issue #195, opened 2026-08-11
S6 — Text-budget validation against every approved locale


ambroise-leclerc:
Part of #15. Blocked by S4 and S5.

## Work

- For every text-bearing component, measure the widest approved translation using the baked font metrics (`mdux.font.schema`) and reject a component whose resolved bounds cannot contain it.
- Validate dynamic text against the restricted charset table from #161, so a format that could escape the charset is a compile error rather than a device-time surprise.

## Acceptance

- The error names the locale, the key, the available width and the required width — enough for an author to act without re-running anything.
- Measurement reuses the committed font package rather than a second metrics path.

## Decision from ADR-011, revised (#190)

Unchanged in substance, and worth restating because the surrounding decision moved: **the widest-
translation check still happens here, at build time**, even though the screen package no longer
carries baked glyph runs.

The compiler reads the per-locale text packages to measure, exactly as it would have to in order to
bake them. What changed is only that the measured runs stay in the text package rather than being
copied into the screen. A component whose resolved bounds cannot contain the widest approved
translation is still a compile error, and that error is still this issue's to produce.

Issue #193, opened 2026-08-11
S4 — Semantic analysis: theme tokens and locale-checked strings


ambroise-leclerc:
Part of #15. Blocked by S3; feeds S5 and S6.

## Work

- Resolve `Theme.Colors.<Token>` against a governed token table. An unknown token is a compile error, not a fallback colour.
- Resolve every `t("STR-KEY")` against the approved text package for **every** approved locale. A key missing in any one locale is an error naming the locale.
- Reject a hardcoded string literal where a component expects text — the rule that makes "no shaping on device" reachable.

## Acceptance

- Text resolution reuses `mdux.text.schema` rather than restating the package model.
- A fixture proves a key present in one locale and absent in another is rejected, and names the locale in the message.

## Decision from ADR-011, revised (#190)

**Validate, do not substitute.** This stage proves every `Theme.Colors.<Token>` exists in the
governed table and every `t("STR-KEY")` is present in **every** approved locale. It then carries the
token and the key into the package **as names**. It does not bake RGBA8 values and it does not bake
glyph runs — the device performs the substitution by a bounded lookup in a governed table, which is
not parsing and is not unbounded.

An earlier revision of ADR-011 required the opposite. It was withdrawn after checking the sibling
project: TrustSC carries `text_key` and `color_token` in `CompiledScreenPackage` and resolves them
on device through `resolve_color_token()` / `ScreenTextLayout::from_screen()`.

The validation work is unchanged and is still all of this issue's job. Only the substitution moves.

PR #204, opened 2026-08-11
Amend ADR-011/012: keep the compiled screen locale-free, matching TrustSC [#203]

### LINKED ISSUE #203: Amend ADR-011/ADR-012: keep the compiled screen locale-free, matching TrustSC
Follow-up to #190, which is closed. The decision it recorded has changed, and the change needs its own record rather than a silent edit.

## What changed

ADR-011 as merged says **"Resolved means literal, not looked up"** — a compiled screen carries RGBA8 values and baked glyph runs — and ADR-012 leaves the locale-multiplicity question open against #197 and #198.

Checking the sibling project before propagating those records to the Wave 5 children found that TrustSC already answers the question, and answers it a third way that neither ADR considered:

- `CompiledScreenPackage` (`crates/trustsc-ui/src/lib.rs:273`) has **no locale field**.
- Nodes carry `text_key` and `color_token` — names, not values.
- Per-locale glyph runs stay in the text package; the join happens on device via `ScreenTextLayout::from_screen(screen, package, locale)`.
- `resolve_color_token()` and `THEME_COLORS` live in the governed `trustsc-ui` crate.

**That is better than the per-locale packaging MduX was about to choose**: adding an approved locale rewrites no screen artifact at all. Per-locale packages would add one artifact per locale and rewrite every screen when a translation changed — and translations change far more often than layouts, so the arrangement makes the frequent change rewrite the stable artifact. In a byte-compared evidence pipeline that is the difference between re-verifying one artifact and re-verifying all of them.

"Resolved means literal" was a **definitional** argument — a runtime holding a token would be "performing the last step of a resolution this ADR places on the build machine" — rather than a consequential one. What the boundary exists to keep off the device is *parsing* and *unbounded work*; a lookup in a fixed governed table is neither.

## Work

- ADR-011: withdraw "resolved means literal", state that a compiled screen is locale-free, and record what the withdrawal does **not** relax — the compiler still validates every token, every key against every approved locale, and every text budget against the widest translation; and the device lookup must be a bounded scan returning a `Result` on miss.
- ADR-012: `package.json` carries validated names, no locale field, no glyph runs, no RGBA8. Close the open locale question.
- Both: the stage table row and the decision-2 wording that still describe literal resolution.

## Why this is a separate issue

The amendment was written and pushed to #190's branch **after PR #202 had already merged**, so it never reached `develop`. Meanwhile #193, #195, #197 and #198 were updated to the new decision. The tree and the issues therefore contradict each other until this lands, which is the failure mode #116 existed to remove — recorded here rather than fixed quietly.

## Acceptance

- `develop`'s ADR-011 and ADR-012 agree with #193, #195, #197, #198 and #199.
- `mdux-docs-lint` clean.

ambroise-leclerc:
## Summary

Re-lands an amendment that was written on #190's branch **after PR #202 had already merged**, so it never reached `develop`. My error — I read `gh pr checks 202`, saw zero failures, and treated a successful `git push` as delivery without checking the commit was on the target branch. @ambroise-leclerc caught it.

Until this merges, `develop`'s ADRs and the Wave 5 child issues **contradict each other**: ADR-011 on `develop` says "Resolved means literal, not looked up", while #193, #195, #197 and #198 were already updated to say a compiled screen is locale-free. That is the failure mode #116 existed to remove, so it is recorded in #203 rather than fixed quietly.

## The decision

Checking TrustSC before propagating #190's records to the Wave 5 children found that the sibling already answers the locale question, a third way neither ADR considered:

- `CompiledScreenPackage` (`crates/trustsc-ui/src/lib.rs:273`) has **no locale field**
- Nodes carry `text_key` and `color_token` — names, not values
- Per-locale glyph runs stay in the text package; the join happens on device via `ScreenTextLayout::from_screen(screen, package, locale)`
- `resolve_color_token()` and `THEME_COLORS` are in the governed `trustsc-ui` crate

**Adding an approved locale therefore rewrites no screen artifact.** The per-locale packaging MduX was about to choose would add one artifact per locale and rewrite every screen whenever a translation changed — and translations change far more often than layouts, so it makes the frequent change rewrite the stable artifact. In a byte-compared pipeline that is the difference between re-verifying one artifact and re-verifying all of them.

"Resolved means literal" is withdrawn. It was mine, and it was a **definitional** argument — a runtime holding a token would be "performing the last step of a resolution this ADR places on the build machine" — rather than a consequential one. What this boundary keeps off the device is *parsing* and *unbounded work*; a lookup in a fixed governed table is neither.

Two things the withdrawal does **not** relax, written into the ADR so it is not read as wider than it is:

1. The compiler still validates every colour token, every text key against every approved locale, and every text budget against the widest approved translation. That work does not move to the device because the substitution does.
2. The device lookup must be a bounded scan returning a `Result` on miss — never an allocation, never a throw. `mdux-governed-lint` and `governed.noThrow.symbolScan` enforce that once the module is in `MduXCore`.

@devjean-75 — this partly reverts your decision-2 correction in `799dbc8` and my resolved-values wording in `28b615c`. You were right that the two records contradicted each other; the fix goes the other way now, and the parity programme's direction is MduX moving toward the sibling.

Closes #203.

Invitation / CLA issue: #184

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked. Supersedes the un-merged tail of #202's branch.
- **Merge order:** mergeable now, and **the sooner the better** — `develop` and #193/#195/#197/#198 disagree until it lands.

## Shared registries touched

- [x] None of the above — two ADR files.

## Rebase state

- **Rebased onto:** `9c4f826` (current `develop`, #202's merge commit)
- **Conflicts resolved as a union:** no conflicts. Cherry-picked cleanly, since the lost commit was written on top of the same tree that merged.

## Verification

Documentation only.

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run, nothing in the build graph changed
- [ ] `ctest --test-dir build-gcc` — not run, same reason
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK (80 files, 0 findings)
- [x] Other: verified on the branch, against `develop` this time rather than against a merged PR's stale checks — ADR-011 now contains "A compiled screen is locale-free", "Resolved means literal" survives only as the withdrawal note, and ADR-012's open locale question is closed.

## Post-merge gate

- [ ] That run is green.

## Regulatory impact

**Amends two accepted ADRs**, so `mdux-regulated-change` applies. The change moves one substitution step from build time to a bounded device-side table lookup. It does **not** move any validation: locale coverage, unknown colour tokens and text-budget overflow all remain compile errors, which is the IEC 62304 §5.5 argument ADR-011 makes for the boundary in the first place.

The traceability consequence is unchanged — `requirement:` is still resolved at compile time and still appears in the committed package.

## Summary by CodeRabbit

* **Documentation**
  * Clarified architecture guidance for deterministic, locale-free compiled screen packages.
  * Documented runtime lookup of validated theme tokens and text keys.
  * Confirmed locale-specific glyph data remains in text packages, so adding locales does not require rebuilding screen layouts.
  * Reaffirmed build-time validation, locale coverage checks, text budgets, rendering requirements, and non-allocating, non-throwing runtime behavior.

### DIFF
--- a/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
+++ b/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
@@ -20,8 +20,11 @@ storage sized from that budget and never grown. A compiler that could not comput
 build time would leave that type with no way to be used as designed.
 
 **Text shaping is already forbidden on device.** ADR-010 settled that static text bakes to
-positioned glyph runs and dynamic text is restricted to a validated charset. A `.medui` runtime
-that resolved `t("STR-KEY")` itself would reopen the question ADR-010 closed.
+positioned glyph runs and dynamic text is restricted to a validated charset. A `.medui` runtime that
+*shaped* `t("STR-KEY")` — chose glyphs, applied substitutions, measured — would reopen the question
+ADR-010 closed. Looking a key up to find the run ADR-010 already baked does not: the shaping
+happened on the build machine either way, and the lookup is a bounded scan of a fixed table. That
+distinction is what lets the Decision below keep keys in the package rather than baked runs.
 
 **The governed zone is a hostile place to put a parser.** ADR-004 keeps `MduXCore` on `std` alone,
 and issue #116 made ADR-005's no-throwing rule mechanical — `mdux-governed-lint` rejects
@@ -40,7 +43,8 @@ consumes. That choice determines whether a device build links a parser at all.
 ## Medical Device Considerations
 
 ### IEC 62304 implications (software lifecycle)
-- **Verification moves left.** A screen whose layout, text budgets and colour tokens are resolved at
+- **Verification moves left.** A screen whose layout is solved, whose text budgets are checked and
+  whose colour tokens and text keys are *validated* at
   build time can be rejected by a build, and a build failure is an artifact of the development
   process rather than a field observation. A screen resolved at runtime can only be verified by
   running it, on every input that changes it.
@@ -78,9 +82,10 @@ device runtime performs none of them, and turns that layout into vertices.**
 | Emit golden references (see rule below) | build machine | host tools |
 | **Build a `DrawList` from a compiled screen** | **device** | **governed** |
 
-*Layout*, not *geometry*: the compiler produces rectangles, colours and budgets, and the runtime
-produces vertices and indices from them. ADR-012 decision 2 explains why the vertex data cannot be
-baked. Both records use the word this way, and neither uses "geometry" for the compiled form.
+*Layout*, not *geometry*: the compiler produces rectangles, budgets and the validated token and key
+*names* each node draws with, and the runtime produces vertices and indices from them. ADR-012
+decision 2 explains why the vertex data cannot be baked. Both records use the word this way, and
+neither uses "geometry" for the compiled form.
 
 **Resolution is validated at build time; the last substitution is a bounded table lookup.** A
 `Theme.Colors.<Token>` and a `t("STR-KEY")` are both *checked* by the compiler — an unknown token
@@ -103,11 +108,17 @@ by looking each node's `text_key` up for the running locale. TrustSC does exactl
 `ScreenTextLayout::from_screen(screen, package, locale)`.
 
 The consequence that decides it: **adding an approved locale touches no screen artifact at all.**
-The alternative — one screen package per locale, carrying baked runs — would add a file and a
-`evidence.screen.<id>` per locale, and would re-baked every screen when a translation changed.
+The alternative — one screen package per locale, carrying baked runs — would add a file and an
+`evidence.screen.<id>` per locale, and would rebake every screen when a translation changed.
 Translations change far more often than layouts, and coupling the two makes the frequent change
-rewrite the stable artifact. It would also oblige a single-package variant to size `DrawBudget` for
-the widest locale on every device, including devices shipping only one.
+rewrite the stable artifact.
+
+**The cost of choosing this way, stated as a cost.** A locale-free screen does not know which locale
+it will be paired with, so its `DrawBudget` has to cover the widest approved translation — and a
+device shipping only `en-US` carries the German or Finnish ceiling in its buffers. Per-locale
+packages would size each budget to its own locale, and that is their one real advantage; it is
+given up deliberately. Buffer headroom is cheap and recoverable, while a translation update that
+rewrites every screen's digest is neither.
 
 Two things this does *not* relax. The compiler still validates every key against every approved
 locale, and still validates text budgets against the widest approved translation (#195) — that work
--- a/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
+++ b/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
@@ -71,7 +71,7 @@ device runtime performs none of them, and turns that layout into vertices.**
 | Stage | Where | Zone |
 |---|---|---|
 | Lex, parse, build AST | build machine | host tools |
-| Resolve theme tokens to literal RGBA8, and text keys to baked glyph runs | build machine | host tools |
+| Validate theme tokens and text keys against every approved locale | build machine | host tools |
 | Solve layout, flatten `Row`, compute absolute rectangles | build machine | host tools |
 | Validate text budgets against every approved locale | build machine | host tools |
 | Compute the `DrawBudget` | build machine | host tools |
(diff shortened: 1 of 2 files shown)


docs/adr/ADR-011-deterministic-medui-compile-boundary.md
```diff
@@ -82,19 +82,38 @@ device runtime performs none of them, and turns that layout into vertices.**
...
+parity programme's direction is MduX moving toward it.
+
+**A compiled screen is locale-free.** It carries no locale field and no glyph runs. Per-locale
+glyph runs stay in the text package where ADR-010 already put them, and the two are joined on device
+by looking each node's `text_key` up for the running locale. TrustSC does exactly this in
+`ScreenTextLayout::from_screen(screen, package, locale)`.
+
+The consequence that decides it: **adding an approved locale touches no screen artifact at all.**
+The alternative — one screen package per locale, carrying baked runs — would add a file and a
+`evidence.screen.<id>` per locale, and would re-baked every screen when a translation changed.
```
0xfritz-dev: Two small grammar slips in this line: "would re-baked every screen" -> "would re-bake", and "a `evidence.screen.<id>`" -> "an `evidence.screen.<id>`".

0xfritz-dev (dismissed):
Review verified against the trees on both sides.

**Scope and process** — exactly one commit on the claimed base `9c4f826`, touching only the two ADR files. All 14 checks green, including docs-lint and the governed-lint job this stack added. The amendment preserves the withdrawn wording as a quoted withdrawal note rather than deleting it, which keeps the decision trail intact; index status stays `Accepted` correctly since this amends rather than supersedes.

**Every cited mechanism verified in-tree on `develop`:** `include/mdux/text/Schema.cppm:145` is indeed the single-`locale` field with the "runtime reads no others" comment; `governed.noThrow.symbolScan` is registered (tests/CMakeLists.txt:771); `mdux_bake_artifact()` and the slug regex exist; `tests/ml/SchemaTests.cpp:126` carries the `static_assert` precedent.

**Every TrustSC parity claim verified against `main`:** `CompiledScreenPackage` (lib.rs:273) has no locale field; nodes carry `text_key`/`color_token`; `resolve_color_token()`/`THEME_COLORS` are in `trustsc-ui`; `ScreenTextLayout::from_screen(screen, package, locale)` exists (crates/trustsc/src/screen_text.rs:39). Both referenced commits (`799dbc8`, `28b615c`) exist on #202's branch, and #197 already records the locale question as SETTLED locale-free — so the contradiction with `develop` the PR describes is real, and this is the fix for it.

**Regulatory dimension (mdux-regulated-change):** amendment to accepted ADRs, handled proportionately — validation stays build-side, the IEC 62304 left-shift argument is preserved, the device gains only a bounded lookup with a `Result` on miss, and no normative text is reproduced. The two "does not relax" clauses are the right guard against over-reading the withdrawal.

**One observation for a follow-up, non-blocking:** three lines this PR does not touch still carry the pre-amendment wording and now sit in slight tension with the Decision — ADR-011:24 ("a `.medui` runtime that resolved `t("STR-KEY")` itself would reopen the question ADR-010 closed" — under the new decision the runtime *does* resolve keys on device; the distinction that keeps both true is shaping vs. lookup), ADR-011:43 ("colour tokens are resolved at build time" — now *validated*), and ADR-012:233 ("why text is resolved before the device sees it" — glyph runs are; the key-to-run join isn't). They can ride along with #197/#198 rather than holding this merge. One inline typo note on the new text.

ambroise-leclerc:
All findings addressed. One was a genuine error in the reasoning, not the wording.

## The widest-locale argument was inverted

@coderabbitai — correct, and thank you for catching it. The paragraph rejecting per-locale packages ended by saying the alternative *"would oblige a single-package variant to size `DrawBudget` for the widest locale"* — and a single-package variant is exactly what this ADR chose. The sentence argued **against its own decision while appearing to argue for it**, and the "two things this does not relax" paragraph twelve lines below already said the chosen design validates budgets against the widest approved translation. The document contradicted itself within one screen of text.

Corrected in the honest direction rather than deleted:

> **The cost of choosing this way, stated as a cost.** A locale-free screen does not know which locale it will be paired with, so its `DrawBudget` has to cover the widest approved translation — and a device shipping only `en-US` carries the German or Finnish ceiling in its buffers. Per-locale packages would size each budget to its own locale, and that is their one real advantage; it is given up deliberately. Buffer headroom is cheap and recoverable, while a translation update that rewrites every screen's digest is neither.

That restores the ADR index's own rule — a record listing only benefits documents an advertisement rather than a decision.

## The three pre-amendment lines

@0xfritz-dev flagged these as non-blocking and rideable with #197/#198. I fixed them here instead: they are contradictions *inside the same document*, so deferring them means shipping a record that disagrees with itself, and the edit is three sentences.

The ADR-010 one is the load-bearing one, and your framing of the reconciliation was right — **shaping versus lookup**. The paragraph now draws it where a reader meets it first, since the Decision below depends on it:

> A `.medui` runtime that *shaped* `t("STR-KEY")` — chose glyphs, applied substitutions, measured — would reopen the question ADR-010 closed. Looking a key up to find the run ADR-010 already baked does not: the shaping happened on the build machine either way, and the lookup is a bounded scan of a fixed table.

The other two: the IEC 62304 bullet now says colour tokens are *validated* at build time rather than resolved, and ADR-012's reference list says ADR-010 explains why text is *shaped* before the device sees it.

## Copilot

Both taken — "resolved nodes" → "compiled nodes" in the `package.json` row, since "resolved" was the exact word the amendment stopped meaning; and "rectangles, colours and budgets" now reads "rectangles, budgets and the validated token and key *names*", which was the same reintroduction one paragraph earlier. Typo `re-baked` → `rebake` fixed.

Documentation lint clean over 80 files.

camille-martin-paris (approved):

Issue #192, opened 2026-08-11
S3 — Lexer, parser, AST and the fixture corpus


ambroise-leclerc:
Part of #15. Blocked by S2 (diagnostics); blocks S4, S5, S6.

## Work

- Host-only lexer, parser and AST for the grammar in the `medui-authoring` skill: `Screen`, `layout:`, `surface:`, components with fields, `@safety_critical(...)`, `Row { }` (single level, no nesting), `t("KEY")`, `Theme.Colors.<Token>`, `Npx`/`Fill` sizes, `position: Xpx, Ypx`.
- Deliberately absent, and rejected with a diagnostic rather than ignored: loops, conditionals, recursion, runtime scripting, nested `Row`.
- A fixture corpus of both accepted and rejected inputs. Every rejection fixture asserts on the **code**, so a fixture that starts failing for a new reason is visible.

## Acceptance

- The parser lives in the host-tools zone and is never linked into `MduXCore` or `MduX`.
- Every diagnostic carries a real line **and column**; a fixture asserts the position, not just the code.
- Hand-written parser, no generator dependency — consistent with `mdux.tools.truetype` and the TOML reader.

## Binding constraint from ADR-011 (#190, PR #202)

**Nested `Row` is rejected for a solver reason, not a budget one.** The parser still rejects it, but
do not defend the rejection with the budget argument that covers loops, conditionals, recursion and
scripting: a nested `Row`'s depth is fully visible to the compiler and the primitive count stays
exact either way. It is excluded to keep the layout solver (#194) a single flattening pass over a
known structure. Getting this right matters for the diagnostic's `fixHint`, which should say what
the author must do rather than cite a constraint that does not apply.

PR #207, opened 2026-08-12
Diagnostics and the MDX-E code registry [#191]

### LINKED ISSUE #191: S2 — Diagnostics and the MDX-E code registry
Part of #15. Blocked by S1's boundary ADR; blocks S3.

## Work

- A stable diagnostic code registry for the compiler, in the shared envelope from #118 (`file`, `line`, `column`, `code`, `severity`, `message`, `fixHint` — `docs/governance/schemas/diagnostic.schema.json`). Codes are stable once published: an agent keys off them and a reworded message must not break it.
- Prefix and numbering convention, consistent with the existing bakers (`SHB###`, `TXT###`, `MLB###`) — `MDX-E` per the epic, reconciled against those if they disagree.
- A registry that a test can enumerate, so an unregistered code and a registered-but-unused code are both detectable.

## Acceptance

- Every code carries a stable id, a severity, and a fix hint that names the action rather than restating the problem.
- A test asserts the registry has no duplicate ids and no gaps that would suggest a removed code.
- The JSON form validates against `diagnostic.schema.json`.

ambroise-leclerc:
## Summary

Wave 5 S2. #118 fixed the *envelope* diagnostics travel in; this fixes the *set of codes* that may appear in one. It is the compiler's contract with an agent, so it lands before the parser that will start emitting them (#192).

### Why a registry rather than string literals

Every other tool in this repository spells its codes at the call site — `constexpr std::string_view recipeUnparsed = "TXT001";` in `tools/text/TextBake.cpp`, and in `mdux-docs-lint` not even that: `"MDX-D011"` appears inline at **nineteen** call sites with no table at all.

That works until someone asks a question about the *set*. Is this number already taken? Is this code still emitted anywhere? Does every code carry a fix hint? None of those is answerable by grep with any confidence. All three are answerable by construction here.

The shape is a `constexpr` table keyed by an enum, so call sites name `Code::UnknownColorToken` and never `"MDX-E030"`:

- **An unregistered code cannot be emitted** — `diagnose()` takes no string overload.
- **A registered-but-unused code is detectable** — the enum is walkable, and `DiagnosticsTests.cpp` walks it.
- **A typo is a compile error**, not a diagnostic nobody can grep for.

### `MDX-E` joins an existing family

`mdux-docs-lint` publishes `MDX-D###`, and the diagnostic schema names `MDX-D011` in its own description, so `MDX-` + a letter + three digits is established. The bakers' three-letter codes (`SHB`, `SHE`, `TXT`, `MLB`, `EVL`, `GOV`) are an older second convention — unchanged, and not needing to change, since the schema's `code` pattern admits both and `code` is opaque to a consumer keying off it.

**The `E` is for the language, not for "error."** Severity is a separate field and an `MDX-E` code may be a warning; worth stating before someone infers the other reading and files a bug about `MDX-E###` at severity `warning`.

### Twenty-two codes, none invented

Each is mandated by ADR-011, ADR-012 or the `medui-authoring` skill. Nothing anticipates a check that has not been argued for — a code invented ahead of its rule has its meaning decided by whoever implements it first, which is the opposite of a stable identifier.

Numbers are assigned in blocks with gaps inside each, so a new grammar rule lands beside the grammar rules rather than at the end of the table.

`retired()` is empty and present from the start, so the first retirement has somewhere to go rather than prompting a decision under time pressure. A retired number is never reused: a consumer that pinned behaviour to `MDX-E017` must not silently start matching a different rule.

Closes #191.

Invitation / CLA issue: #184

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked. #202 and #204 (the ADRs it implements) and #206 (the CI pin) are all merged.
- **Merge order:** mergeable now. It is the predecessor of #192, which needs these codes to report anything.

## Shared registries touched

- [x] `tools/CMakeLists.txt` — new `MduXMeduiLib` host-tools target
- [x] `tests/CMakeLists.txt` — new `medui_tools_spec` suite
- [x] `FILE_SET CXX_MODULES` lists — one new module interface on the new target
- [x] None of the others

## Rebase state

- **Rebased onto:** `2c02cc6` (current `develop`, with the gcc pin)

## Verification

Built and run on GCC 16.1.0 — deliberately **after** #206 landed. An earlier run against red `develop` would have been measuring the container bump, not this change.

- [x] `cmake --preset ninja-gcc && cmake --build build-gcc` — clean, 0 errors
- [x] `ctest --test-dir build-gcc` — **446/446** (438 + 8 new)
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK (80 files)
- [x] Other:
  - `mdux_verify_trust_zones()` — 2 governed targets clean; `MduXMeduiLib` is deliberately **not** among them, so a governed target linking it is a configure-time error
  - `mdux-governed-lint` — OK (29 files), unchanged, since medui is host-tools
  - `mdux-evidence-lint` — OK (40 files)
  - all 22 identifiers checked against the schema's `code` pattern: 22 unique, 0 failing

### The completeness test was checked for teeth

A test that only ever runs against a correct registry passes identically when it is working and when it is checking nothing. Adding an enumerator with no table row fails it on **both** assertions:

```
- enumerator 22 has its own row (DiagnosticsTests.cpp:89)
- table has 23 rows, one per enumerator (has 22) (DiagnosticsTests.cpp:93)
```

Reverted after checking; the suite is back to 8/8.

## Post-merge gate

- [ ] That run is green.

## Regulatory impact

**Diagnostic stability is a traceability property.** These codes are what an agent or a reviewer keys off to identify a class of compile failure, and ADR-011 makes several of them the mechanism by which a safety-relevant rule is enforced — `MDX-E070` for a `@safety_critical` node with no `requirement:`, `MDX-E032` for a text key missing from an approved locale, `MDX-E053` for dynamic text that could escape the restricted charset.

Fixing the set before the parser exists is deliberate: a code that changes meaning after publication silently invalidates whatever was keyed to it, and in a regulated context that includes review records. Nothing here asserts any check is *implemented* — the codes exist, the checks that emit them are #192 onward.

## Summary by CodeRabbit

- **New Features**
  - Added a comprehensive diagnostic system for `.medui` compiler errors.
  - Diagnostics include stable identifiers, severity levels, summaries, suggested fixes, and source locations.
  - Added consistent diagnostic lookup and reporting, with clear handling for unrecognized codes.
  - Added the initial `.medui` compiler tools foundation for future parsing and code generation capabilities.

- **Documentation**
  - Documented the `.medui` compiler tools and diagnostic registry in the architecture overview.

- **Tests**
  - Added coverage for diagnostic completeness, formatting, ordering, metadata, source positions, and reporting behavior.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -20,7 +20,7 @@ boundary between them enforced at configure time rather than by review:
 |---|---|---|---|
 | **Governed** | `MduXCore`, `MduX_warnings` | `std` only | never throws ([ADR-005](adr/ADR-005-error-handling-and-exceptions-policy.md)) |
 | **Adapter** | `MduX` | governed + Vulkan | throws where Vulkan makes it unavoidable |
-| **Host tools** | `MduXToolsCommon`, `MduXShaderBakeLib`, `MduXMlBakeLib`, `MduXTextBakeLib` | anything | may throw freely; never linked into a device target |
+| **Host tools** | `MduXToolsCommon`, `MduXShaderBakeLib`, `MduXMlBakeLib`, `MduXTextBakeLib`, `MduXMeduiLib` | anything | may throw freely; never linked into a device target |
 
 `mdux_verify_trust_zones()` in [`cmake/MduXTrustZones.cmake`](../cmake/MduXTrustZones.cmake) walks
 the full link graph of every declared-governed target at the end of configure and fails on a
@@ -98,6 +98,7 @@ performs no checking and confers no compliance.
 | `MduXShaderBakeLib` | `tools/shader/` | `mdux-shaderbake`, `mdux-shaderemit` |
 | `MduXMlBakeLib` | `tools/ml/` | `mdux-mlbake` |
 | `MduXTextBakeLib` | `tools/text/` | `mdux-textbake`; also hosts `mdux.tools.truetype` (the host-only glyf parser with cmap/hmtx, #158), `mdux.tools.atlaspacker` (the shelf packer, #160) and `mdux.text.raster` (the glyph rasteriser, #159) |
+| `MduXMeduiLib` | `tools/medui/` | the `.medui` compiler (#15); at S2 only the `MDX-E` diagnostic registry (#191) — the parser (#192), emitters (#197) and `mdux-meduic` (#198) land on this same target |
 
 Host tools parse untrusted input, so they are deliberately outside the governed zone. They are
 never linked into `MduXCore` or `MduX` and are absent from the install/export set.
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -653,6 +653,38 @@ endif()
 
 mdux_discover_tests(text_tools_spec)
 
+# .medui compiler host-tools suite (issue #191)
+#
+# Links MduX::MeduiLib. At S2 the library is the diagnostic code registry and nothing else, so this
+# suite is entirely registry invariants - completeness, uniqueness, identifier shape, ordering
+# against the retired list, and the envelope diagnose() produces. Those are the questions a
+# registry exists to answer mechanically; without them the module is a table of string literals
+# with extra steps.
+add_executable(medui_tools_spec
+    medui/MeduiToolsSpecMain.cpp
+    medui/DiagnosticsTests.cpp
+)
+
+target_link_libraries(medui_tools_spec PRIVATE MduX::MeduiLib speclab::speclab)
+target_include_directories(medui_tools_spec PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+set_target_properties(medui_tools_spec
+    PROPERTIES
+        CXX_STANDARD 23
+        CXX_STANDARD_REQUIRED ON
+        CXX_EXTENSIONS OFF
+)
+
+if(TARGET __CMAKE::CXX23)
+    set_target_properties(medui_tools_spec PROPERTIES CXX_MODULE_STD ON)
+endif()
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    target_compile_options(medui_tools_spec PRIVATE /experimental:module /std:c++latest)
+endif()
+
+mdux_discover_tests(medui_tools_spec)
+
 # ML host-tools suite (issue #60)
 #
 # Links MduX::MlBakeLib, the host-tools-zone target. Deliberately separate from ml_spec, which
--- a/tests/medui/DiagnosticsTests.cpp
+++ b/tests/medui/DiagnosticsTests.cpp
@@ -1,0 +1,249 @@
+/**
+ * @file DiagnosticsTests.cpp
+ * @brief BDD scenarios for the `.medui` diagnostic code registry (issue #191).
+ *
+ * @compliance ADR-004 Trust zones in C++ (host tools zone)
+ * @compliance ADR-011 The deterministic `.medui` compile boundary
+ *
+ * The registry's whole value is that questions about the *set* of codes are answerable
+ * mechanically. These are the questions, as tests. A registry nobody enumerates is a table of
+ * string literals with extra steps, so if these scenarios were deleted the module below would be
+ * worth deleting with them.
+ */
+
+import std;
+import speclab;
+import mdux.tools.cli;
+import mdux.tools.medui.diagnostics;
+
+#include "../framework/SpecLabBridge.hpp"
+
+namespace {
+
+namespace md = mdux::tools::medui;
+namespace cli = mdux::tools::cli;
+
+/// Every enumerator, so a scenario can walk the enum rather than the table it is checking against.
+///
+/// Written out rather than derived: C++ has no reflection over an enum, and a `Count` sentinel
+/// cast in a loop would assume contiguity this enum does not promise. The duplication is the point
+/// - adding an enumerator without adding it here fails `every code is registered`, which is the
+/// nudge that also gets it a table row.
+constexpr std::array<md::Code, 22> allCodes{
+    md::Code::RecipeUnreadable,
+    md::Code::RecipeUnparsed,
+    md::Code::RecipeMissingMember,
+    md::Code::SourceUnreadable,
+    md::Code::SourceNotUtf8,
+    md::Code::UnexpectedToken,
+    md::Code::UnknownComponent,
+    md::Code::MissingRequiredField,
+    md::Code::UnknownField,
+    md::Code::DuplicateNodeId,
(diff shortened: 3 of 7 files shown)


0xfritz-dev (commented):
## Review

Reviewed the full diff against ADR-011, ADR-012, the `medui-authoring` skill, `mdux.tools.cli`, the diagnostic schema and the existing host-tools conventions. The design holds up; I pushed four corrections in 9d0cb59, none touching behaviour.

### What I verified

- **All 22 codes are grounded.** Each maps to a rule in ADR-011 (token/key validation, forbidden constructs, nested `Row`, budgets, `requirement:`), ADR-012 (recipe/source handling, `cv_check` via the goldens predicate) or the skill (closed component dictionary, hardcoded strings, restricted charset, `@safety_critical`). Numbers sit inside their documented blocks; enum order, table order and `allCodes` agree.
- **The completeness test has the teeth claimed.** A missing row fails the round-trip assertion; a row with no enumerator fails the size assertion (`registry().size() == allCodes.size()`). Duplication of the enum in `allCodes` is unavoidable without reflection and is documented as the nudge.
- **Trust-zone claim is mechanically true.** `cmake/MduXTrustZones.cmake:83-89` fatal-errors when a governed target reaches a non-governed one, so a governed target linking `MduXMeduiLib` fails at configure, exactly as the PR states. Governed count is 2 (`MduXCore`, `MduX_warnings`).
- **Schema fit.** `diagnostic.schema.json`'s `code` pattern `^([A-Za-z0-9][A-Za-z0-9-]*)?$` admits `MDX-E###`; the test's narrower 8-char shape is strictly additional coverage, as its comment claims. The six baker prefixes (`SHB`, `SHE`, `TXT`, `MLB`, `EVL`, `GOV`) all exist in the tree.
- **CMake.** Both new blocks mirror `MduXTextBakeLib`/`text_tools_spec` line for line; the `evidence-unit` tag matches the other tool suites.
- CI is green on both toolchain legs at the reviewed commit.

### Corrections pushed (9d0cb59)

1. `Diagnostics.cppm` said `"MDX-D011"` appears inline at nineteen call sites. It appears at **two**; nineteen is the call-site count for the whole `MDX-D` family (`grep -c 'MDX-D[0-9]' tools/docs-lint/mdux_docs_lint.py` → 19). Reworded.
2. The header paired `Code::UnknownColorToken` with `"MDX-E004"`; its registered id is `MDX-E030`, which `diagnoseCarriesRegisteredIdentity` itself asserts. Fixed.
3. `registry()`'s doc described the table as a "function-local `static constexpr` array"; it is a namespace-scope `constexpr` array in the module implementation unit (`Diagnostics.cpp:29`). The one-table property holds either way — the doc now describes the mechanism that exists.
4. `docs/architecture.md` enumerates the host-tools targets in two tables and is held to the build by AGENTS.md § 2, so `MduXMeduiLib` gets a row in both.

`mdux-docs-lint` re-run after the edits: OK (80 files).

### Honest limitation

This environment has GCC 13 and no CMake/Ninja, so I could not build or run the suite locally; verification here is static review plus the green CI runs. The correction commit is comments-and-docs only, but its own CI run should be green before merge.

### One observation, no action required

`info()` returning `table.front()` for an unregistered enumerator is defensible only while the completeness test exists — the comment at `Diagnostics.cpp:118-121` says exactly that. If `DiagnosticsTests.cpp` is ever deleted, that fallback silently becomes the wrong answer the comment denies; the test file's own header already argues the module and the tests stand or fall together, which is the right coupling.

ambroise-leclerc:
@0xfritz-dev — thank you, and your four corrections in `9d0cb59` are all accurate. I verified the load-bearing one rather than taking it on trust, since I had repeated it in the PR description too:

```
grep -c '"MDX-D011"' tools/docs-lint/mdux_docs_lint.py   -> 2
grep -oE '"MDX-D[0-9]{3}"' ... | wc -l                    -> 19
```

So `MDX-D011` appears **twice**, and nineteen is the whole `MDX-D` family across thirteen distinct codes. I generalised from a family count to a single code and stated it as fact in a header comment whose entire subject is the cost of not being able to answer questions about a code set — which is a poor place to get a code count wrong. Your rewording says what is true and keeps the point intact.

The other three are the same shape and all mine:

- The header paired `Code::UnknownColorToken` with `"MDX-E004"` while the registry and `diagnoseCarriesRegisteredIdentity` both say `MDX-E030`. A doc comment contradicting the test two files away.
- `registry()`'s doc described a function-local `static constexpr` array. I wrote a namespace-scope one. Describing a mechanism I did not implement is exactly the failure #116 existed to remove, in a file I wrote after arguing that at length.
- `docs/architecture.md` — I added a host-tools target and did not add it to the document AGENTS.md § 2 holds to the build. Both tables, correctly.

## The `info()` fallback — changed, not defended

@Copilot raised it as actionable and you as a no-action observation. Two independent objections is enough, and on re-reading, the defence was wrong rather than merely cautious.

The completeness test covers every **enumerator**. `Code` has a fixed underlying type, so `static_cast<Code>(200)` is a well-formed value of the type that the test cannot reach — a cast, a value read from data, or an uninitialised member would all have mapped silently to `MDX-E000`, *"the recipe file could not be opened"*, with that code's severity and fix hint attached to an unrelated failure.

A registry whose whole purpose is that a code means one thing cannot have a path that quietly assigns the wrong one. Your framing — "defensible only while the completeness test exists" — was the polite version; the sharper version is that it was a mechanism that looked safe because a test covered the case *next to* the one that mattered.

Split as @Copilot suggested: `tryInfo()` returns a pointer and stays `noexcept`, `info()` throws `std::logic_error` naming the numeric value, `id()` follows. Throwing is available because this is the host-tools zone.

Two knock-on changes worth noting:

- The completeness scenario now uses `tryInfo()`, so a missing row still fails with the **enumerator named** rather than throwing out of the loop and losing which one. It also gained a second assertion that the row names its code back.
- A ninth scenario covers the new behaviour and records *why the case is reachable at all*, so the next reader does not restore the fallback on the grounds that the scenario above it already covers everything.

## On your limitation note

Appreciated, and it is the right thing to state. For what it is worth from this side: I built and ran on GCC 16.1.0 after #206 landed — **447/447** now — and re-ran `mdux-docs-lint` (80 files) and `mdux-governed-lint` (29 files) after your commit and mine.

One consequence of #206 worth flagging if you retry locally: `develop` is pinned to `gcc:16.1.0` because the floating `gcc:16` tag moved to 16.2.0 and GCC 16.2 cannot read the module BMIs the SpecLab sources produce. GCC 13 will not build this tree at all, so the static review was the right call rather than a compromise.

0xfritz-dev (approved):

Issue #16, opened 2026-07-26
Epic: Rendered-truth verification


ambroise-leclerc:
TrustSC's ADR-016 inspired the rendered-truth goal, but MduX exposes it through a dedicated host
tool rather than through an example application: render a compiled screen offscreen, check that
golden-scoped content appears where the compiled screen says it will and that every localized text
run appears inside its node across every approved locale, then emit that as an evidence artifact.

```sh
mdux-verify-ui --screen=generated/screen/<id> --locales=all
```

## Current status (2026-09-02)

**S3 is next.** #15 closed on 23 August at 12/12, and #13 and #14 were already complete. #196
supplied the golden-reference sidecar consumed here. ADR-014 (#251) fixes the verification boundary
and obligation model, and `mdux.verify` (#252) implements the four governed checks. #253 is therefore
unblocked; #254 and #255 remain sequentially blocked behind it.

Two things landed after the epic opened: #242 draws a `Label`, so `InkContainment` and
`LocalizedTextPresence` have real ink to check rather than waiting on #17, and #244 binds a screen
to the text packages it was compiled against by digest, so "every approved locale" is a property of
the manifest rather than a caller-supplied list.

The driver can be implemented and tested on purpose-built screens now, but the currently committed
`endoscope-monitor` screen cannot pass its golden obligations yet: its golden `NumericDisplay` and
`SignalTrace` nodes remain deferred by the runtime. ADR-014 records #17 as the planned route to make
those nodes render. That integration dependency is not permission to remove the goldens, change the
verified screen merely to obtain a pass, or weaken the obligation model.

## Driver boundary

`mdux-verify-ui` is a host-only tool under `tools/verify/`, split into a directly testable
`MduXVerifyUiLib` and a thin executable entry point, following the existing host-tool pattern. It
imports the governed `mdux.verify` checks and uses the adapter/offscreen rendering path; it is not
part of `MduXCore`, is not installed or exported, and must not be linked into a governed target.

No example owns this behavior. In particular, `MedicalUiExample` remains the no-device/no-window
demonstration it is today; it does not gain a `--verify-ui` mode or a Vulkan-device dependency.

`--screen` names the committed `generated/screen/<id>/` bundle. The tool reads `package.json` and
`goldens.json` there and resolves the governed theme, font and approved per-locale text artifacts
from the package identities and digests. The caller chooses locations, not expectations: it cannot
substitute bounds, tints, runs, fonts, goldens or a narrower locale set.

[...]

PR #208, opened 2026-08-13
Pin SpecLab to v0.1.0, its first tagged release

ambroise-leclerc:
## Summary

SpecLab was pinned to a bare commit because it had no tag. It has one now — and `v0.1.0` is the commit that pins **its own** toolchain to `gcc:16.1.0`, so this bump is what makes MduX's #206 pin a shared position rather than a local workaround.

Upstream hit the same defect independently ([SpecLab #15](https://github.com/ambroise-leclerc/SpecLab/issues/15)): the floating `gcc:16` tag moved to 16.2.0, corrupting the module BMIs for `speclab.core.testsuite` and `speclab.runners.testrunner` — the exact translation units that turned this repository's `develop` red. Same diagnosis, same fix, reached separately. A working 16.2.x is tracked upstream as SpecLab #17, and MduX's follow-up waits on it.

### On tag versus SHA

The comment records the commit the tag resolves to (`ea7f2e363`). A git tag is mutable in a way a bare SHA is not, and that mutability was the **one** property the old pin had over this one — recording the SHA beside the tag keeps a moved tag detectable rather than silently different. The reproducibility argument in that comment is ADR-007's, unchanged.

## The SOUP register needed more than a version bump

Bumping a SOUP entry means re-reading it, and several claims had gone stale — one by my own hand:

| Field | Was | Now |
|---|---|---|
| `usage` | "seven test executables", naming seven | **eleven**, named |
| `version` | commit, "no tagged release upstream" | `v0.1.0`, with its commit |
| `license` | "See upstream repository" | EUPL-1.2 + commercial terms |
| `support_model` | explained the bare-commit pin by the absence of a tag | says what changed, without erasing the earlier state |
| `risk_controls` | "pinned to an immutable commit" | a tag is not immutable; states the actual control |
| `known_anomaly_tracking` | "none open" | records the GCC 16.2 BMI anomaly, shared with upstream |

The `usage` count is the one worth calling out: Wave 4 added `font_spec`, `text_spec` and `text_tools_spec`, and **#191 added `medui_tools_spec` last night** — so I made a compliance artifact staler yesterday and am fixing it today. The "nine further executables use MduXTest" claim beside it is still exactly right; I counted both rather than assuming either.

Invitation / CLA issue: #184

## Dependency

- **Base branch:** `develop`
- **Merge order:** mergeable now. Independent of Wave 5; #192 does not need it.

## Shared registries touched

- [x] `tests/CMakeLists.txt` — the `CPMAddPackage` pin only, no target or source changes
- [x] Schemas / governance documents — `docs/governance/soup-register.toml`
- [x] None of the others

## Rebase state

- **Rebased onto:** `b75a6ac` (current `develop`, with #207 merged)

## Verification

Run from a **clean build directory**, since the delta touches `TestSuite.cppm` and `TestRunner.cppm` and a stale BMI would have hidden exactly the problem worth looking for.

- [x] `cmake -S . -B <fresh>` — `CPM: Adding package SpecLab@0.1.0 (v0.1.0)`
- [x] `cmake --build <fresh>` — clean; no `Bad file data`, which is the failure signature this pin exists to avoid
- [x] `ctest --test-dir <fresh>` — **447/447**
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK (80 files); unit tests OK
- [x] Other: `soup-register.toml` re-parsed with `tomllib` after editing; the eleven/nine executable split counted from `tests/CMakeLists.txt` rather than from the old text

Toolchain: GCC 16.1.0, matching the container `develop` pins.

## Post-merge gate

- [ ] That run is green.

## Regulatory impact

**A SOUP register update**, so `mdux-regulated-change` applies. IEC 62304 §5.3.4 and §8 are the relevant clauses, and the entry is now accurate on the three things a register is for: what version is in the build, what licence it carries, and what is known to be wrong with it.

The last is the substantive change. `known_anomaly_tracking` said no anomalies were open; one is — GCC 16.2.0 corrupts this dependency's module BMIs — and it is recorded with both projects' issue numbers and the shared mitigation. A register that omits a known defect because the defect is in the toolchain rather than the dependency is not doing its job.

No change to trust zones, to what is shipped, or to any claim about a standard. SpecLab remains test-only, absent from the install/export set, and unreachable from any governed target.

## Upstream note, not blocking

SpecLab's squash commit for #16 is titled *"Pin toolchain to GCC 16.2 (#16)"* while the PR title, body and the change itself all say **16.1.0**. The commit message is the odd one out. Worth correcting upstream so `git log` does not contradict the tag's contents.

## Summary by CodeRabbit

* **Documentation**
  * Updated SpecLab integration guidance to reference the `v0.1.0` release and verified commit.
  * Added EUPL-1.2 licensing details and alternative commercial terms.
  * Expanded documented test-framework coverage to eleven test executables.
  * Documented release-tag changes, reproducibility requirements, GCC 16.2.0 compatibility risks, and related issue tracking.

* **Chores**
  * Pinned the integration to a verified commit for consistent builds.
  * Added coordinated tag and commit update instructions.


ambroise-leclerc:
## The tag moved while this PR was open — repinned to the commit

SpecLab's release was published after I opened this, and `v0.1.0` was repointed in the process:

| | Commit |
|---|---|
| When this PR pinned `v0.1.0` | `ea7f2e363` |
| Now | `37b51fcec` |

The added commit is [SpecLab #18](https://github.com/ambroise-leclerc/SpecLab/pull/18), "Fix release package staging path" — it touches `.github/workflows/release.yml` and makes SpecLab's project version a CMake cache variable. **No library source changed**, so this is not a correctness fix; it is a correction to the pin's *shape*.

### What that says about pinning to a tag

The previous commit argued that a tag is readable and that recording its SHA alongside keeps a move detectable. That held exactly as written — the move was detected within the hour, precisely because the SHA was written down.

But detecting a moved reference after the fact is a weaker property than not having one. ADR-007's reproducibility argument does not distinguish between a branch and a retagged release: both can change the source under a frozen MduX commit. The original bare-commit pin had this right, and my tag was a step back dressed up as an upgrade — I traded an immutable reference for readability and it cost me inside an hour.

Now: **SHA is the pin, tag name is a comment above it.** To bump, resolve the tag and change both, in a reviewable diff.

```cmake
    # SpecLab v0.1.0
    GIT_TAG 37b51fcec
```

### SOUP register

`support_model` now records the *observed supplier practice*, not just the policy — a maintainer who moves a published release tag is a fact about the supply chain, and carrying that is what a register is for. `risk_controls` goes back to claiming an immutable commit, which it can again mean literally.

### Re-verified

Second clean build directory, since the pinned source changed under me:

- `CPM: Adding package SpecLab@37 (37b51fcec)`
- fetched tree confirmed at `37b51fcec` via `git rev-parse` in `_deps/speclab-src`
- **447/447**, no `Bad file data`
- documentation lint OK (80 files)

### Upstream, non-blocking

Two things worth raising with the SpecLab maintainer, neither affecting this PR:

1. **A published release tag was moved.** Consumers who pinned `v0.1.0` in the intervening window get different source from consumers who pin it now, with no version change to signal it. Cutting `v0.1.1` would have preserved that contract; this is why MduX now pins the commit.
2. **The squash commit for SpecLab #16 is titled "Pin toolchain to GCC 16.2"** while its PR title, body and content all say **16.1.0**. `git log` currently contradicts what the commit does.

ambroise-leclerc:
All three findings addressed. Two were claims the entry could not support, and one of them the entry **disproved two lines further down**.

## "A SpecLab regression cannot silence the whole suite"

@coderabbitai — right, and this is the one I should have caught while writing it. The nine `MduXTest` executables give independent coverage only *after* a successful configure and build. A SpecLab configure, compile or module-BMI failure blocks everything before any executable runs.

The proof is in the same entry: the GCC 16.2.0 anomaly recorded two lines below did exactly that — `develop` went fully red and **no test ran at all**. So the register asserted a safety property in one field and documented its violation in the next. The half that was wrong is the half that reads as reassurance, which is the worse half to get wrong.

Now scoped to a *runtime* regression, with the build-blocking case named and pointed at the anomaly that demonstrates it.

## "Its risk is false confidence in test results, not patient harm"

Also too absolute, and this one matters more than it looks. Test-only linkage establishes that **no code from this dependency executes on a device** — worth saying, and all it establishes.

A test framework that reports a pass for failing code lets a real defect through verification, and release decisions rest on verification. The risk is to the integrity of verification rather than to runtime behaviour, which is a different thing from *no risk*. A SOUP register is the wrong place to blur that distinction, and "not patient harm" blurred it.

## Ambiguous upstream issue numbers

@Copilot — correct, and worse than the comment suggests. In this repository a bare `#17` is the **Content components** epic and `#15` is the **`.medui` compiler** epic. So "SpecLab #15" collides with the very work this dependency bump serves. Both files now use `ambroise-leclerc/SpecLab#N` throughout, with a note saying why.

@Copilot's second point, on `"or fails outright"`: reworded to say what makes it true — the fetch fails because the commit is unreachable, and there is no third outcome where it silently builds something else. That claim is accurate for a commit pin in a way it would not have been for the tag pin this PR started with.

## Re-verified

447/447, documentation lint clean (80 files), register re-parsed as TOML, and `grep -nE "SpecLab #[0-9]"` returns nothing in either file.

---

Worth noting what this review round did: the SOUP entry went from three sentences that sounded reassuring to three that are checkable, and the correction came from the register's own recorded anomaly rather than from anything new. That is the register working — but it only worked because someone read it against itself.

camille-martin-paris (changes requested):
Two documentation inconsistencies remain in the final revision and should be corrected before merge:

1. `docs/governance/soup-register.toml:97` — The pinned SpecLab `LICENSING.md` says commercial terms **may** be available and are granted only through a separate written agreement. “available from the project steward” states their availability unconditionally. Please mirror the upstream qualification (for example, “alternative commercial terms may be available under a separate written agreement”) and cite the pinned commit rather than the mutable `v0.1.0` tag.

2. `tests/CMakeLists.txt:448` — “Pinned to a tag, not to a branch” contradicts both the implementation and the explanation immediately below: `GIT_TAG` contains commit SHA `37b51fcec`, not a tag. Please change this to “Pinned to a commit, not to a tag or branch” (or equivalent).

I otherwise found the final pin, SOUP boundary/risk corrections, qualified upstream issue references, and verification evidence consistent. All substantive checks are green.

ambroise-leclerc:
@camille-martin-paris — both correct, both fixed. Each is documentation contradicting either upstream or itself, which is the failure mode this PR spent two commits correcting elsewhere in the same entry.

## 1. The licence field overstated what is available

Upstream `LICENSING.md` at the pinned commit says:

> Alternative commercial terms **may** be available from the project steward … A commercial licence is granted **only through a separate written agreement**. The presence of this notice does not itself grant commercial-licence rights.

The register said *"with alternative commercial terms available from the project steward"* — dropping both the conditional and the agreement requirement. This is the field someone reads to answer what the project is entitled to rely on, and an unconditional reading answers that wrongly. It now mirrors upstream's qualification and records why.

Your second point on that line was the sharper one: the field cited `LICENSING.md` **at `v0.1.0`** — a mutable tag, for provenance, inside the entry whose entire subject is that the tag moved. Self-defeating. Now cites commit `37b51fcec`, as every other provenance reference in the entry does.

## 2. The pin comment opened by describing a pin it no longer is

`tests/CMakeLists.txt:448` still read *"Pinned to a tag, not to a branch"* — written in this PR's first commit, left behind by the second, and contradicted by both the `GIT_TAG` below it (which holds a SHA) and by its own next paragraph explaining why a tag was rejected.

Now: *"Pinned to a commit, not to a tag or a branch."*

## Worth naming

I introduced that contradiction **while correcting a different one**, and it survived my own re-read, three CI runs, and two bot review rounds. The mechanical checks in this repository do not read prose — `mdux-docs-lint` validates links and citation shape, not whether a comment agrees with the code beneath it — so nothing here would ever have caught it. That is the same gap this PR's earlier commits addressed in the SOUP entry, and it needed a human read both times.

## Re-verified

447/447, documentation lint clean (80 files), register re-parsed as TOML, and `grep -c "at v0.1.0"` in the register now returns 0.

camille-martin-paris (approved):

PR #211, opened 2026-08-15
Semantic analysis: theme tokens and locale-checked strings

### LINKED ISSUE #193: S4 — Semantic analysis: theme tokens and locale-checked strings

ambroise-leclerc:
## Summary

Adds the host-tools `mdux.tools.medui.semantic` stage: the closed component/field dictionary, field value-domain checks, governed theme-token lookup, hardcoded-text rejection, and `t("KEY")` validation across every approved `mdux.text.schema` package. Analysis leaves the AST unresolved and accumulates stable `MEDUI-E###` diagnostics with source positions.

The shared contract was extended first in [Compliatory/MedUI#10](https://github.com/Compliatory/MedUI/pull/10) and [Compliatory/MedUI#11](https://github.com/Compliatory/MedUI/pull/11). This branch pins released contract version `0.1.0` at merge commit `d5136a8518bd499760ecff2aad215d3721329f20`, claims `syntax` and `semantics`, executes its portable conformance cases, and mechanically checks that MduX's component dictionary exactly matches the pinned table.

Closes #193

Invitation / CLA issue: not applicable — this is a maintainer-authored change.

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [x] `tools/CMakeLists.txt` — host-tool targets and their sources
- [ ] None of the above

## Rebase state

- **Rebased onto:** `d4f969c877fe27cdf51a0c45a1d85fdad91cca85`

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run locally; this macOS host has no supported C++23-modules toolchain, CI covers GCC 16 and MSVC
- [ ] `ctest --test-dir build-gcc` — not run locally; CI covers the full suite
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 80 files, 0 findings
- [x] Other: MedUI `python3 tools/validate.py` at the pinned commit — 14 cases, 23 diagnostics
- [x] Other: `python3 tools/docs-lint/generate_ai_reference.py --check`
- [x] Other: `python3 tools/docs-lint/check_schema_type_drift.py` — 7 schemas, 7 bindings, 2 closed vocabularies
- [x] Other: `git diff --check`

The docs-lint unittest suite was also run. It has four existing macOS temporary-link-test failures (`MDX-D012` instead of the expected fixture outcomes); the repository-wide lint itself passes.

## Post-merge gate

- **Post-merge `develop` run:** merge commit `f2c4654305242e47f7cbdcd9fb7222d1423fcdbc` — [GCC/CTest](https://github.com/ambroise-leclerc/MduX/actions/runs/31930195316), [MSVC/CTest](https://github.com/ambroise-leclerc/MduX/actions/runs/31930195274), and [sanitizers](https://github.com/ambroise-leclerc/MduX/actions/runs/31930195359); all `develop` workflows green
- [x] That run is green.

## Regulatory impact

Safety-relevant build-time acceptance behavior changes for authored UI: known fields now fail closed when their syntactic value has the wrong semantic domain, and shared conformance inputs cannot be silently ignored. Affected evidence and traceability artifacts are the semantic unit/conformance tests, `medui-conformance.toml`, ADR links, architecture/status documentation, and the authoring skill. Existing diagnostic identities and ADR decisions are reused; no certification or production-readiness claim is introduced. Maintainer/domain review completed before merge.

## Summary by CodeRabbit

- **New Features**
  - Added semantic validation for MedUI screens, including component fields, theme tokens, localized text, nested content, and value types.
  - Added support for image references and standalone numeric values.
  - Added clearer diagnostics for invalid field values, unknown names, missing locale keys, and malformed inputs.

- **Bug Fixes**
  - Improved dotted-path handling and pixel-value parsing.
  - Prevented invalid pixel syntax from being silently treated as plain numbers.

- **Documentation**
  - Updated compiler status, architecture guidance, conformance details, and implementation roadmap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Derived from the targets and tests that build on `develop`.
 | Risk management, QMS, lifecycle *code* | **Not started** | no `mdux::risk`, `mdux::qms` or `mdux::lifecycle` exists |
 | **Not started** | | |
 | Text and glyph rendering | Planned | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) |
-| `.medui` compiler | Planned | [#15](https://github.com/ambroise-leclerc/MduX/issues/15) |
+| `.medui` compiler | Partial | parser and semantic name validation implemented; [#15](https://github.com/ambroise-leclerc/MduX/issues/15) tracks the remaining stages |
 | Content components (`SignalTrace`, `StatusIndicator`, …) | Planned | [#17](https://github.com/ambroise-leclerc/MduX/issues/17) |
 
 The HTML/CSS path that earlier revisions described was **deleted** by
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ Derived from the targets and tests that build on `develop`.
 | **Host tools** (never linked into a device target) | | |
 | `mdux-shaderbake`, `mdux-shaderemit` | Implemented | SPIR-V reflection, byte-verified packages, generated C++ |
 | `mdux-mlbake` | Implemented | safetensors import, golden generation, byte-verified model packages |
+| `MduXMeduiLib` | Partial | `.medui` parsing and semantic validation; layout, packaging, and emission remain ([#15](https://github.com/ambroise-leclerc/MduX/issues/15)) |
 | `mdux-docs-lint`, `mdux-evidence-lint` | Implemented | run in CI |
 | **Regulatory material** | | |
 | Standards corpus under `docs/` | Documentation only | five clause-structured references with generated indexes and schemas |
@@ -151,7 +152,6 @@ Derived from the targets and tests that build on `develop`.
 | Risk management, QMS, lifecycle *code* | **Not started** | no `mdux::risk`, `mdux::qms` or `mdux::lifecycle` exists |
 | **Not started** | | |
 | Text and glyph rendering | Planned | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) |
-| `.medui` compiler | Partial | parser and semantic name validation implemented; [#15](https://github.com/ambroise-leclerc/MduX/issues/15) tracks the remaining stages |
 | Content components (`SignalTrace`, `StatusIndicator`, …) | Planned | [#17](https://github.com/ambroise-leclerc/MduX/issues/17) |
 
 The HTML/CSS path that earlier revisions described was **deleted** by
--- a/tests/medui/SharedConformanceTests.cpp
+++ b/tests/medui/SharedConformanceTests.cpp
@@ -19,41 +19,43 @@
  *   and must be backed by a case that really executed. Claiming a phase with no adapter fails
  *   here instead of passing quietly.
  *
- * Only the parse phase is observable: `md::parse` answers acceptance, codes and positions.
- * Claiming `semantics`, `layout` or `safety` additionally needs the resolver (#193), the solver
- * (#194) and the golden pass (#196), plus a convention for the `inputs` member the case schema
- * declares and no case yet uses — so a claim to any of them is rejected below rather than
- * silently evaluated by a parser that cannot see those phases.
+ * Syntax and semantics are observable: `md::parse` answers syntax acceptance, and `md::analyze`
+ * checks the portable theme-token and per-locale key views in semantic case inputs. Claiming
+ * `layout` or `safety` still needs the solver (#194) and golden pass (#196), so a claim to either
+ * is rejected below rather than silently evaluated by a stage that cannot see it.
  */
 
 import std;
 import speclab;
 import mdux.core.result;
 import mdux.evidence.json;
+import mdux.text.schema;
 import mdux.tools.cli;
 import mdux.tools.medui.diagnostics;
 import mdux.tools.medui.parser;
+import mdux.tools.medui.semantic;
 import mdux.tools.toml;
 
 #include "../framework/SpecLabBridge.hpp"
 
 namespace {
 
-namespace md = mdux::tools::medui;
-namespace cli = mdux::tools::cli;
+namespace md   = mdux::tools::medui;
+namespace cli  = mdux::tools::cli;
 namespace json = mdux::evidence::json;
 namespace toml = mdux::tools::toml;
 
 /// Phases this file can genuinely observe. Anything else in `capabilities` is a claim with no
 /// adapter behind it.
-constexpr std::array<std::string_view, 1> runnableCapabilities{"syntax"};
+constexpr std::array<std::string_view, 2> runnableCapabilities{"syntax", "semantics"};
 
-[[noreturn]] void fail(std::string message,
-                       std::source_location where = std::source_location::current()) {
+[[noreturn]] void fail(std::string message, std::source_location where = std::source_location::current()) {
     throw speclab::core::AssertionFailure(std::move(message), where);
 }
 
-[[nodiscard]] const char* env(const char* name) { return std::getenv(name); }
+[[nodiscard]] const char* env(const char* name) {
+    return std::getenv(name);
+}
 
 [[nodiscard]] bool isSet(const char* name) {
     const char* value = env(name);
@@ -62,7 +64,9 @@ constexpr std::array<std::string_view, 1> runnableCapabilities{"syntax"};
 
 [[nodiscard]] std::string readFile(const std::filesystem::path& path) {
     std::ifstream in{path, std::ios::binary};
-    if (!in) { fail(std::format("could not open {}", path.generic_string())); }
+    if (!in) {
+        fail(std::format("could not open {}", path.generic_string()));
+    }
     std::ostringstream buffer;
     buffer << in.rdbuf();
     return buffer.str();
@@ -175,32 +179,29 @@ constexpr std::array<std::string_view, 1> runnableCapabilities{"syntax"};
 enum class Positions : std::uint8_t { Full, LineOnly, None };
 
 struct Manifest {
-    std::string commit;
+    std::string              commit;
     std::vector<std::string> capabilities;
-    Positions positions{Positions::Full};
+    Positions                positions{Positions::Full};
 };
 
 /// `governance/versioning.md` makes an unknown key an error rather than something to ignore, so a
 /// misspelled `capabilties` fails here instead of silently claiming nothing.
-constexpr std::array<std::string_view, 5> knownManifestKeys{"repository", "version", "commit",
-                                                            "capabilities", "positions"};
+constexpr std::array<std::string_view, 5> knownManifestKeys{"repository", "version", "commit", "capabilities", "positions"};
 
 [[nodiscard]] Manifest manifest() {
-    const std::filesystem::path path =
-        std::filesystem::path{MDUX_REPO_ROOT} / "medui-conformance.toml";
-    const std::string text = readFile(path);
-    const toml::Document document = toml::parse(text);
-    const toml::Table& root = document.root();
+    const std::filesystem::path path     = std::filesystem::path{MDUX_REPO_ROOT} / "medui-conformance.toml";
+    const std::string           text     = readFile(path);
+    const toml::Document        document = toml::parse(text);
+    const toml::Table&          root     = document.root();
 
-    Manifest result{.commit = root.require("commit").asString(),
-                    .capabilities = root.require("capabilities").asStringArray(),
-                    .positions = Positions::Full};
+    Manifest result{.commit = root.require("commit").asString(), .capabilities = root.require("capabilities").asStringArray(), .positions = Positions::Full};
 
     for (const auto& entry : root.entries()) {
         if (std::ranges::find(knownManifestKeys, entry.first) == knownManifestKeys.end()) {
             fail(std::format("{}: unknown key '{}'; the consumer manifest rejects keys it does "
                              "not define",
-                             path.generic_string(), entry.first));
+                             path.generic_string(),
+                             entry.first));
         }
     }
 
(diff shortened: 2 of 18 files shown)


tools/medui/Semantic.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+namespace {
+
+struct ComponentRule {
+    std::string_view name;
+    std::string_view required;
+    std::string_view allowed;
+};
+
+constexpr std::array<ComponentRule, 11> componentRules{
```
ambroise-leclerc, line 101: Two things about this table.

**It is correct and nothing checks that it stays correct.** I diffed all eleven rows against `spec/component-model.md` at the pinned `f966667`, optional `colors` and `charset` included — it is a faithful transcription. But if upstream adds a component, renames a field, or moves one between required and optional, no pinned case need change and this suite stays green while the compiler rejects screens the contract accepts. `SharedConformanceTests` already holds the checkout in `MEDUI_CONFORMANCE_DIR` and `spec/component-model.md` is a fixed-shape markdown table: parse it there and assert set equality against this array, in the same run that verifies the commit SHA.

**The space-separated encoding has a sharp edge.** A stray double space in `required` yields an empty word, and line 158 then reports ``component 'Row' requires field ''``. `containsWord` absorbs the same empty word silently on the `allowed` side, so the two halves disagree about what it means. A `std::array<std::string_view, N>` per rule costs nothing and cannot do this.

tools/medui/Semantic.cpp
```diff
@@ -0,0 +1,198 @@
...
+        if (value.kind == ast::ValueKind::List) {
+            for (const std::shared_ptr<ast::Value>& element : value.list) {
+                if (element != nullptr) {
+                    analyzeColorValue(*element);
+                }
+            }
+            return;
+        }
+        if (value.kind != ast::ValueKind::ColorToken) {
+            report(Code::UnexpectedToken, value.position, "this colour field requires Theme.Colors.<Token>");
```
ambroise-leclerc: `MEDUI-E010` is phase `syntax` in `spec/diagnostics.md` at the pinned commit, which also says a code's meaning, severity **and phase** are stable. This emits it from semantics — the same applies at line 108.

There is also a double-report. For `color: Theme.Color.Title` (`Color`, not `Colors`), `Parser.cpp:509-514` has already emitted `MEDUI-E010`, and `parse()` returns the screen regardless (`Parser.cpp:580`), so `runCase` runs the analyzer on it and this line fires a second `MEDUI-E010` at a different column. `runCase`'s "every emitted diagnostic is expected" check would fail against any upstream case pinning one.

The parser declines to error there on the grounds that "#193 owns what a field may hold"; #193 now also needs to know the parser already spoke.

tools/medui/Semantic.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+            for (const ast::Field& field : node.fields) {
+                if (!containsWord(rule->allowed, field.name)) {
+                    report(Code::UnknownField, field.namePosition, std::format("field '{}' is not defined for component '{}'", field.name, node.component));
+                    continue;
+                }
+                if (field.value == nullptr) {
+                    continue;
+                }
+                if (isTextField(node.component, field.name)) {
```
ambroise-leclerc: Value kinds are inspected only for four component/field pairs and three colour names, so every other field accepts any `ValueKind`. `width: "120px"` (a `String`, per `Parser.cpp:410`), `id: t("STR-TITLE")` and `spacing: Theme.Colors.Header` all compile today.

`Ast.cppm:65` assigns that rule to this stage: `String` is "a literal, which most fields must not accept (#193)". Either implement it or move the sentence to whichever issue does — a comment specifying behaviour the code does not have is the failure mode this tree keeps repeating, and here the comment is the correct half.

tests/medui/SharedConformanceTests.cpp
```diff
@@ -277,60 +281,86 @@ struct Case {
...
     // `case.schema.json` binds the phase to the directory holding the case, so a case filed under
     // the wrong phase is a contract error rather than something to route around.
     const std::string directoryPhase = path.parent_path().parent_path().filename().string();
     if (result.phase != directoryPhase) {
-        fail(std::format("{}: declares phase '{}' but sits under '{}'", path.generic_string(),
-                         result.phase, directoryPhase));
+        fail(std::format("{}: declares phase '{}' but sits under '{}'", path.generic_string(), result.phase, directoryPhase));
+    }
+
+    if (const json::Value* inputs = document->find("inputs")) {
```
ambroise-leclerc, line 339: `case.schema.json` is `additionalProperties: false` at every level, including inside `inputs`, but this reads `themeTokens` and `textPackages` and silently ignores anything else.

`manifest()` above makes the opposite choice for the manifest, with an argument that applies verbatim here: "governance/versioning.md makes an unknown key an error rather than something to ignore." `spec/component-model.md` already says unknown image IDs, templates, charsets and formats fail compilation, so a future `inputs.imageIds` is a plausible next addition — and MduX would read the case, ignore the new input, and keep passing it.

README.md
```diff
@@ -151,7 +151,7 @@ Derived from the targets and tests that build on `develop`.
 | Risk management, QMS, lifecycle *code* | **Not started** | no `mdux::risk`, `mdux::qms` or `mdux::lifecycle` exists |
 | **Not started** | | |
 | Text and glyph rendering | Planned | [#14](https://github.com/ambroise-leclerc/MduX/issues/14) |
-| `.medui` compiler | Planned | [#15](https://github.com/ambroise-leclerc/MduX/issues/15) |
+| `.medui` compiler | Partial | parser and semantic name validation implemented; [#15](https://github.com/ambroise-leclerc/MduX/issues/15) tracks the remaining stages |
```
ambroise-leclerc: This row is now `Partial` but still sits under the `| **Not started** | | |` section divider a few lines above. It belongs in the **Host tools** block — which also has no `MduXMeduiLib` row at all, even though `docs/architecture.md` gained one in this PR.

ambroise-leclerc (commented):
## Review

I read the analyzer against the **pinned** contract at `f966667` (`spec/component-model.md`, `spec/diagnostics.md`, `MEDUI-DEC-005`, `schemas/case.schema.json`) rather than against the in-repo skill, and hand-traced all eight `conformance/semantics/*` cases through `runCase`. **All eight pass**, including the two the case set added — `missing-approved-locale` (one E032, en-US present / fr-FR empty) and `unknown-text-key` (one E031, no E032 cascade). The `componentRules` table is a faithful transcription of the upstream table, all eleven rows, optional `colors` and `charset` included. The `validate, do not substitute` boundary is respected: nothing is written back into the AST.

Six things below I think should change before merge, then four smaller ones. The first two are latent — no pinned case reaches them today, so CI will be green either way.

### 1. `MEDUI-E010` is emitted from the semantics stage, and the pinned contract fixes each code's phase

`spec/diagnostics.md` at `f966667`: *"a code's meaning, severity, **and phase** are stable"*, and its table assigns `MEDUI-E010` phase `syntax`. `Semantic.cpp:108` and `Semantic.cpp:140` both emit it from semantics, for a value that is syntactically well-formed and semantically wrong. This PR is what turns that into a conformance question, because it is what adds `"semantics"` to `capabilities`.

There is no registered code for "this field's value has the wrong shape", and `Diagnostics.cppm` is explicit that nothing may be invented ahead of its rule — so I think the honest resolution is upstream (a `MEDUI-E0##` for value shape) rather than local. In the meantime, borrowing a syntax-phase code is the one option that makes the manifest claim untrue.

### 2. A mistyped colour path produces two `MEDUI-E010`s

```
color: Theme.Color.Title;   // Color, not Colors
```

`Parser.cpp:509-514` already reports `MEDUI-E010` ("'Theme.Color.Title' is not a Theme.Colors token"), and `parse()` returns the screen anyway (`Parser.cpp:580`), so `runCase` (`SharedConformanceTests.cpp:455`) hands that AST to the analyzer, which sees `ValueKind::Identifier` and reports `MEDUI-E010` again at a different column. Two diagnostics for one typo — and `runCase`'s *"every emitted diagnostic is expected"* assertion would fail against any upstream case that pins one. The parser's comment says it declines to error there because *"#193 owns what a field may hold"*; #193 now also needs to know the parser already spoke.

### 3. Value kinds go unchecked outside text and colour fields — `width: "120px"` compiles

`Ast.cppm:65` states the contract for this stage: `String` is *"a literal, which most fields must not accept (#193)"*. `Semantic.cpp` inspects `isTextField` (four component/field pairs) and `isColorField` (three names); every other field accepts any `ValueKind`. All of these are currently accepted:

```
Label {
    id: t("STR-TITLE");        // TextKey where an identifier belongs
    width: "120px";            // String -> Parser.cpp:410, never inspected
    height: 24px;
    text: t("STR-TITLE");
    color: Theme.Colors.Title;
}
Row { id: r; height: 32px; spacing: Theme.Colors.Header; }
```

That is the AST comment describing behaviour the code does not have — the class this tree keeps producing, and the comment is the correct specification here. Either implement the rule or move the sentence in `Ast.cppm` to whichever issue does.

### 4. The dictionary is a hand transcription with nothing that detects drift

`componentRules` (`Semantic.cpp:28`) is correct against `f966667` — I diffed all eleven rows by hand. Nothing in the suite will notice when it stops being correct. If upstream adds a component, renames a field, or moves one between required and optional, **no pinned case need change**, and MduX stays green while rejecting screens the contract accepts. This is `identifierFor()` again: a parity claim asserted on one side only.

It is cheap to close, and the machinery is already here. `SharedConformanceTests` has the pinned checkout in `MEDUI_CONFORMANCE_DIR`; `spec/component-model.md` is a markdown table with a fixed shape. Parse it there and assert set equality with `componentRules` — then the transcription is checked against the thing it transcribes, in the same run that verifies the commit SHA.

### 5. `readCase` ignores unknown members inside `inputs`; the case schema forbids them

`manifest()` rejects an unknown manifest key on an argument worth keeping (`SharedConformanceTests.cpp:187-206`): *"governance/versioning.md makes an unknown key an error rather than something to ignore."* `readCase` applies no such rule, although `case.schema.json` is `additionalProperties: false` at every level. `spec/component-model.md` already says unknown image IDs, templates, charsets and formats fail compilation, so a future `inputs.imageIds` is a plausible next addition — and MduX would read the case, silently ignore the new input, and keep passing it. Same argument, same file, one level down.

### 6. Space-separated field lists let a typo invent a field named `''`

`{"Row", "id height", "id height spacing background"}`. A stray double space in `required` yields an empty word, and `Semantic.cpp:158` reports `component 'Row' requires field ''`. `containsWord` absorbs it silently on the `allowed` side, so the two halves of the table disagree about what an empty word means. For the table that *is* the normative dictionary, a `std::array<std::string_view, N>` per rule costs nothing and cannot do this.

---

### Smaller

**7. README: "Partial" under the "Not started" heading.** `README.md:154` — the row now reads `Partial` while sitting under `| **Not started** | | |`. It belongs in the **Host tools** block, which also still has no `MduXMeduiLib` row at all.

**8. The skill's component table is now a strict subset of what is enforced.** `SKILL.md:76-87` lists required fields only, and the same file tells an author to *"review a `.medui` file by hand against ... the component table above"*. The compiler now accepts `position` on ten components, `colors` on `StatusIndicator`, `charset` and `requirement` on `TextInput`. An Optional column would make the table match the dictionary the analyzer enforces.

**9. `version = "0.1.0-candidate"` is stale.** The pinned commit's `VERSION` is `0.1.0` (it was at `c8cc45e` too, so this predates the PR). Informational per `consumer-manifest.schema.json` and never read by `manifest()`, but the PR is editing the two lines beneath it.

**10. The reformat buries the change.** `SharedConformanceTests.cpp` shows 251/185 lines, still 150/84 ignoring whitespace, for roughly forty lines of new adapter. The new sources also sit at `ColumnLimit: 160` while the existing `tools/medui/*.cpp` wrap near 100 and CI enforces neither, so the directory now carries two widths. Not worth redoing here; worth deciding once for the directory.

---

On the verification note in the description: given the GCC floor, CI-only evidence is the right call and matches how the rest of this epic has landed. What I would add is that a green run does not currently distinguish "the dictionary is right" from "the dictionary is a copy nobody compares" — which is exactly what #4 fixes.

AM-L (dismissed):

tools/medui/Semantic.cpp
ambroise-leclerc: Addressed in fdfd243. The dictionary now uses typed FieldRule arrays, and shared conformance parses the pinned component table and asserts exact component, field, and requiredness set equality.

tools/medui/Semantic.cpp
ambroise-leclerc: Addressed in fdfd243 after adding MEDUI-E033 to the shared contract in Compliatory/MedUI#11. Non-colour dotted paths remain syntactically valid identifiers, and the semantic stage emits one E033; regression and shared conformance cases cover the phase and duplicate-report behavior.

tools/medui/Semantic.cpp
ambroise-leclerc: Addressed in fdfd243. Every known field now carries a FieldDomain and is checked, including identifiers, sizes, points, strings, image references, lists, and positive integers. The reported id, width, and spacing examples are covered by E033 tests with exact positions.

tests/medui/SharedConformanceTests.cpp
ambroise-leclerc: Addressed in fdfd243. The adapter now rejects unknown inputs members and unknown members inside each text package, so new contract inputs cannot be silently ignored.

README.md
ambroise-leclerc: Addressed in fdfd243. README now lists MduXMeduiLib as Partial under Host tools and removes the compiler row from Not started.

AM-L (approved):

PR #212, opened 2026-08-16
S5 — Bounded layout and Row flattening

### LINKED ISSUE #194: S5 — Bounded layout and Row flattening
Part of #15. Blocked by S4; blocks S6 and S8.

## Work

- Resolve `Vertical`/`Row` layout with `spacing`/`padding`, `Npx` and `Fill` sizes, and out-of-flow `position: Xpx, Ypx` into absolute pixel rectangles at compile time.
- Flatten `Row` into its members. A nested `Row` is rejected by S3's parser; this stage assumes the single level and should assert it rather than trust it.
- Integer arithmetic only. The layout result feeds committed evidence, so it must be byte-identical across toolchains — the argument `mdux.text.raster` makes for the same reason.

## Acceptance

- No floating-point arithmetic anywhere in the solver.
- A fixture pins a non-trivial screen's resolved rectangles exactly; the expected values are derivable by hand rather than recorded from a run.
- Overflow (a child wider than its parent, a `Fill` with no room) is a diagnostic, not a clamp.

## Binding constraint from ADR-011 (#190, PR #202)

**The single flattening pass is the reason nested `Row` is banned.** ADR-011 decision 5 separates
the two rationales: loops, conditionals, recursion and scripting are excluded because they make the
`DrawBudget` uncomputable; nested `Row` is excluded so this solver stays one flattening pass over a
known structure. If the solver is ever generalised to handle nesting, that is a decision to revisit
here — the budget argument does not forbid it.

**Terminology:** this stage produces *layout* — rectangles, colours, budgets. Vertices and indices
are the runtime's output (#199). ADR-011 fixes the two words and neither record uses "geometry"
for the compiled form.

## Cross-implementation gaps found while verifying #192 against TrustSC

**`position` with a `Fill` dimension must be rejected.** TrustSC does this and MduX does not:

> `component {id}: \`position\` requires fixed \`width\`/\`height\` — Fill is flow-only`
> — `crates/trustsc-ui-dsl-authoring/src/lib.rs:667`

The reasoning is sound and belongs to this stage: a node taken out of flow at exact coordinates has
no parent to distribute `Fill` against, so the combination is meaningless rather than merely
unusual. The parser (#192) deliberately does not check it — it needs no dictionary, but it does
need to understand what `position` means, which is layout's job.

`MDX-E051` (LayoutOverflow) is the closest registered code but is not a good fit; this is a
contradiction in the declaration, not an overflow. Consider whether it warrants its own code in the
050-069 block, and add it to the registry if so — #191 left gaps in each block for exactly this.

ambroise-leclerc:
## Summary

Adds the host-only `mdux.tools.medui.layout` stage required by S5. It resolves `Vertical` flow and one-level `Row` groups with integer-only padding, spacing, fixed sizes, and `Fill`; removes positioned nodes from flow; validates fixed dimensions, containment, overlaps, and generated-ID collisions; and emits a fail-closed flat sequence of absolute rectangles. Row backgrounds become synthetic `<row-id>-background` Panels.

The shared MedUI `layout` capability is now advertised and executed. Review follow-up tightened zero-area and padded-surface handling, made all diagnostic results discard partial nodes, keyed background overlap exemptions on provenance rather than component names, added the required semantic gate, removed the invented shared-conformance surface, and documented absolute positioned coordinates.

Closes #194

Invitation / CLA issue: not applicable — maintainer-authored change.

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [x] `tools/CMakeLists.txt` — host-tool targets and their sources
- [ ] None of the above

## Rebase state

- **Rebased onto:** `f2c4654305242e47f7cbdcd9fb7222d1423fcdbc`

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run locally: this checkout is on unsupported macOS; supported Linux GCC 16 and MSVC PR checks cover the C++ build.
- [ ] `ctest --test-dir build-gcc` — not run locally for the same platform reason.
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 80 files checked, 0 findings.
- [x] Other: `python3 tools/docs-lint/generate_ai_reference.py --check` passed; `python3 tools/docs-lint/check_schema_type_drift.py` passed; `clang-format` and `git diff --check` passed. The docs-lint Python unit suite retains four pre-existing macOS-only internal-link expectation failures (`MDX-D012` instead of expected `MDX-D010`/`MDX-D011`). All required PR checks pass on review commit `f856468`, including GCC 16, ASan+UBSan, MSVC, CodeQL, CodeRabbit, compliance/documentation, governance, evidence, and security.

Test coverage includes hand-derived 1280×720 rectangles; flow and Row `Fill`; the TrustSC-compatible unused integer remainder; absolute positioned Row children; out-of-flow placement; surface and Row containment; zero fixed dimensions; padding that leaves no content; authored/build surface mismatch; positioned `Fill`; overflow; fail-closed partial-node cleanup; overlap including an authored `Panel` name; synthetic-ID collision; and the directly-built nested-Row invariant.

## Post-merge gate

- **Post-merge `develop` run:** pending after merge; required before the S6 (#195) or S8 (#197) successor merges.
- [ ] That run is green. *(Tick after merging this PR and before merging its successor.)*

## Regulatory impact

Classification: **safety-relevant**. This build-time stage decides the absolute bounds accepted for authored UI nodes and implements the layout restriction and fail-closed overflow behavior recorded by ADR-011.

Affected design and software item: ADR-011 and host-tools target `MduXMeduiLib`. Verification evidence is in `tests/medui/LayoutTests.cpp` and the pinned shared layout case. `README.md`, `docs/architecture.md`, and the MedUI authoring skill distinguish implemented layout from still-planned packaging, emission, and runtime stages. No certification or production-readiness claim is made.

Maintainer/domain-expert review is required before merge.

### DIFF
--- a/tests/medui/SharedConformanceTests.cpp
+++ b/tests/medui/SharedConformanceTests.cpp
@@ -19,10 +19,11 @@
  *   and must be backed by a case that really executed. Claiming a phase with no adapter fails
  *   here instead of passing quietly.
  *
- * Syntax and semantics are observable: `md::parse` answers syntax acceptance, and `md::analyze`
- * checks the portable theme-token and per-locale key views in semantic case inputs. Claiming
- * `layout` or `safety` still needs the solver (#194) and golden pass (#196), so a claim to either
- * is rejected below rather than silently evaluated by a stage that cannot see it.
+ * Syntax, semantics and layout are observable: `md::parse` answers syntax acceptance,
+ * `md::analyze` checks the portable theme-token and per-locale key views in semantic case inputs,
+ * and `md::resolveLayout` runs the integer-only bounded solver. Claiming `safety` still needs the
+ * golden pass (#196), so that claim is rejected below rather than silently evaluated by a stage
+ * that cannot see it.
  */
 
 import std;
@@ -32,6 +33,7 @@ import mdux.evidence.json;
 import mdux.text.schema;
 import mdux.tools.cli;
 import mdux.tools.medui.diagnostics;
+import mdux.tools.medui.layout;
 import mdux.tools.medui.parser;
 import mdux.tools.medui.semantic;
 import mdux.tools.toml;
@@ -47,7 +49,7 @@ namespace toml = mdux::tools::toml;
 
 /// Phases this file can genuinely observe. Anything else in `capabilities` is a claim with no
 /// adapter behind it.
-constexpr std::array<std::string_view, 2> runnableCapabilities{"syntax", "semantics"};
+constexpr std::array<std::string_view, 3> runnableCapabilities{"syntax", "semantics", "layout"};
 
 [[noreturn]] void fail(std::string message, std::source_location where = std::source_location::current()) {
     throw speclab::core::AssertionFailure(std::move(message), where);
@@ -597,6 +599,12 @@ void runCase(const Case& item, Positions positions, mdux::spec::Checks& checks) 
 
         md::SemanticResult semantic = md::analyze(*parsed.screen, file, md::SemanticInputs{.themeTokens = themeTokens, .textPackages = packages});
         diagnostics.insert(diagnostics.end(), std::make_move_iterator(semantic.diagnostics.begin()), std::make_move_iterator(semantic.diagnostics.end()));
+    }
+    if (item.phase == "layout" && parsed.screen) {
+        const std::int64_t width    = parsed.screen->surface ? parsed.screen->surface->x : 800;
+        const std::int64_t height   = parsed.screen->surface ? parsed.screen->surface->y : 600;
+        md::LayoutResult   resolved = md::resolveLayout(*parsed.screen, file, {.surfaceWidth = width, .surfaceHeight = height});
+        diagnostics.insert(diagnostics.end(), std::make_move_iterator(resolved.diagnostics.begin()), std::make_move_iterator(resolved.diagnostics.end()));
     }
 
     const bool accepted = parsed.screen.has_value() && diagnostics.empty();
--- a/tests/medui/SharedConformanceTests.cpp
+++ b/tests/medui/SharedConformanceTests.cpp
@@ -1,6 +1,6 @@
 /**
- * @file SharedConformanceTests.cpp
  * @brief The pinned `Compliatory/MedUI` conformance cases, run against this parser.
+ * @file SharedConformanceTests.cpp
  *
  * @compliance ADR-004 Trust zones in C++ (host tools zone)
  * @compliance ADR-011 The deterministic `.medui` compile boundary
@@ -32,6 +32,7 @@ import mdux.core.result;
 import mdux.evidence.json;
 import mdux.text.schema;
 import mdux.tools.cli;
+import mdux.tools.medui.ast;
 import mdux.tools.medui.diagnostics;
 import mdux.tools.medui.layout;
 import mdux.tools.medui.parser;
@@ -534,6 +535,32 @@ void checkComponentDictionary(const std::filesystem::path& checkout, mdux::spec:
     return joined;
 }
 
+/// Collects authored external names so layout cases can pass the semantic shape gate without
+/// inventing theme or locale failures that belong to the semantics phase.
+void collectLayoutSemanticNames(const md::ast::Node& node, std::set<std::string>& themeTokens, std::set<std::string>& textKeys) {
+    const auto collectValue = [&](auto&& self, const md::ast::Value& value) -> void {
+        if (value.kind == md::ast::ValueKind::ColorToken) {
+            themeTokens.insert(value.text);
+        } else if (value.kind == md::ast::ValueKind::TextKey) {
+            textKeys.insert(value.text);
+        }
+        for (const std::shared_ptr<md::ast::Value>& element : value.list) {
+            if (element != nullptr) {
+                self(self, *element);
+            }
+        }
+    };
+
+    for (const md::ast::Field& field : node.fields) {
+        if (field.value != nullptr) {
+            collectValue(collectValue, *field.value);
+        }
+    }
+    for (const md::ast::Node& child : node.children) {
+        collectLayoutSemanticNames(child, themeTokens, textKeys);
+    }
+}
+
 /// `spec/diagnostics.md`, "Positions in conformance cases": a pinned position is matched as far as
 /// the declared precision goes, and reporting more than was declared fails. The declaration is
 /// checked rather than tolerated, so precision cannot move in either direction without a manifest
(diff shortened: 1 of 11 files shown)


tools/medui/Layout.cpp
```diff
@@ -0,0 +1,399 @@
...
+        }
+        return true;
+    }
+
+    bool validatePositionedOverlap() {
+        for (std::size_t first = 0; first < result_.nodes.size(); ++first) {
+            for (std::size_t second = first + 1; second < result_.nodes.size(); ++second) {
+                const ResolvedNode& a = result_.nodes[first];
+                const ResolvedNode& b = result_.nodes[second];
+                if (a.component == "Panel" || b.component == "Panel" || (!a.positioned && !b.positioned) || !overlaps(a.bounds, b.bounds)) {
```
ambroise-leclerc: **medium** — The overlap exemption is keyed on the component *name* rather than on the `synthetic` flag that exists for exactly this purpose.

`ResolvedNode` already carries `synthetic`, and the only node this line means to exempt is the `<row-id>-background` Panel synthesised a few lines above. Matching on `component == "Panel"` instead makes the exemption apply to *any* node named `Panel`:

- `resolveLayout` is a public entry point taking a public AST, and the module header explicitly contemplates callers that build an AST directly (that is why `assertSingleRowLevel` throws). Such a caller can emit `Panel` leaves that skip the fail-closed overlap check entirely.
- `checkComponentDictionary` pins the implemented dictionary to upstream `spec/component-model.md`. The day upstream adds a `Panel` construct, every authored `Panel` silently stops being overlap-checked and no test in this suite notices — a safety-relevant restriction turning fail-open by a name collision.

```suggestion
                if (a.synthetic || b.synthetic || (!a.positioned && !b.positioned) || !overlaps(a.bounds, b.bounds)) {
```

tools/medui/Layout.cpp
```diff
@@ -0,0 +1,399 @@
...
+            }
+            const LayoutRect childBounds{.x = cursorX, .y = bounds.y, .width = (*resolvedWidths)[flowIndex], .height = resolvedHeight};
+            appendLeaf(child, childBounds, false);
+            cursorX += childBounds.width + rowSpacing;
+            ++flowIndex;
+        }
+        return true;
+    }
+
+    [[nodiscard]] std::int64_t fixedSize(const ast::Node& node, std::string_view name) const {
```
ambroise-leclerc, line 391: **medium** — `fixedSize` accepts `0px`, so a zero-area positioned node is emitted instead of being rejected.

The lexer only produces unsigned digits, so a size is either `Fill` or `>= 0`. The flow path rejects the degenerate case explicitly (`resolveAxis`: `dimension.pixels <= 0` → `MEDUI-E051`, and the vertical width check repeats `resolvedWidth <= 0`), but the positioned path has no equivalent guard: `contains()` returns true for a zero-width/zero-height rectangle (`inner.width < 0` is the only rejection, and `0 <= outer.width - relativeX` holds).

Concretely, `Label { id: ghost; width: 0px; height: 0px; position: 10px, 10px; ... }` is accepted and reaches `result_.nodes` as a `0x0` rectangle, while the identical `width: 0px` in flow is a diagnostic. Per ADR-011 every positioned node also emits a `Bounds` golden reference, so a degenerate rectangle propagates into the evidence rather than being caught here. Suggest rejecting non-positive fixed dimensions in `fixedSize` (or before `contains`) with `MEDUI-E051`.

tools/medui/Layout.cpp
```diff
@@ -0,0 +1,399 @@
...
+            const ast::Value& value = valueOf(*field);
+            if (value.kind != ast::ValueKind::Size || value.size.fill || value.size.pixels < 0) {
+                report(Code::LayoutOverflow, value.position, std::format("Row '{}' spacing must be a non-negative pixel size", rowId));
+                return false;
+            }
+            rowSpacing = value.size.pixels;
+        }
+
+        if (fieldFor(row, "background") != nullptr) {
+            result_.nodes.push_back(
```
ambroise-leclerc, line 348: **low** — The synthetic background is appended before the Row is validated, so a failed layout can return diagnostics *and* a partial node list.

`resolveRow` pushes `<row-id>-background` first, then can still fail on spacing, `resolveAxis`, containment, or a too-tall child — each returning `false` and leaving the Panel in `result_.nodes`. Scenario: a `Row` with `background:` whose child is wider than the Row returns one node plus `MEDUI-E051`. The two neighbouring scenarios in `LayoutTests.cpp` pin the opposite contract ("no partial rectangle is returned", "the failed axis produces no partial layout"), and `wideRowChildIsRejected` only passes because its fixture Row has no `background`. Either clear `result_.nodes` on the failure paths or drop the emptiness expectations from the tests so the contract is stated once.

tools/medui/Layout.cpp
```diff
@@ -0,0 +1,399 @@
...
+                report(Code::LayoutOverflow, value.position, std::format("layout {} must be a non-negative pixel size", name));
+                return false;
+            }
+            destination = value.size.pixels;
+            return true;
+        };
+        if (!readPixel("spacing", 0, spacing_) || !readPixel("padding", 0, padding_)) {
+            return false;
+        }
+        if (padding_ > inputs_.surfaceWidth / 2 || padding_ > inputs_.surfaceHeight / 2) {
```
ambroise-leclerc: **low** — Off-by-one in the padding guard: padding exactly half the surface passes, leaving a zero-sized content box.

The message says "layout padding leaves no surface content box", but `padding_ > surfaceWidth / 2` is false when `padding_ == surfaceWidth / 2`, which is precisely the case that produces `content.width = surfaceWidth - 2 * padding_ = 0` on an even surface. With `surface: 100px, 100px;` and `padding: 50px;`, a screen with no nodes is reported `ok()` with a 0x0 content box; with nodes, the author gets "… exceeds its available space" pointing at a child instead of the intended `MEDUI-E052` at the layout block. `>=` here (or a direct `content.width <= 0 || content.height <= 0` check) reports the real cause.

tools/medui/Layout.cpp
```diff
@@ -0,0 +1,399 @@
...
+        const auto resolvedWidths = resolveAxis(widths, bounds.width, rowSpacing, row.position, std::format("Row '{}'", rowId));
+        if (!resolvedWidths) {
+            return false;
+        }
+
+        std::int64_t cursorX   = bounds.x;
+        std::size_t  flowIndex = 0;
+        for (const ast::Node& child : row.children) {
+            if (const ast::Point* position = positionOf(child)) {
+                const LayoutRect childBounds{.x = position->x, .y = position->y, .width = fixedSize(child, "width"), .height = fixedSize(child, "height")};
```
ambroise-leclerc, line 367: **low** — Inside a Row, `position:` coordinates are absolute but must land in a band the author cannot compute.

`contains(bounds, childBounds)` compares the child's authored absolute coordinates against the Row rectangle, whose `y` is `cursorY` — a value derived from the padded origin plus every preceding flow item and gap. So a positioned Row child is only accepted if the author hand-computes the solver's vertical cursor; any natural coordinate (e.g. `position: 0px, 0px`) yields "component 'x' escapes Row 'r'". The same holds at screen level, where `contains(content, bounds)` rejects any position inside the padding band even though `SKILL.md` line 57 documents `position:` as "exact pixel coords". No scenario covers `position` together with non-zero `padding` (`sourceWithBody` always uses `padding: 0px`). Worth either documenting that positions are surface-absolute *and* restricted to the padded content box / the resolved Row band, or measuring positions from the containing box's origin.

tests/medui/SharedConformanceTests.cpp
```diff
@@ -598,6 +600,12 @@ void runCase(const Case& item, Positions positions, mdux::spec::Checks& checks)
         md::SemanticResult semantic = md::analyze(*parsed.screen, file, md::SemanticInputs{.themeTokens = themeTokens, .textPackages = packages});
         diagnostics.insert(diagnostics.end(), std::make_move_iterator(semantic.diagnostics.begin()), std::make_move_iterator(semantic.diagnostics.end()));
     }
+    if (item.phase == "layout" && parsed.screen) {
```
ambroise-leclerc: **medium** — The layout adapter calls `resolveLayout` without running `md::analyze` first, but `resolveLayout` throws on any AST that has not passed semantics.

`Layout.cppm` states the contract plainly: "Other malformed-AST conditions are treated the same way: semantic analysis is a required gate before layout" — `idOf`, `sizeOf` and `positionOf` all `throw std::logic_error`. Here the only gate is `parse`. A pinned `layout` case that parses but is semantically invalid — a component missing `width`/`height`, an `id` that is not an `Identifier`, a construct outside the dictionary, a `position:` that is not a `Point` — escapes as an uncaught `std::logic_error` out of `runCase`, aborting the whole `medui-shared-conformance` scenario instead of reporting a failed expectation. The corpus is external and pinned by commit, so bumping the pin can turn a case into a hard crash. Running `md::analyze` for the `layout` phase too (and folding its diagnostics in, or bailing when it produces any) keeps the stage's precondition honest.

tests/medui/SharedConformanceTests.cpp
```diff
@@ -598,6 +600,12 @@ void runCase(const Case& item, Positions positions, mdux::spec::Checks& checks)
...
+        const std::int64_t width    = parsed.screen->surface ? parsed.screen->surface->x : 800;
```
ambroise-leclerc: **low** — The 800x600 fallback surface is a local invention, and taking the build surface from the declared one makes `readSurface`'s mismatch branch unreachable here.

Two consequences: (1) a pinned `layout` case that declares no `surface:` is evaluated against a surface no part of the shared contract specifies, so its expected `MEDUI-E051`/`MEDUI-E052` can be satisfied — or spuriously violated — for a reason unrelated to the case; (2) because `width`/`height` are read back from `parsed.screen->surface`, the "declared surface {} does not match build surface {}" diagnostic in `readSurface` can never fire through this adapter, so no shared case can ever cover it. Consider requiring a declared `surface:` for `layout` cases (failing the case otherwise) rather than defaulting.

AM-L (approved):

Issue #201, opened 2026-08-11
S12 — First end-to-end screen


ambroise-leclerc:
Part of #15, and what closes it. Blocked by S9 and S10.

## Work

A committed `.medui` screen compiled at build time, linked as generated source, and drawn through `mdux.render.vulkan` — the first pixel that came from an authored screen rather than from example code.

- Reuse the existing offscreen target and readback (#125) rather than requiring a window.
- A pixel test under lavapipe, in the shape of the Wave 3 and Wave 4 tests.

## Acceptance

- The screen exercises at least one safety-critical node, so S7's golden reference has a real consumer.
- `docs/architecture.md` and the README describe the path from `.medui` to pixels as built rather than planned.
- The `medui-authoring` skill's "Status: planned, not yet implemented" banner is removed or rewritten.

PR #214, opened 2026-08-17
S7 — Golden references for safety-critical nodes [#196]

### LINKED ISSUE #196: S7 — Golden references for safety-critical nodes
Part of #15. Blocked by S5; consumed by #16.

## Work

Emit a golden reference per the `medui-authoring` skill's rules:

- `@safety_critical(cv_check: [...])` emits an entry with `node_id`, `bounds`, `text_key?`, `color_token?`, `cv_checks`.
- **Any** node with an explicit `position:` gets an automatic `Bounds` entry, annotation or not — a declared position is a safety-relevant claim by itself.
- A node with both gets exactly **one** merged entry with deduplicated `cv_checks`, never two.
- A `@safety_critical` node without an explicit `requirement:` is an error.
- Dynamic components pin bounds and colour, never the varying value.

## Acceptance

- A fixture covers each of the four rules above, including the merge case.
- The emitted shape is what #16's verifier will consume; agree it here rather than after.

## Binding constraints from ADR-011/ADR-012 (#190, PR #202)

**The predicate is fixed in ADR-011** and both this issue and #16 must use the same one: a
`@safety_critical` node emits an entry; **any** node with an explicit `position:` emits a `Bounds`
entry whether annotated or not; a node matching both emits exactly one merged entry with
deduplicated `cv_checks`. A `@safety_critical` node with no `requirement:` is an error.

**The consistency test compares sets, not references.** Do not write a test that walks
`goldens.json` and checks each id resolves in `package.json` — that is one-way and the wrong way
round. Because the predicate makes the file's content exactly determined, the test must apply the
predicate to `package.json`, build the expected id set, and assert **set equality**. The failure
that matters is a selected node whose golden is missing, and a resolve-only check cannot see it.

**Cross-checking cannot live in the governed `validate()`.** ADR-012 decision 4 keeps the goldens
outside the schema because the runtime never reads them, so a `static_assert` has no ids to compare
against. This is a host-side test.

ambroise-leclerc:
## Summary

S7 of the `.medui` compiler: the stage that decides which nodes a rendered-truth verifier must
check, where they must appear, and in what tint. Adds `mdux.tools.medui.goldens`
(`tools/medui/Goldens.{cppm,cpp}`).

ADR-011 fixes the predicate and #16 must use the same one, so it is restated rather than re-decided:

- a node carrying `@safety_critical` is selected;
- **any** node with an explicit `position:` is selected with an automatic `Bounds` check, annotated
  or not;
- a node matching both is selected **exactly once**, with the check lists merged and deduplicated;
- a `@safety_critical` node whose `requirement:` is absent **or blank** is `MEDUI-E070` — the field
  is a `String` and the semantic domain accepts `""`, so presence alone would let an untraceable
  node through the rule written to stop it;
- and only a component the dictionary gives a `requirement:` can carry the annotation at all, which
  is `CriticalButton`, `Button`, `NumericDisplay`, `StatusIndicator` and `TextInput`. `Row`,
  `Label`, `Clock`, `Image`, `SignalTrace` and `VulkanViewport` take none, so an annotation on any
  of them is rejected rather than silently unpinned.

**Two entry points, because the shared contract requires it.** `validateSafetyAnnotations()` applies
the annotation rules to a *parsed* screen and needs no layout;
`MEDUI-CASE-SAFETY-MISSING-REQUIREMENT` pins `MEDUI-E070` on a screen that declares no `surface:`,
which the solver cannot resolve at all — so a design that could only report it after layout could not
satisfy the contract it implements. `collectGoldens()` derives the set from the *resolved* screen and
returns references rather than a result type, because every rule that can fail was already decided
by the first call. A driver's order is parse → analyze → validate annotations → layout → collect.

**The safety phase becomes observable**, so `medui-conformance.toml` claims it and the shared harness
executes the pinned case instead of listing it unclaimed. The adapter deliberately skips semantic
analysis: a safety case carries no theme or locale inputs, and running `analyze` would add findings
the case never claimed.

**The shape is settled here**, as the issue asks, because #197 serialises it and #16 consumes it:
`node_id`, `bounds`, `text_key?`, `color_token?`, `cv_checks`, with `cv_checks` sorted and
deduplicated so one screen has one serialisation — `goldens.json` is byte-compared like every other
committed artifact.

**Dynamic content pins place and tint, never the value.** A list-valued `states:` or `colors:` leaves
`text_key` and `color_token` empty, because the state on screen is exactly the varying part; a golden
that pinned one would fail whenever the device showed another, which trains a team to ignore it.

Three judgement calls, each documented where it is made and listed here so a reviewer does not have
to find them:

1. **An annotation naming no `cv_check` still pins `Bounds`.** Read from ADR-011's rule that a
   declared position alone is enough — if that makes a rectangle worth verifying, an explicit
   safety-critical claim cannot be worth less. The alternative is an entry a verifier reads and does
   nothing with.
2. **A `ColorHash` check is refused where no single declared tint exists** — a `StatusIndicator`
   whose `colors:` is a list, or a `Clock`, `Image` or `VulkanViewport` with no colour at all.
   `cv_checks` and `color_token` are halves of one claim, and a reference that asked for a
   comparison while carrying no expectation is one #16 could only skip. Encoding per-state tints
   instead would extend the shape both #16 and #197 consume, so it belongs with the wave that owns
   state-dependent components (#17) — but it is cheaper to decide now than after `goldens.json` is
   committed, so say the word and it lands here.
3. **A malformed `cv_check` value reuses `MEDUI-E033`** rather than taking a new code. An annotation
   argument is an `ast::Field`, and "a field value has the wrong semantic kind" is exactly what a
   quoted or numeric check is; a new code would mean the same thing twice.

Closes #196

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked — #213 (S6) merged as `163371b`, and its post-merge `develop` run
  was green before this branch was cut

## Shared registries touched

- [x] `FILE_SET CXX_MODULES` lists — `medui/Goldens.cppm` added
- [x] `tools/CMakeLists.txt` — `MduXMeduiLib` sources
- [x] `tests/CMakeLists.txt` — `medui_tools_spec` sources

Also edits `medui-conformance.toml`, which is a pinned-contract manifest rather than a build
registry: `capabilities` gains `safety`.

## Rebase state

- **Rebased onto:** `163371b` (current `develop`)

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — **not run locally.** The preset is
      Linux-only, and the local macOS toolchain cannot complete a module build: with CMake 4.1.2 and
      Homebrew GCC 16.1.0 the configure succeeds but the build fails inside CMake's synthesized
      `__CMAKE::CXX23` target, because GCC's scan of `bits/std.cc` reports no module interface unit.
      That is before any MduX source compiles, and disabling ccache does not change it.
- [ ] `ctest --test-dir build-gcc` — not run locally, for the same reason.
- [x] `clang-format` applied to the new sources with the repository's `.clang-format`.

- [x] CI on `b6c8326` (this head): Linux (GCC 16, Vulkan) **521/521 tests passed**, Windows (MSVC,
      Vulkan) and ASan + UBSan (GCC 16) green, every lint and analysis leg green. The golden suite is
      14 scenarios. (The first Windows and osv-scan runs on this head failed while GitHub was
      returning 429/503 for action downloads — neither reached a compiler — and passed on re-run.)

With `safety` claimed, the pinned `MEDUI-CASE-SAFETY-MISSING-REQUIREMENT` case now executes at full
position precision under `medui-shared-conformance` rather than being skipped as unclaimed.

## Review round two

`b6c8326` addresses every finding on the first head; each is answered in its own thread.

- **Empty requirement (P1).** Blank is now the same error as absent, with `medui-goldens-blank-requirement`
  driving analyze → validate so the scenario shows *where* the gap was: semantic analysis genuinely
  accepts `requirement: ""`.
- **`ColorHash` with no tint (P1).** Refused, per judgement call 2 above, with
  `medui-goldens-colorhash-needs-a-tint` covering the annotated `StatusIndicator`, the no-colour
  `Clock`, and a positive control.
- **Annotated `Row` unreachable (P1, also raised by Copilot).** The behaviour is removed rather than
  supported: `Row` takes no `requirement:` in the pinned model, so it never reaches collection, and
  the previous scenario only saw that path by skipping the validation gate. `collectGoldens()` now
  skips synthetic nodes explicitly, and `medui-goldens-annotated-row` pins the rejection through
  parse → validate.
- **AGENTS.md remaining children.** Now lists all five (#197–#201) rather than three.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

Safety-relevant, and this is the change that creates the evidence rather than one that could weaken
it. A golden reference is the compiler's written statement of what a rendered frame must show for
the content that matters clinically, produced at build time so #16's verifier checks a frame against
a reviewed expectation rather than against the source that produced it. ADR-012 § "Decision 1"
commits the file beside the screen package precisely so that a moved safety-critical rectangle is a
reviewable hunk with its own digest.

The rule with the most traceability weight is `MEDUI-E070`: a `@safety_critical` node with no
`requirement:` fails to compile, so a safety-critical node cannot exist without something to trace
it to. `requirement:` is optional on `Button` and `TextInput`, which is exactly why the annotation
needs its own rule.

Artifacts updated: `docs/architecture.md`, `README.md`, `.agents/skills/medui-authoring/SKILL.md`,
`docs/roadmap.md`, and `AGENTS.md` — whose skill table still described the compiler as "Planned — no
`.medui` compiler exists yet", an underclaim by five children that misroutes an agent deciding
whether to read the skill. No ADR changes: ADR-011 and ADR-012 already decided this and are cited
from the module rather than restated. No new certification or compliance claim.

## Summary by CodeRabbit

* **New Features**
  * Added golden references for safety-critical interface elements.
  * Added validation for safety annotations, required checks, and unsupported checks.
  * Added support for collecting and validating pinned bounds, text, colors, and dynamic content.

* **Bug Fixes**
  * Improved diagnostics for missing safety requirements and invalid checks.
  * Prevented golden-reference collection when safety validation has not passed.

* **Tests**
  * Added comprehensive coverage for accepted and rejected safety scenarios.

* **Documentation**
  * Updated implementation status, architecture documentation, and roadmap progress.

### DIFF
--- a/tools/medui/Goldens.cpp
+++ b/tools/medui/Goldens.cpp
@@ -63,6 +63,29 @@ constexpr std::string_view cvCheckArgument = "cv_check";
     return nullptr;
 }
 
+/// Whether the component dictionary gives this component a `requirement:` field at all.
+///
+/// `Row` does not have one, and the dictionary is the pinned shared contract rather than something
+/// this compiler may extend - so a container cannot carry `@safety_critical`, and the diagnostic
+/// should say that rather than asking an author to add a field the language will then reject.
+[[nodiscard]] bool admitsRequirement(const ast::Node& node) noexcept {
+    const std::span<const ComponentRule> dictionary = componentDictionary();
+    const auto                           component  = std::ranges::find(dictionary, node.component, &ComponentRule::name);
+    if (component == dictionary.end()) {
+        return false;
+    }
+    return std::ranges::any_of(component->fields, [](const FieldRule& rule) {
+        return rule.name == "requirement";
+    });
+}
+
+/// Whether `text` carries anything a requirement identifier could be. Blank is not traceable.
+[[nodiscard]] bool isBlank(std::string_view text) noexcept {
+    return std::ranges::all_of(text, [](unsigned char character) {
+        return std::isspace(character) != 0;
+    });
+}
+
 /// The text a golden pins, or empty when the node draws none - or draws one that varies.
 ///
 /// Only a single static key qualifies. A list-valued field such as `StatusIndicator`'s `states:` is
@@ -170,11 +193,8 @@ private:
     void validateAnnotation(const ast::Node& node, const ast::Annotation& annotation) {
         // Reported at the annotation rather than at the node: the annotation is the claim that
         // needs a requirement behind it, and the shared case pins that position.
-        if (fieldFor(node, "requirement") == nullptr) {
-            report(Code::SafetyCriticalWithoutRequirement,
-                   annotation.position,
-                   std::format("component '{}' is annotated @safety_critical but declares no requirement", node.component));
-        }
+        validateRequirement(node, annotation);
+        validateColorHashIsDerivable(node, annotation);
 
         const ast::Field* argument = argumentFor(annotation, cvCheckArgument);
         if (argument == nullptr || argument->value == nullptr) {
@@ -189,6 +209,89 @@ private:
             return;
         }
         validateCheckValue(*argument->value);
+    }
+
+    /**
+     * @brief Requires something to trace the annotation to, not merely a field that exists.
+     *
+     * Presence alone was the first revision's rule and it was too weak: `requirement:` is a
+     * `String` field, `Semantic.cpp`'s domain check accepts every quoted string including `""`, and
+     * a safety-critical node whose requirement is the empty string is exactly the untraceable node
+     * this rule exists to prevent. A blank value therefore reports `MEDUI-E070` as an absent one
+     * does - the code means "carries no requirement", and an empty string carries none.
+     *
+     * The rule is enforced here rather than by tightening `FieldDomain::String`, because the
+     * traceability requirement belongs to the annotation. `template:` and `source:` are the same
+     * domain and are not the same claim.
+     */
+    void validateRequirement(const ast::Node& node, const ast::Annotation& annotation) {
+        const ast::Field* requirement = fieldFor(node, "requirement");
+        if (requirement != nullptr && requirement->value != nullptr && requirement->value->kind == ast::ValueKind::String
+            && !isBlank(requirement->value->text)) {
+            return;
+        }
+
+        // A container cannot satisfy this rule at all: the pinned dictionary gives `Row` no
+        // `requirement:`, so saying "add requirement:" would send an author to a field the language
+        // then rejects as unknown. Say what is actually wrong instead.
+        if (!admitsRequirement(node)) {
+            report(Code::SafetyCriticalWithoutRequirement,
+                   annotation.position,
+                   std::format("component '{}' takes no requirement in the shared component model, so it cannot be "
+                               "@safety_critical; annotate the content it contains instead",
+                               node.component));
+            return;
+        }
+        report(Code::SafetyCriticalWithoutRequirement,
+               annotation.position,
+               requirement == nullptr ? std::format("component '{}' is annotated @safety_critical but declares no requirement", node.component)
+                                      : std::format("component '{}' is annotated @safety_critical but its requirement is empty", node.component));
+    }
+
+    /**
+     * @brief Refuses a `ColorHash` check on a node whose expected tint cannot be derived.
+     *
+     * `cv_checks` and `color_token` are two halves of one claim: a verifier asked for `ColorHash`
+     * reads the token to know what tint to expect. A `StatusIndicator` declares `colors:` as a list,
+     * one per state, and a `Clock`, `Image` or `VulkanViewport` declares none at all - so a golden
+     * over either would ask for a check with nothing to check it against.
+     *
+     * Rejecting is the fail-closed reading, and it is a decision this issue has to take rather than
+     * defer, because the reference shape is frozen here for #16 and #197. The alternative - carrying
+     * a per-state list of expected tints - is a genuine extension of the shape, and one both
+     * consumers would have to agree to; it belongs with the components that would need it (#17),
+     * not with a stage inventing a field ahead of them.
+     */
+    void validateColorHashIsDerivable(const ast::Node& node, const ast::Annotation& annotation) {
+        const ast::Field* argument = argumentFor(annotation, cvCheckArgument);
+        if (argument == nullptr || argument->value == nullptr || !colorTokenOf(node).empty()) {
+            return;
+        }
+
+        const auto namesColorHash = [](const ast::Value& value) {
+            return value.kind == ast::ValueKind::Identifier && parseCvCheck(value.text) == CvCheck::ColorHash;
+        };
+
+        const ast::Value* offending = nullptr;
+        if (argument->value->kind == ast::ValueKind::List) {
+            for (const std::shared_ptr<ast::Value>& element : argument->value->list) {
+                if (element != nullptr && namesColorHash(*element)) {
+                    offending = element.get();
+                    break;
+                }
+            }
+        } else if (namesColorHash(*argument->value)) {
+            offending = argument->value.get();
+        }
+
+        if (offending != nullptr) {
+            report(Code::UnknownCvCheck,
+                   offending->position,
+                   std::format("{} needs one declared colour token to compare against, and component '{}' declares "
+                               "none that a golden could pin",
+                               spell(CvCheck::ColorHash),
+                               node.component));
+        }
     }
 
     void validateCheckValue(const ast::Value& value) {
@@ -243,6 +346,15 @@ std::vector<GoldenReference> collectGoldens(const LayoutResult& layout) {
     std::vector<GoldenReference> references;
 
     for (const ResolvedNode& node : layout.nodes) {
+        // Goldens describe authored nodes. The only synthetic node the solver produces is a Row's
+        // background, and a Row cannot be annotated - the pinned dictionary gives it no
+        // `requirement:`, so `validateSafetyAnnotations()` rejects the annotation before layout runs
+        // - nor positioned, since `position:` is not one of its fields either. Skipping it states
+        // that rather than leaving a reader to derive it from two other files.
+        if (node.synthetic) {
+            continue;
+        }
+
         std::vector<CvCheck> checks;
 
         const bool annotated = std::ranges::any_of(node.source.annotations, isSafetyCritical);
--- a/tools/medui/Goldens.cpp
+++ b/tools/medui/Goldens.cpp
@@ -1,0 +1,287 @@
+/**
+ * @file Goldens.cpp
+ * @brief The `@safety_critical` annotation rules, and the golden set they select.
+ *
+ * @compliance ADR-004 Trust zones in C++ (host tools zone)
+ * @compliance ADR-011 The deterministic `.medui` compile boundary
+ * @compliance ADR-012 What a compiled screen emits, and which parts are committed
+ */
+module;
+
+module mdux.tools.medui.goldens;
+
+import std;
+import mdux.tools.cli;
+import mdux.tools.medui.ast;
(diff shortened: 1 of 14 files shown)


tools/medui/Goldens.cpp
```diff
@@ -0,0 +1,287 @@
...
+        }
+        for (const ast::Node& child : node.children) {
+            validateNode(child);
+        }
+    }
+
+    void validateAnnotation(const ast::Node& node, const ast::Annotation& annotation) {
+        // Reported at the annotation rather than at the node: the annotation is the claim that
+        // needs a requirement behind it, and the shared case pins that position.
+        if (fieldFor(node, "requirement") == nullptr) {
```
ambroise-leclerc: **[P1] Validate a non-empty requirement identifier, not only field presence.**

The intended driver runs semantic analysis first, but `Semantic.cpp::matches(FieldDomain::String)` accepts every quoted string, including the empty string. Consequently a legal `Button` or `TextInput` containing `requirement: "";` passes semantic analysis and this check, then receives a safety golden even though there is no requirement identifier to trace.

That contradicts both the shared rule (requires an explicit requirement identifier) and the regulatory rationale in this PR that a safety-critical node cannot exist without something to trace it to. Please inspect the field value here, or strengthen the semantic domain used specifically for requirement identifiers, reject an empty value, and add an end-to-end regression covering analyze → safety validation.

tools/medui/Goldens.cpp
```diff
@@ -0,0 +1,287 @@
...
+        // reads and does nothing with.
+        if (checks.empty()) {
+            addCheck(checks, CvCheck::Bounds);
+        }
+        std::ranges::sort(checks);
+
+        references.push_back(GoldenReference{.nodeId     = node.id,
+                                             .bounds     = node.bounds,
+                                             .textKey    = textKeyOf(node.source),
+                                             .colorToken = colorTokenOf(node.source),
```
ambroise-leclerc, line 392: **[P1] Do not emit `ColorHash` when the reference has no expected color.**

A valid `StatusIndicator` can carry `requirement:`, list-valued `colors:`, and `@safety_critical(cv_check: [ColorHash])`. Semantic and safety validation both accept that source; `collectAnnotationChecks()` preserves `ColorHash`, while `colorTokenOf()` deliberately returns an empty string for the list-valued color domain. The resulting reference therefore asks #16 to perform a color check but contains no tint from which the expected result can be derived.

Because this PR freezes the consumer shape, this combination needs a defined representation now: encode the allowed or state-dependent colors, reject `ColorHash` when no single applicable token exists, or otherwise define a verifiable expectation. Please add a regression using an annotated `StatusIndicator` with `colors:` and `ColorHash`.

tools/medui/Goldens.cpp
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+
+    void validateAnnotation(const ast::Node& node, const ast::Annotation& annotation) {
+        // Reported at the annotation rather than at the node: the annotation is the claim that
+        // needs a requirement behind it, and the shared case pins that position.
+        if (fieldFor(node, "requirement") == nullptr) {
+            report(Code::SafetyCriticalWithoutRequirement,
+                   annotation.position,
+                   std::format("component '{}' is annotated @safety_critical but declares no requirement", node.component));
```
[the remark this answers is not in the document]
ambroise-leclerc: The pinned shared component model makes the resolution narrower than the two alternatives above: `Row` permits only `id`, `height`, `spacing`, and `background`, so adding `requirement` locally would diverge from the contract this PR claims to implement. Unless the shared contract is revised first, the consistent fix is to remove or reject annotated-Row behavior and make the scenario exercise the full analyze → validate → layout → collect order. There is a second consequence worth covering: an annotated Row without a background currently produces no resolved node and therefore no golden, despite ADR-011 stating that every `@safety_critical` node emits an entry.

tools/medui/Goldens.cpp
ambroise-leclerc: Right, and fixed in `b6c8326`. `requirement:` is a `String` field and `Semantic.cpp`'s domain check accepts every quoted string, `""` included, so presence alone let through exactly the untraceable node the rule exists to stop. A blank value (empty, or whitespace only) is now MEDUI-E070 with a message that says the requirement is empty rather than absent.

I enforced it here rather than by tightening `FieldDomain::String`: `template:` and `source:` share that domain and are not the same claim, and narrowing a shared domain to serve one rule would put the traceability requirement somewhere no reader of the annotation would look for it. If you would rather the domain rejected blank strings generally, that is a different change with a wider blast radius and I would rather it be its own PR than a side effect of this one.

The regression is `medui-goldens-blank-requirement`, and it drives the order you asked for: it asserts that `md::analyze` *accepts* the screen — otherwise the scenario would prove nothing about where the gap was — and then that `validateSafetyAnnotations` rejects it.

tools/medui/Goldens.cpp
ambroise-leclerc: Agreed, and fixed in `b6c8326` by refusing the combination.

A `ColorHash` check on a node with no single declared colour token is now MEDUI-E071, reported at the check that cannot be satisfied. That covers the `StatusIndicator` case you named and two more with the same shape: `Clock`, `Image` and `VulkanViewport` declare no colour at all, so an annotation asking for a tint comparison there was equally underivable.

On the three options you listed, I took "reject" rather than "encode the state-dependent colours" deliberately. Encoding them is a real extension of the reference shape — a list of expected tints, plus a rule for which one applies when — and both #16 and #197 would have to agree to it. This issue freezes the shape, so the conservative move is to make the underivable combination a compile error now and leave the extension to the wave that owns state-dependent components (#17). If you would rather the shape carried per-state tints, that decision is cheaper here than after `goldens.json` is committed, so say so and I will land it in this PR instead.

Regression: `medui-goldens-colorhash-needs-a-tint`, covering the annotated `StatusIndicator` with `colors:` you asked for, the no-colour `Clock`, and a positive control so the rule cannot pass by rejecting everything.

tools/medui/Goldens.cpp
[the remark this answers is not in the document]
ambroise-leclerc: You are right, and I have taken the narrower resolution you point to: the annotated-Row behaviour is removed rather than supported, in `b6c8326`.

Row takes only `id`, `height`, `spacing` and `background` in the pinned model, so an annotated Row always fails MEDUI-E070 and never reaches collection. My scenario only saw the Row-background path because it called `collectGoldens()` without validating first — it was pinning behaviour a real driver cannot reach, which is worse than having no scenario. Adding `requirement` to Row locally would diverge from the contract this PR claims to implement, so that option was out.

What replaces it: `medui-goldens-annotated-row` drives parse → validate and asserts the rejection, and the diagnostic now says the component takes no requirement in the shared component model rather than "declares no requirement" — the earlier wording sent an author to a field the language would then reject as unknown. `collectGoldens()` also skips synthetic nodes explicitly, with `medui-goldens-synthetic-nodes` covering it; that was previously implied by two other files, and the implication is what let the unreachable path look reachable.

On your second consequence — an annotated Row with no background producing no golden despite ADR-011 saying every `@safety_critical` node emits an entry — it dissolves rather than needing a rule: no annotated Row compiles at all now, so there is no valid screen in which ADR-011's predicate selects a node that emits nothing. The module comment states the general form, since Row is not alone: `Label`, `Clock`, `Image`, `SignalTrace` and `VulkanViewport` take no requirement either, so only `CriticalButton`, `Button`, `NumericDisplay`, `StatusIndicator` and `TextInput` can carry the annotation.

AM-L (approved):

PR #217, opened 2026-08-19
Settle the compiled-screen JSON member casing [#197]

ambroise-leclerc:
## Summary

Settles one convention per screen directory — **camelCase, in `package.json`, `goldens.json` and
`report.json` alike** — before anything writes a byte of it.

ADR-011's golden entry becomes `nodeId`, `bounds`, `textKey`, `colorToken`, `cvChecks`, with the
amendment recorded in the ADR itself; ADR-012 states the rule for the directory as a whole; the
`medui-authoring` skill and the `mdux.tools.medui.goldens` module comment follow.

**Why the spelling it replaces was never a format decision.** `node_id`, `text_key`, `color_token`
and `cv_checks` were transcribed from TrustSC's *Rust identifiers*. Two facts, both checked rather
than assumed, settle it the other way:

- **TrustSC emits no screen JSON at all.** `CompiledScreenPackage` derives `Clone, Copy, Debug, Eq,
  PartialEq` and no `Serialize`, and the repository holds no `generated/screens/*/package.json` — a
  compiled screen there is generated Rust source. There was no sibling file being matched.
- **Every committed MduX package is camelCase** — `byteLength`, `occupancyPercent`, `advanceWidth`,
  `codePoint` — across four artifact kinds, as is the repository's own tooling schema (`fixHint`,
  `filesChecked`). A directory in two conventions would have made #16's verifier, which reads two of
  these three files, carry two vocabularies to do it.

The parity that *is* achievable here is untouched: the same five facts per entry, the same file
structure, under names spelled the way this repository spells every artifact it commits.

**Why now rather than with the emitters.** Renaming a member of a committed, byte-compared artifact
costs a `schemaVersion` bump and a rebake of every screen. Today it costs a line in a writer that
does not exist yet. #198's first bake is the point of no return, and this lands well before it.

Part of #197. No code changes — nothing serialises these names until the emitters, which is exactly
why the decision is cheap today.

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked — #216 merged as `7cdceb8`, post-merge `develop` green (11/11)
- **Merge order:** mergeable now; the S8 emitters PR should be written against it so it emits these
  names from the start

## Shared registries touched

- [ ] `FILE_SET CXX_MODULES` lists — no module added, removed or moved
- [ ] `tests/CMakeLists.txt` — test source lists and suite membership
- [ ] Schemas under `docs/*/schemas/` or `docs/governance/schemas/` — none define these members
- [ ] Committed artifacts under `generated/` — no screen artifact exists yet, which is the point
- [x] None of the above

## Rebase state

- **Rebased onto:** `7cdceb8` (current `develop`)

## Verification

- [x] Swept the tree for the old spelling. What remains is deliberate and was checked line by line:
      ADR-011 quoting TrustSC's own `CompiledScreenPackage` fields, and the two places that record
      what the previous spelling *was*. `resolve_color_token()` and `THEME_COLORS` stay untouched
      throughout — they are the sibling's API, not ours.
- [x] `clang-format` applied to the three touched sources.
- [ ] `cmake --preset ninja-gcc && ctest` — not run locally; the macOS toolchain cannot build the
      `std` module. Comment-only changes to compiled files, so CI is the check that they still parse.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

None to the behaviour of any code: no rule changes, no diagnostic changes, and nothing that runs on
a device. It fixes the vocabulary of an artifact that is a configuration item under IEC 62304 before
that artifact exists, which is the cheapest moment for it and the only one where no digest moves.
ADR-011 and ADR-012 carry the reasoning; the skill and the module comment follow them rather than
restating the argument.

## Summary by CodeRabbit

* **Documentation**
  * Standardized compiled-screen and safety-reference field names to camelCase.
  * Clarified canonical JSON member names, nested bounds ordering, and byte-order requirements.
  * Documented the naming decision and related schema/versioning implications.
  * Updated guidance for text, color, and visual-check fields.

### DIFF
--- a/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
+++ b/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
@@ -150,8 +150,11 @@ format. Two facts settle it the other way, and both were checked rather than ass
   a compiled screen there is generated Rust source. There is therefore no sibling file for these
   member names to match, and no parity is given up by choosing them freely.
 - **Every committed MduX package is camelCase** — `byteLength`, `occupancyPercent`, `advanceWidth`,
-  `codePoint` — across the font, text, shader and model kinds, and so is the repository's own
-  tooling schema (`fixHint`, `filesChecked`). A screen directory holding `package.json`,
+  `codePoint` — across the three kinds that have committed artifacts today: font, shader and model
+  (`generated/font/`, `generated/shader/`, `generated/model/`). The text pipeline uses the same
+  convention in `mdux.text.schema` and `mdux-textbake`, but commits no package, so it is implemented
+  machinery rather than evidence and is named here as such. The repository's own tooling schema
+  (`fixHint`, `filesChecked`) is camelCase too. A screen directory holding `package.json`,
   `report.json` and `goldens.json` in two conventions would make #16's verifier, which reads two of
   the three, carry two vocabularies for no gain.
 
--- a/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
+++ b/docs/adr/ADR-011-deterministic-medui-compile-boundary.md
@@ -111,7 +111,7 @@ parity programme's direction is MduX moving toward it.
 
 **A compiled screen is locale-free.** It carries no locale field and no glyph runs. Per-locale
 glyph runs stay in the text package where ADR-010 already put them, and the two are joined on device
-by looking each node's `text_key` up for the running locale. TrustSC does exactly this in
+by looking each node's `textKey` up for the running locale. TrustSC does exactly this in
 `ScreenTextLayout::from_screen(screen, package, locale)`.
 
 The consequence that decides it: **adding an approved locale touches no screen artifact at all.**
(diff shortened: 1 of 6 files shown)


docs/adr/ADR-011-deterministic-medui-compile-boundary.md
```diff
@@ -137,9 +137,30 @@ and `governed.noThrow.symbolScan` hold the runtime to that.
...
+earlier revision of this ADR spelled them `node_id`, `text_key`, `color_token` and `cv_checks`, and
+that spelling was a transcription of TrustSC's *Rust identifiers* rather than a decision about a file
+format. Two facts settle it the other way, and both were checked rather than assumed:
+
+- **TrustSC emits no screen JSON at all.** Its `CompiledScreenPackage` derives `Clone, Copy, Debug,
+  Eq, PartialEq` and no `Serialize`, and its repository holds no `generated/screens/*/package.json` —
+  a compiled screen there is generated Rust source. There is therefore no sibling file for these
+  member names to match, and no parity is given up by choosing them freely.
+- **Every committed MduX package is camelCase** — `byteLength`, `occupancyPercent`, `advanceWidth`,
+  `codePoint` — across the font, text, shader and model kinds, and so is the repository's own
```
ambroise-leclerc: The evidence cited here overstates the current repository state. `generated/` contains committed packages for three artifact kinds only: `font`, `shader`, and `model`; there is no committed `generated/text/<id>/package.json`. The text schema and `mdux-textbake` exist, but that is implemented machinery rather than committed evidence. Because this ADR is being used to justify the format decision from checked repository facts, please either say “across the committed font, shader and model kinds” or explicitly separate those committed packages from the camelCase text schema/baker. That keeps the implemented-evidence versus project-state distinction accurate.

tools/medui/Goldens.cppm
```diff
@@ -64,20 +64,27 @@
  * one-to-one onto the canonical JSON:
  *
  * ```json
- * { "node_id": "sedation-index",
+ * { "nodeId": "sedation-index",
```
ambroise-leclerc: This example is described as canonical JSON, but it does not have the byte order that `mdux.evidence.json` emits. That writer sorts every object lexicographically by UTF-8 code unit, so the outer members serialize as `bounds`, `colorToken`, `cvChecks`, `nodeId`, `textKey`; the nested rectangle serializes as `height`, `width`, `x`, `y`. JSON object order is semantically irrelevant in general, but this repository deliberately makes it relevant to byte identity. Please reorder the example so the format being settled here matches the canonical artifact that #197 will actually write.

ambroise-leclerc:
## Review summary

Two documentation-contract issues should be corrected before merge:

1. ADR-011 counts `text` among committed artifact kinds, although no `generated/text/<id>/package.json` is committed. The schema and baker exist, but that is distinct from committed evidence.
2. The `Goldens.cppm` example is labeled canonical JSON but does not use the lexicographic object-member order emitted by `mdux.evidence.json`, including within `bounds`.

Detailed corrections are attached inline. The camelCase decision itself is consistent with current committed packages, the repository-wide spelling sweep is clean, and the checked TrustSC comparison supports the rationale.

Local verification: `mdux-docs-lint` passed (80 files, 0 findings), `clang-format --dry-run --Werror` passed for all three touched C++ sources, and `git diff --check` passed.

Impact classification: potentially safety-relevant documentation change because it defines future rendered-truth evidence, with no current runtime behavior change.

GitHub does not permit the PR author to submit a formal “request changes” review on their own PR; this comment records that disposition instead.

docs/adr/ADR-011-deterministic-medui-compile-boundary.md
ambroise-leclerc: Corrected in `cc76ac7`. You are right, and the failing is worth naming precisely: this ADR argues a format decision *from checked repository facts*, so an overstated fact is not a wording slip — it weakens the argument it was there to support.

`generated/` holds three kinds, not four: `font`, `shader` and `model`. The text pipeline uses camelCase in `mdux.text.schema` and `mdux-textbake`, but it commits no package, so it is implemented machinery rather than committed evidence, and the ADR now names it as exactly that in a separate clause. The same overstatement had been copied into ADR-012 and into the goldens module comment; both are fixed in the same commit.

The argument itself is unaffected — three committed kinds, all camelCase, plus the repository's own tooling schema — but it now rests on what a reader can verify with `ls generated/`.

tools/medui/Goldens.cppm
ambroise-leclerc: Fixed in `cc76ac7`. The example was written in the order a person types, not the order the file has.

`mdux.evidence.json` sorts every object's members lexicographically by UTF-8 code unit, nested objects included, so the entry serialises as `bounds`, `colorToken`, `cvChecks`, `nodeId`, `textKey`, and the rectangle as `height`, `width`, `x`, `y`. The example now reads that way.

Your framing is the part I want to keep in the file rather than only in this thread: object order is semantically irrelevant in JSON *in general*, and this repository deliberately makes it relevant to byte identity. For a file `evidence.screen.<id>` compares across toolchains, an example in hand-written order was describing a file that cannot exist — and it would have been copied into the emitter by whoever implements #197. So the comment now states the sorting rule and why the example follows it, which is the part that stops the next person from quietly restoring a readable order.

AM-L (approved):

Issue #199, opened 2026-08-11
S10 — Allocation-free screen runtime


ambroise-leclerc:
Part of #15. Blocked by S8. This is the governed half of the epic.

## Work

- A governed `mdux.medui.screen` (name to be settled in S1's ADR) consuming a compiled screen and producing a `DrawList` against a compiler-computed `DrawBudget` — buffers sized once, never grown.
- No parsing, no layout, no shaping, no allocation. `mdux.ml.runtime`'s `predict()` is the pattern: caller-supplied scratch, `noexcept`, `Result`-returning.

## Acceptance

- Added to `MduXCore`, so `mdux-governed-lint` and `governed.noThrow.symbolScan` cover it automatically — both derive their scope from the target rather than from a second list.
- A no-heap test in the shape of `tests/ml/NoHeapTests.cpp`, plus the object-level allocator scan extended to cover it.
- The draw output is byte-identical across toolchains for a fixed input screen.

## Binding constraint from ADR-011 (#190, PR #202)

**Nothing currently bounds this runtime's frame cost, and this issue owns the decision.**

Stated precisely because the neighbouring guarantees make it easy to assume otherwise: a no-heap
test proves the runtime does not allocate. That is neither termination nor a bound on iteration, and
the two must not be conflated in this issue's acceptance criteria.

The *invariant* is available. A compiled screen has a fixed node count and a `DrawBudget` whose
`maxVertices`, `maxIndices` and `maxCommands` are computed at build time, every `DrawList` operation
fails closed once a budget is exhausted, and the ban on loops and recursion in the source language
keeps those numbers finite. A runtime that iterates only over the package's nodes and writes only
through `DrawList` therefore does work bounded by numbers known before the device runs.

What does not exist is any mechanism holding the runtime to it — not in the schema, not in a test.
This issue must do one of two things and say which:

1. **Encode the bound in the schema** — a declared maximum work figure the runtime asserts against.
   If so, the field belongs to #197 and must land **before** the first screen artifact is committed:
   `package.json` is byte-compared, so a late field is a version bump and a rebake of every screen.
2. **Record that it did not**, as an open availability risk with an owner.

ADR-011 deliberately does not choose. What it does forbid is inheriting the guarantee as settled.

**Terminology:** this runtime computes *vertices* from a compiled screen's *layout*. It may compute
geometry; it may not compute structure.

## Decision on the frame-cost bound — supersedes the section above

The two options recorded above were a false choice, and this replaces them.

[...]

PR #218, opened 2026-08-19
Give each component its own compiled payload [#197]

ambroise-leclerc:
## Summary

Replaces the compiled node's flat record with **one typed payload per component**, before anything
serialises or bakes it. This is the S8 blocker identified in the parity note: the flat node was not
merely unlike TrustSC's, it was **lossy**.

A flat `CompiledNode { id, component, bounds, textKey, colorToken, requirement }` had nowhere to put

- a `Clock`'s `format`,
- a `NumericDisplay`'s `template` and `source`,
- a `StatusIndicator`'s state keys and per-state tints,
- a `CriticalButton`'s `on_press`,
- a `VulkanViewport`'s `stream_source`,
- a `TextInput`'s `max_length` and `charset`,

so a device holding the package could not have rendered four of the eleven components in the closed
dictionary. Nothing consumes the schema yet, which is exactly why this is cheap today and a
`schemaVersion` bump plus a rebake of every screen after #198.

**Semantics preserved from TrustSC, representation not copied.** Rust's enum carries payloads for
free; this had to be built for `constexpr` construction, for storage the generated translation unit
owns, and for the governed no-allocation rules. `std::variant` reached only through
`holds_alternative` and `get_if` — **`std::get` is deliberately never used**, because it throws on
the wrong alternative and ADR-005 does not allow a governed type whose natural accessor throws. The
alternative order is not a wire contract: the emitter writes the dictionary name `kindName()`
returns, so an alternative can be added without renumbering an artifact.

**One deliberate difference from the sibling.** `format`, `charset` and `onPress` stay
`std::string_view` names rather than becoming enumerations as TrustSC's `ClockFormat` and
`SystemEvent` do. Closing two of them was tried on this branch and withdrawn — see the review round
below — and the coordinated change is #219.

> ### What changed while this PR was in review
>
> The description above is what the PR does *now*. Two mechanisms were added and then removed
> during review, and a reader comparing the diff to an earlier version of this text deserves to
> know why rather than to reconstruct it:
>
> - **`NodeProvenance`** — two booleans per node so golden completeness could be re-derived from
>   `package.json`. Withdrawn (`52b370b`) once TrustSC's compiler showed the property is guaranteed
>   by construction: one implementation of the predicate, applied once, guarded by the tests that
>   live with it. The comparison it enabled could only ever have detected two writers disagreeing,
>   never a wrong predicate.
> - **`ClockFormat` and `SystemEvent` as closed enums** — withdrawn (`9b4e4d6`), because the
>   implemented front end accepts any identifier for those fields and #195's budget resolves a clock
>   format through a product table, so closing the set here alone left the canonical type unable to
>   represent a screen the compiler accepts. #219 carries the coordinated change, and opens with the
>   question that blocks it: no registered diagnostic reports a named value outside a closed set.

**Two rules the typed payload makes checkable, that a shared field could not express:**

- `color` is required by the dictionary on six components and optional only inside
  `StatusIndicator`'s per-state list. One common field cannot be required for some components and
  absent for others, so the flat node could only ever permit an empty tint everywhere.
- Per-state tints must pair one-to-one with the states. A short list leaves a state with no colour
  and no way to say which state that is.

`validate()` enforces both, plus the required names each dictionary entry declares.

Also lands two ADR-012 paragraphs the review asked for: the typed payload, and the goldens sidecar
named explicitly as a **deliberate divergence** from TrustSC's package-field arrangement — with the
package-versus-goldens cross-check made an acceptance criterion for the baker and #16. That check
currently has no owner at all: #197's text assigns it to #196, ADR-012's consequences assign it to
#197, and #196 necessarily shipped without it because no package existed to compare against.

Part of #197.

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked — #217 merged as `72edb1f`
- **Merge order:** mergeable now, and before the S8 emitters, which should be written against this shape

## Shared registries touched

- [ ] `CMakeLists.txt` (root) — module already registered by #215
- [ ] `FILE_SET CXX_MODULES` lists — no module added, removed or moved
- [ ] `tests/CMakeLists.txt` — suite membership unchanged
- [ ] Schemas under `docs/*/schemas/`
- [ ] Generated indexes
- [ ] Committed artifacts under `generated/` — none exists for screens, which is the point
- [x] None of the above

## Rebase state

- **Rebased onto:** `72edb1f` (current `develop`)

## Verification

- [x] `clang-format` applied to both touched sources.
- [ ] `cmake --preset ninja-gcc && ctest` — not run locally; the macOS toolchain cannot build the
      `std` module.

**CI is the prototype here, and deliberately so.** The parity note said `std::variant` was worth
proving on both toolchains before committing to it, and with no local module build the honest way to
do that is to let the compile legs answer. Two things to watch:

- the namespace-scope `static_assert`s, which now include `kindName()` and `get_if` on a
  `constexpr` node — if `std::variant` is not usable in a constant expression on either toolchain,
  `medui_spec` fails to compile rather than failing a test;
- `governed-throw` on MduXCore. `std::get` is never called, so `__throw_bad_variant_access` should
  not be emitted; the profile forbids `__cxa_throw` and reports library helpers, so a surprise here
  would show up as a report rather than a silent pass.

If either objects, the fallback is an explicit tagged union — a contained change, and not a lesser
design.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

Traceability improves, and no rule about what a screen may say changes: the set of named values the
compiler accepts is exactly what it was before this PR, since the two closed sets were withdrawn.
`requirementOf()` replaces a field every node pretended to have with a function over the five
components that can actually carry one, so a traceability export walks what the dictionary permits
rather than what the record shape allowed.

The safety-relevant rule this unblocks is #17's and #199's: a device can now be given what it needs
to render every component, which the previous shape could not express for four of them. Two
safety-adjacent defects were found in review and are recorded rather than quietly dropped — a
required field that value-initialised to a plausible valid value, and a spelling that named every
out-of-range `SystemEvent` `TriggerHalt`. Both belonged to the withdrawn enums; both are requirements
on #219 so the next attempt cannot reintroduce them.

ADR-012 carries both new decisions. No ADR is contradicted; no certification or compliance claim is
added.

## Summary by CodeRabbit

* **New Features**
  * Added typed specifications for supported screen components, including component-specific settings and source properties.
  * Added tracking for safety-critical and explicitly positioned components.

* **Bug Fixes**
  * Screen packages now reject invalid, incomplete, or unknown component data with clearer validation errors.
  * Golden-reference selection consistently reflects safety-critical and positioned components.

* **Documentation**
  * Documented compiled screen artifact structure and golden-reference acceptance requirements.

* **Tests**
  * Expanded coverage for component types, provenance-based selection, colors, and validation rules.

### DIFF
--- a/docs/adr/ADR-012-compiled-screen-artifacts.md
+++ b/docs/adr/ADR-012-compiled-screen-artifacts.md
@@ -129,6 +129,13 @@ is, *how much* it may draw, and *which* validated token and key it draws with. T
 vertices is a bounded table lookup followed by arithmetic over a known bound, which is exactly what
 a governed, allocation-free runtime can do.
 
+**A node's payload is typed per component.** The package carries one spec type per dictionary entry
+behind a `std::variant`, not a single record with every field on it. The flat form was tried first
+and was lossy: it had nowhere to put a `Clock`'s format, a `NumericDisplay`'s template and source, a
+`StatusIndicator`'s state keys and per-state tints, a `CriticalButton`'s `on_press`, or a
+`VulkanViewport`'s stream, so a device could not have rendered four of the eleven components. Every
+field remains a validated *name* rather than a resolved value, which keeps the boundary ADR-011 fixes.
+
 ### 3. Generated C++ is emitted, never committed
 
 `<binary>/mdux_generated/screen/<identifier>.cppm` and `.hpp`, produced from the committed
@@ -151,6 +158,21 @@ They have a different consumer (#16's verifier, not the runtime), a different li
 different review audience — a reviewer checking that a safety-critical node's expected bounds
 changed deliberately should not have to find that hunk inside the whole screen. A separate file also
 gets its own digest in `report.json`, so a goldens-only change is visible as one.
+
+**This is a deliberate divergence from TrustSC, recorded as one.** Its `CompiledScreenPackage`
+carries `golden_references` as a field, and `trustsc-ui-verify::verify_frame()` walks them straight
+out of the compiled screen; the field is not feature-gated, so a shipped binary carries expectations
+only a verifier reads. MduX keeps them outside the package for the reasons above — the runtime never
+reads them, the review audience differs, and the sidecar gets its own digest — and pays for that with
+a verifier that opens two files instead of walking one structure. That is a driver-shaped cost, not
+an evidence-shaped one, and it is the trade this decision accepts knowingly rather than by omission.
+
+**The consistency test is an acceptance criterion for the baker and for #16**, not an aspiration.
+Neither file alone can carry it: applying ADR-011's predicate to the compiled nodes yields exactly
+the id set `goldens.json` must contain, so the check is a set comparison across the pair, and #198 is
+the first place both files exist at once. Until then it has no home — #197's text assigns it to #196,
+this ADR's consequences assign it to #197, and #196 necessarily shipped without it because no package
+existed to compare against.
 
 ### 5. One recipe per screen, under `recipes/screen/<id>.toml`
 
--- a/docs/adr/ADR-012-compiled-screen-artifacts.md
+++ b/docs/adr/ADR-012-compiled-screen-artifacts.md
@@ -167,29 +167,26 @@ reads them, the review audience differs, and the sidecar gets its own digest —
 a verifier that opens two files instead of walking one structure. That is a driver-shaped cost, not
 an evidence-shaped one, and it is the trade this decision accepts knowingly rather than by omission.
 
-**The consistency test is an acceptance criterion for #198's baker**, not an aspiration, and #16
-derives its expectations the same way rather than trusting the sidecar to be complete.
+**Completeness is guaranteed by construction, not by comparing the two files.** The predicate has
+one implementation - `collectGoldens()` (#196) - and the baker applies it once, in the same pass that
+writes the compiled nodes. That is TrustSC's arrangement: its DSL compiler pushes a golden at the
+moment it compiles a selected node, from the AST it already holds, and the guard is a set of unit
+tests over known screens rather than a check over the emitted artifact.
 
-It needs something the package must therefore carry. ADR-011's predicate reads two facts about the
-*source* — whether a node was annotated `@safety_critical`, and whether its `position:` was explicit
-— and compilation erases both: every compiled node has resolved bounds, and `requirement` is
-mandatory on `CriticalButton`, `NumericDisplay` and `StatusIndicator` whether or not anyone
-annotated them, so it is not a proxy for the annotation. Without those two facts a checker can
-confirm that every listed golden resolves to a node and still not see the failure that matters: a
-node that should have a golden and does not.
+An earlier revision of this decision required an artifact-level set comparison instead, and carried
+`NodeProvenance { safetyCritical, positioned }` in every compiled node so the expected id set could be
+re-derived from `package.json`. Both are withdrawn, for two reasons that point the same way:
 
-So each compiled node carries `NodeProvenance { safetyCritical, positioned }` — two booleans the
-runtime never reads, for a check that is otherwise impossible. That is the same trade this decision
-refused for the goldens themselves, and it is granted here because the sizes and the consequences
-differ: an entire expectation list against two bits, and a check that could not be written at all
-against one that would merely have been more convenient.
+- **It could not catch the failure that matters.** Both files come from one AST in one pass, so a
+  comparison between them detects two writers disagreeing, never a predicate that is wrong. The
+  tests that *do* catch a wrong predicate already exist, with the predicate, in #196.
+- **It put data on a device that only a tool reads**, which is what this decision refuses two
+  paragraphs above. Two booleans is a small breach of a rule, and a rule breached for something small
+  is how the next thing gets in.
 
-**One owner: #198.** It is the first stage where both files exist at once, so it is the first place
-the test can run. #197's issue text assigned the cross-check to #196 and an earlier revision of this
-ADR assigned it to #197; #196 necessarily shipped without it, because no package existed to compare
-a golden set against. Three owners for a safety-relevant criterion is how a requirement is lost at a
-handoff, so the other two references are corrected rather than left to be reconciled by whoever
-notices.
+What replaces the check is a constraint on the baker (#198): **it must call `collectGoldens()`, not
+re-apply the predicate.** A second implementation is the only way the two files can disagree once
+they are written in one pass, and it is the thing a reviewer should refuse.
 
 ### 5. One recipe per screen, under `recipes/screen/<id>.toml`
 
@@ -253,28 +250,21 @@ human should read.
   against. The mitigation is host-side — the baker writes both files from one AST, and #198 owns
   the test.
 
-  That test compares **sets, not references**. Applying ADR-011's predicate to the nodes in
-  `package.json` yields exactly the ids `goldens.json` must contain, so the test derives that set
-  and requires equality: an id in the goldens with no node behind it fails, and — the direction
-  that matters more — a node the predicate selects with no golden fails too. Checking only that
-  listed ids resolve would accept the dangerous case silently, since a safety-critical node whose
-  golden was dropped looks exactly like a screen with fewer safety-critical nodes. #16's verifier
-  derives its expectations the same way rather than trusting the file to be complete.
+  The mitigation is structural rather than a comparison: one implementation of ADR-011's predicate,
+  called once, in the pass that writes both files. An earlier revision of this ADR required the baker
+  to derive the expected id set from `package.json` and assert set equality with `goldens.json`, and
+  decision 4 records why that was withdrawn — it cannot see a wrong predicate, and it needed
+  provenance on the device that only a tool would read. #16 derives its expectations from
+  `goldens.json` itself; what it must never do is re-apply the predicate, since a second
+  implementation is precisely what would let the two disagree.
 
 ### Risks
 - **The package grows to carry things the runtime does not need**, because it is the convenient
   place to put them. *Mitigation*: the rule to apply is that `package.json` holds what the runtime
-  reads, plus exactly one named exception; anything else read only by a tool belongs in a sidecar,
-  as the goldens do.
+  reads; anything read only by a tool belongs in a sidecar, as the goldens do. Decision 4 records one
+  attempt to breach this rule - two booleans of golden provenance - and why it was withdrawn rather
+  than granted an exception.
 
-  **The exception is `NodeProvenance`, and it is bounded on purpose.** Decision 4 admits two
-  booleans per node that the runtime never reads, because the golden-completeness check cannot be
-  written without them and no sidecar can carry them — a file cannot supply the input that proves
-  that same file is complete. Naming it here rather than leaving it as an unremarked contradiction
-  is what stops a later reader applying the general rule and deleting the check's only input. The
-  test for a further exception is the one decision 4 applies: it has to be a check that is otherwise
-  *impossible*, not one that would merely be more convenient, and it has to cost bits rather than
-  lists.
 - **A regenerated screen differs on a second toolchain** despite the integer-only layout rule.
   *Mitigation*: `evidence.screen.<id>` runs on both legs, which is the check that would catch it;
   ADR-011's integer-only rule is what makes it expected to pass.
--- a/docs/adr/ADR-012-compiled-screen-artifacts.md
+++ b/docs/adr/ADR-012-compiled-screen-artifacts.md
@@ -264,7 +264,17 @@ human should read.
 ### Risks
 - **The package grows to carry things the runtime does not need**, because it is the convenient
   place to put them. *Mitigation*: the rule to apply is that `package.json` holds what the runtime
-  reads; anything read only by a tool belongs in a sidecar, as the goldens do.
+  reads, plus exactly one named exception; anything else read only by a tool belongs in a sidecar,
+  as the goldens do.
+
+  **The exception is `NodeProvenance`, and it is bounded on purpose.** Decision 4 admits two
+  booleans per node that the runtime never reads, because the golden-completeness check cannot be
+  written without them and no sidecar can carry them — a file cannot supply the input that proves
+  that same file is complete. Naming it here rather than leaving it as an unremarked contradiction
+  is what stops a later reader applying the general rule and deleting the check's only input. The
+  test for a further exception is the one decision 4 applies: it has to be a check that is otherwise
+  *impossible*, not one that would merely be more convenient, and it has to cost bits rather than
+  lists.
 - **A regenerated screen differs on a second toolchain** despite the integer-only layout rule.
   *Mitigation*: `evidence.screen.<id>` runs on both legs, which is the check that would catch it;
   ADR-011's integer-only rule is what makes it expected to pass.
--- a/docs/adr/ADR-012-compiled-screen-artifacts.md
+++ b/docs/adr/ADR-012-compiled-screen-artifacts.md
@@ -167,12 +167,29 @@ reads them, the review audience differs, and the sidecar gets its own digest —
 a verifier that opens two files instead of walking one structure. That is a driver-shaped cost, not
 an evidence-shaped one, and it is the trade this decision accepts knowingly rather than by omission.
 
-**The consistency test is an acceptance criterion for the baker and for #16**, not an aspiration.
-Neither file alone can carry it: applying ADR-011's predicate to the compiled nodes yields exactly
-the id set `goldens.json` must contain, so the check is a set comparison across the pair, and #198 is
-the first place both files exist at once. Until then it has no home — #197's text assigns it to #196,
-this ADR's consequences assign it to #197, and #196 necessarily shipped without it because no package
-existed to compare against.
+**The consistency test is an acceptance criterion for #198's baker**, not an aspiration, and #16
+derives its expectations the same way rather than trusting the sidecar to be complete.
+
+It needs something the package must therefore carry. ADR-011's predicate reads two facts about the
+*source* — whether a node was annotated `@safety_critical`, and whether its `position:` was explicit
+— and compilation erases both: every compiled node has resolved bounds, and `requirement` is
+mandatory on `CriticalButton`, `NumericDisplay` and `StatusIndicator` whether or not anyone
+annotated them, so it is not a proxy for the annotation. Without those two facts a checker can
+confirm that every listed golden resolves to a node and still not see the failure that matters: a
+node that should have a golden and does not.
+
+So each compiled node carries `NodeProvenance { safetyCritical, positioned }` — two booleans the
+runtime never reads, for a check that is otherwise impossible. That is the same trade this decision
+refused for the goldens themselves, and it is granted here because the sizes and the consequences
+differ: an entire expectation list against two bits, and a check that could not be written at all
+against one that would merely have been more convenient.
+
+**One owner: #198.** It is the first stage where both files exist at once, so it is the first place
+the test can run. #197's issue text assigned the cross-check to #196 and an earlier revision of this
+ADR assigned it to #197; #196 necessarily shipped without it, because no package existed to compare
+a golden set against. Three owners for a safety-relevant criterion is how a requirement is lost at a
+handoff, so the other two references are corrected rather than left to be reconciled by whoever
+notices.
 
 ### 5. One recipe per screen, under `recipes/screen/<id>.toml`
 
@@ -233,7 +250,7 @@ human should read.
   nothing in the format prevents them diverging. The check cannot live in the governed
   `validate()`: decision 4 puts the goldens outside the schema precisely because the runtime never
   reads them, so the `static_assert` in §3 sees only `package.json` and has no ids to compare
-  against. The mitigation is host-side — the baker writes both files from one AST, and #197 owns
+  against. The mitigation is host-side — the baker writes both files from one AST, and #198 owns
   the test.
 
   That test compares **sets, not references**. Applying ADR-011's predicate to the nodes in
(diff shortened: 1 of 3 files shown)


docs/adr/ADR-012-compiled-screen-artifacts.md
```diff
@@ -152,6 +159,21 @@ different review audience — a reviewer checking that a safety-critical node's
...
 
+**This is a deliberate divergence from TrustSC, recorded as one.** Its `CompiledScreenPackage`
+carries `golden_references` as a field, and `trustsc-ui-verify::verify_frame()` walks them straight
+out of the compiled screen; the field is not feature-gated, so a shipped binary carries expectations
+only a verifier reads. MduX keeps them outside the package for the reasons above — the runtime never
+reads them, the review audience differs, and the sidecar gets its own digest — and pays for that with
+a verifier that opens two files instead of walking one structure. That is a driver-shaped cost, not
+an evidence-shaped one, and it is the trade this decision accepts knowingly rather than by omission.
+
+**The consistency test is an acceptance criterion for the baker and for #16**, not an aspiration.
```
ambroise-leclerc: **[P1] Preserve the provenance needed to prove golden completeness.** This acceptance criterion says the expected golden id set is obtained by applying ADR-011’s predicate to the compiled nodes, but the new `CompiledNode` shape retains only `id`, resolved `bounds`, and the component payload. It does not retain whether the source node carried `@safety_critical` or whether its `position:` was explicit. Neither remaining property reconstructs that predicate: every compiled node has resolved bounds, while `requirement` is optional on `Button`/`TextInput` and mandatory on `CriticalButton`/`NumericDisplay`/`StatusIndicator` independently of a safety annotation. As a result, a checker reading `package.json` and `goldens.json` cannot distinguish a valid sidecar from one that silently omitted a required golden—the dangerous direction this paragraph says must fail. Please preserve the selection provenance in the compiled package (for example, explicit flags or an equivalent closed marker), or define and commit another independent input from which both the baker check and #16 can derive the expected set. The ADR should then describe that implementable mechanism.

include/mdux/medui/Schema.cppm
```diff
@@ -341,6 +552,147 @@ struct ScreenPackage {
...
+        for (const std::string_view token : spec->colorTokens) {
+            if (const auto colour = requireColor(token); !colour.has_value()) {
+                return colour;
+            }
+        }
+        return {};
+    }
+    const auto* spec = std::get_if<TextInputSpec>(&payload);
+    if (spec == nullptr) {
+        return {};
```
ambroise-leclerc: **[P2] Fail closed for an unrecognized or valueless payload.** The residual `get_if<TextInputSpec>` branch returns success when it gets `nullptr`. Current generated construction normally makes that unreachable, but it is still a fail-open validation contract: a `NodePayload` made valueless through an exceptional emplacement, or a future alternative added without updating this function, is accepted as a valid node. `kindName()` has the parallel risk of labeling any future residual alternative as `TextInput`. Because `ScreenPackage::validate()` is the governed compile-time/runtime boundary that consumers are entitled to trust, please make dispatch exhaustive at compile time where possible and return a dedicated schema error for a genuinely valueless payload. Any approach should preserve the no-throw policy; a plain potentially-throwing `std::visit` is not automatically suitable here.

docs/adr/ADR-012-compiled-screen-artifacts.md
```diff
@@ -152,6 +159,21 @@ different review audience — a reviewer checking that a safety-critical node's
...
+Neither file alone can carry it: applying ADR-011's predicate to the compiled nodes yields exactly
+the id set `goldens.json` must contain, so the check is a set comparison across the pair, and #198 is
+the first place both files exist at once. Until then it has no home — #197's text assigns it to #196,
```
ambroise-leclerc: **[P2] Assign the consistency test to one issue and update every reference.** This paragraph makes #198 the first stage where both files exist, but the Consequences section still says “#197 owns the test,” while issue #197 itself says the cross-check is a host-side test in #196. That leaves three owners for a safety/traceability acceptance criterion and makes omission during handoff plausible. Please choose the actual implementation owner—#198 appears consistent with the reasoning added here—and update this paragraph, the Consequences section, and the relevant issue text so the requirement has one discoverable owner.

ambroise-leclerc (commented):
Detailed review completed. I recommend addressing the three inline findings before merge. The golden-selection provenance issue is the blocking concern because the documented completeness check cannot be implemented from the package shape introduced here; the other two findings concern fail-closed validation and unambiguous verification ownership. All current GitHub checks are green, and local documentation lint plus clang-format verification passed.

docs/adr/ADR-012-compiled-screen-artifacts.md
ambroise-leclerc: This is the finding of the round, and you are right: **I wrote an acceptance criterion that could not be implemented.** Fixed in `335903f`.

The paragraph said the expected golden id set comes from applying ADR-011's predicate to the compiled nodes. The predicate reads two facts about the *source* — annotated `@safety_critical`, and an explicit `position:` — and compilation erases both. Your two counters are exactly the ones that matter: every compiled node ends up with resolved bounds, and `requirement` is mandatory on `CriticalButton`, `NumericDisplay` and `StatusIndicator` independently of any annotation, so it is not a proxy for it either. A checker could have confirmed that every listed golden resolves to a node and still missed the omission — the dangerous direction the paragraph itself says must fail.

`CompiledNode` now carries `NodeProvenance { safetyCritical, positioned }`, with `selectsGolden()` beside it so the predicate has one spelling rather than one per consumer.

Two things I want to be explicit about rather than let pass as obvious:

**This is the trade ADR-012 decision 4 refused**, for goldens. Granting it here needs an argument, not a shrug, and the one I put in both the module and the ADR is: sizes and consequences differ. An entire expectation list against two bits; and a check that could not otherwise be written at all against one that would merely have been more convenient. If you read that as decision 4 leaking, say so — the alternative is committing a third independent input, and I would rather hear that objection now than after #198 bakes.

**The gap predates this PR**, and that is not a defence. The flat node carried `component`, `textKey`, `colorToken`, `requirement` and no provenance either. What this PR did was write down an acceptance criterion whose inputs the schema did not carry, which is how a plausible-sounding requirement becomes unimplementable at handoff — and it took your reading of the two together to see it.

include/mdux/medui/Schema.cppm
ambroise-leclerc: Agreed, and fixed in `335903f`. Copilot flagged the same branch independently, which is a fair signal that "unreachable today" was doing too much work in my head.

- `validatePayload()` returns `SchemaError::UnknownPayload` for a payload it cannot name, instead of success.
- `kindName()` names `TextInput` because the payload *is* a `TextInputSpec`, not because it is the last branch. Anything else returns an empty name — a package that acquires a plausible wrong answer is worse than one that admits it has none.
- Exhaustiveness is now a build failure rather than a runtime discovery: `static_assert(std::variant_size_v<NodePayload> == 11)` sits next to both functions, so adding an alternative without teaching them about it stops the build at the point of the change.

On `std::visit`: you are right that it is not automatically suitable, and I did not reach for it — it throws `bad_variant_access` on a valueless variant, which is exactly the accessor shape ADR-005 rules out for a governed type. The `static_assert` buys the exhaustiveness that `visit` would have provided, without the throwing path.

docs/adr/ADR-012-compiled-screen-artifacts.md
ambroise-leclerc: Fixed in `335903f`, in all three places rather than two.

#198 is the owner: it is the first stage where `package.json` and `goldens.json` exist at once, so it is the first place the test can run at all. Updated accordingly:

- the paragraph I added in decision 4;
- this ADR's Consequences section, which said #197;
- **#197's own issue text**, which said #196 — with a short note recording why it moved, since #196 could not have carried it: no package existed then to compare a golden set against.

Your framing is the part worth keeping: three owners for a safety-relevant criterion is not an inconsistency to tidy, it is how a requirement is lost at a handoff. Each of the three references looked locally reasonable to whoever wrote it.

tests/medui/SchemaTests.cpp
```diff
@@ -31,24 +31,14 @@ namespace ms = mdux::medui;
...
-                     .colorToken  = "Theme.Colors.Title",
-                     .requirement = {}          },
-    ms::CompiledNode{            .id          = "score",
-                     .component   = "NumericDisplay",
-                     .bounds      = {250, 60, 120, 60},
-                     .textKey     = {},
-                     .colorToken  = "Theme.Colors.ScoreDigits",
-                     .requirement = "REQ-NS-001"}
+    ms::CompiledNode{.id = "topbar-background",.bounds = {0, 0, 400, 40},.payload = ms::PanelSpec{.colorToken = "Theme.Colors.TopbarBackground"}                                                                          },
+    ms::CompiledNode{            .id = "title", .bounds = {8, 8, 200, 24}, .payload = ms::LabelSpec{.textKey = "STR-TITLE", .colorToken = "Theme.Colors.Title"}},
```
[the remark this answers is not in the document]
ambroise-leclerc: Right, and worth recording why the tool did not catch it: `clang-format --dry-run --Werror` reports the file as **clean**. That block *is* clang-format's output — its alignment of long designated initialisers pads the node list into something no reviewer can scan, which is the formatter being satisfied while the reader is not.

So the fix is to shorten the source rather than reformat it: the three payloads are now named `constexpr` values above the array, and the node list is one readable line per node. Same formatter, same settings, legible result.

docs/adr/ADR-012-compiled-screen-artifacts.md
```diff
@@ -152,6 +159,38 @@ different review audience — a reviewer checking that a safety-critical node's
...
+mandatory on `CriticalButton`, `NumericDisplay` and `StatusIndicator` whether or not anyone
+annotated them, so it is not a proxy for the annotation. Without those two facts a checker can
+confirm that every listed golden resolves to a node and still not see the failure that matters: a
+node that should have a golden and does not.
+
+So each compiled node carries `NodeProvenance { safetyCritical, positioned }` — two booleans the
+runtime never reads, for a check that is otherwise impossible. That is the same trade this decision
+refused for the goldens themselves, and it is granted here because the sizes and the consequences
+differ: an entire expectation list against two bits, and a check that could not be written at all
+against one that would merely have been more convenient.
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `355019b`, and the finding is sharper than "document the exception": as written, the ADR **contradicted itself in the dangerous direction**. The Risks rule authorised deleting anything the runtime does not read, which now includes `NodeProvenance` — and deleting it would not break a build or fail a test. The golden-completeness check would simply stop being able to detect an omitted golden, silently, which is the exact failure decision 4 says must not be silent.

The rule now carries exactly one named exception, plus the test for admitting another: it has to be a check that is otherwise *impossible* rather than merely more convenient, and it has to cost bits rather than lists.

One point I added beyond the suggested wording, because it is what makes this exception genuinely narrow rather than a precedent: **no sidecar can carry this particular data**, since a file cannot supply the input that proves that same file is complete. That is why the general "tool-only data goes in a sidecar" rule cannot absorb it, and it is the boundary a future reader needs in order to refuse the next request to put something convenient in the package.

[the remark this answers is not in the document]
ambroise-leclerc: Thanks — but that thread has been overtaken, and the learning recorded from it is now the opposite of what the ADR says. Flagging it here so the record is not left asserting a withdrawn rule.

`NodeProvenance` is **gone**, and with it the artifact-level set comparison it existed to serve (`52b370b`). Reading TrustSC's compiler settled it: its DSL applies the predicate once, at the moment it compiles a node, pushes the golden into the same structure, and guards completeness with unit tests over known screens — no provenance in the compiled node, and nothing re-derived from an artifact afterwards.

Two reasons pointed the same way:

- the comparison could not catch the failure that matters. Both files come from one AST in one pass, so it detects two writers disagreeing, never a wrong predicate — and the tests that *do* catch a wrong predicate already exist, with the predicate, in #196;
- it put data on a device that only a tool reads, which is what ADR-012's own risk mitigation refuses. Two booleans is a small breach, and a rule breached for something small is how the next thing gets in.

So the named exception you and I converged on is withdrawn rather than kept: the mitigation now records the attempt and its refusal, and the guarantee is structural — one implementation of the predicate, called once, with #198 required to *call* `collectGoldens()` rather than re-apply it.

The reasoning in your learning was sound for the design as it stood; the design changed under it.

include/mdux/medui/Schema.cppm
```diff
@@ -277,27 +292,312 @@ struct NodeRect {
...
+
+struct LabelSpec {
+    std::string_view textKey{};
+    std::string_view colorToken{};
+
+    [[nodiscard]] constexpr bool operator==(const LabelSpec&) const noexcept = default;
+};
+
+struct ClockSpec {
+    ClockFormat format{};
```
ambroise-leclerc: **[P1] Required enum fields now silently default to valid behavior.** `ClockSpec::format{}` value-initializes to the first enumerator, `ClockFormat::TimeSeconds`; `CriticalButtonSpec::onPress{}` likewise becomes `SystemEvent::NoOp`. `validatePayload()` then deliberately performs no format check and no event check. Therefore `ClockSpec{}` and an otherwise-complete `CriticalButtonSpec` whose emitter accidentally omitted `onPress` both validate, even though `format` and `on_press` are required by the component dictionary. This weakens the previous empty-name validation into a plausible valid value and defeats the generated `static_assert(package.validate())` that is supposed to catch emitter defects. Please make omission representable and rejected—for example with an explicit `Unspecified` sentinel rejected by validation—or require construction of these fields rather than defaulting required members. Add focused tests proving that both omissions fail.

include/mdux/medui/Schema.cppm
```diff
@@ -277,27 +292,312 @@ struct NodeRect {
...
+ * resolve against tables a product supplies (#195). TrustSC closes the same set in its governed
+ * crate, and the parity criterion decides it; what makes the exception defensible on its own terms
+ * is that a closed set lets the *compiler* know what a clock renders - `TimeSeconds` is `HH:MM:SS`,
+ * eight glyphs of known width - and therefore lets it measure a clock against its bounds. An open
+ * name cannot be measured, only looked up.
+ *
+ * `charset` on a `TextInput` deliberately stays a name: TrustSC leaves its equivalent
+ * (`glyph_set_id`) open too, because a glyph set is a product's to define.
+ */
+enum class ClockFormat : std::uint8_t {
```
ambroise-leclerc: **[P1] Coordinate this closed-set change with the implemented front end.** The current semantic analyzer accepts any identifier for `Clock.format` and `CriticalButton.on_press`, while the text-budget stage intentionally resolves clock formats through a product-supplied `DynamicTextRule` table. Existing tests exercise accepted names including `HH_MM`, `HH_MM_TZ`, `DIGITS_TO_A`, `TimeFormat.HoursMinutes`, and `SystemEvent.Stop`; none can be represented by these two-value enums. As a result, a screen accepted by the implemented parser/semantic/budget pipeline can reach the future emitter with no valid `NodePayload` representation. Because this module is intended to be the canonical type imported by that compiler, please update the semantic domains, budget logic, fixtures, authoring guidance, and relevant ADR decision together—or retain string names here until that coordinated consumer change lands in a declared successor.

include/mdux/medui/Schema.cppm
```diff
@@ -277,27 +292,312 @@ struct NodeRect {
...
+ * that out. A product's own actions reach the device through `source:` and the realtime path, not
+ * through this.
+ */
+enum class SystemEvent : std::uint8_t {
+    NoOp,
+    TriggerHalt,
+};
+
+/// The spelling the emitter writes and a reader recognises, e.g. `TimeSeconds`.
+[[nodiscard]] constexpr std::string_view spell(ClockFormat format) noexcept {
```
ambroise-leclerc: **[P1] Reject invalid enum representations instead of aliasing them to valid values.** A fixed-underlying-type C++ enum can hold an out-of-range value through a cast or malformed/manual package construction. These binary conditionals map every invalid `ClockFormat` to `DateTimeSeconds` and every invalid `SystemEvent` to `TriggerHalt`; `validatePayload()` also accepts both without checking membership. That means a malformed critical-button action can acquire the spelling and apparent behavior of `TriggerHalt`, directly contradicting this correction’s fail-closed goal. Please use exhaustive switches that return an empty/error result for the default case, validate both enum fields in `validatePayload()`, and cover out-of-range values with constexpr tests.

include/mdux/medui/Schema.cppm
```diff
@@ -277,27 +292,312 @@ struct NodeRect {
...
+ * revision of this module did exactly that, and it was worse than untidy: it was **lossy**. A flat
+ * node had nowhere to put a `Clock`'s format, a `NumericDisplay`'s template and source, a
+ * `StatusIndicator`'s state keys and per-state tints, a `CriticalButton`'s `on_press`, or a
+ * `VulkanViewport`'s stream - so a device holding the package could not have rendered four of the
+ * eleven components at all.
  *
- * `requirement` is empty for a node that declares none, and non-empty for every node that is
- * safety-critical, because #196 refuses the annotation without it. Carrying it here is what makes
- * the requirement-to-node mapping diffable in a committed artifact.
+ * Every field is a **validated name**, never a resolved value, exactly as `colorToken` is. The
```
ambroise-leclerc, line 305: **[P2] Update the public documentation to the design implemented immediately below it.** This block still says every payload field is a validated name and explicitly says `format` and `onPress` remain `std::string_view` rather than enums. The next declarations do the opposite. ADR-012’s typed-payload paragraph and the PR description repeat the stale name-based decision, and the PR’s regulatory-impact section still says no behavior changes. Since this commit reverses #195’s stated boundary for two fields, please update all of those architectural and review-facing statements and describe the coordinated consumer impact honestly.

ambroise-leclerc (commented):
Re-review of the latest correction commits found three blocking issues in the newly introduced closed enums, plus one documentation contradiction. The earlier unknown-variant fail-open is corrected. The golden-completeness concern has been redesigned around a single-pass `collectGoldens()` strategy, so the remaining blockers are the required-field defaults, incompatibility with the implemented authoring pipeline, and invalid enum representations being accepted or misidentified. All latest-head GitHub checks are green.

include/mdux/medui/Schema.cppm
ambroise-leclerc: You are right, and this one decides the other two. Withdrawn in `9b4e4d6`; the coordinated change is now **#219**.

What I got wrong is worth naming precisely: closing the set in the module that the compiler *imports* left the canonical type unable to represent a screen the implemented pipeline accepts. That is the exact drift this schema exists to prevent, introduced by the schema — and it was latent rather than visible, because no emitter exists yet to hit it.

The blocker I hit while writing #219 is the part I could not have resolved inside this PR anyway: **no registered diagnostic reports a named value outside a closed set.** `MEDUI-E071` is specific to `cv_check`, `MEDUI-E053` means dynamic text escaping the baked charset, and `MEDUI-E033` means a value of the wrong *kind* rather than an unknown member of a known set. The shared contract lists "unknown … formats, system events" among the things that must fail compilation and assigns no code; TrustSC uses its own `UNREGISTERED`. Inventing a `MEDUI-E###` is `Compliatory/MedUI`'s decision, so #219 opens with that question rather than with code.

#219 also carries the two findings below, since they are properties of the closure rather than of this PR: the `Unspecified` sentinel, and exhaustive spellings with a membership check.

The module now records why the attempt was made and withdrawn, so the next attempt starts from the constraint instead of rediscovering it.

include/mdux/medui/Schema.cppm
ambroise-leclerc: Correct, and the failure mode is worse than "weaker": it is weaker *in the direction the generated `static_assert` exists to cover*. An omitted name is an empty `string_view` that `require()` rejects; an omitted enum is `{}`, which is the first enumerator — so `ClockSpec{}` meant `TimeSeconds` and a `CriticalButtonSpec` whose emitter forgot `onPress` meant `NoOp`. An emitter defect stopped being a compile error and became a plausible screen.

Withdrawn with the rest in `9b4e4d6`, so the string-name validation is back. The fix you propose — an `Unspecified` first enumerator, rejected by `validate()` — is recorded in #219 as a requirement rather than an option, with the reason it is needed at all: a C++ aggregate value-initialises to the first enumerator, where a Rust struct literal must name every field. The sentinel is what recovers in C++ the property TrustSC gets for free, which is precisely the "design the representation for C++ rather than copying Rust's" instruction from the earlier round.

include/mdux/medui/Schema.cppm
ambroise-leclerc: Right, and this is the finding I am most glad to have had before an emitter existed. `spell(SystemEvent)` was `event == NoOp ? "NoOp" : "TriggerHalt"` — so **every** out-of-range value, from a cast or a hand-built package, would have been named `TriggerHalt`. A malformed action acquiring the spelling and apparent meaning of *halt the system* is not a fail-open detail; it is the worst available answer, on the one field where the wrong answer stops a device.

Withdrawn in `9b4e4d6` along with the enums. #219 carries the fix as a requirement: exhaustive switches returning an empty name for the default case, membership checked in `validatePayload()`, and a `constexpr` test over a cast out-of-range value — so the representation is proven not to alias rather than assumed not to.

include/mdux/medui/Schema.cppm
ambroise-leclerc: Fixed by the withdrawal, and I want to be exact about the state rather than claim more than is true.

The module comment now describes what the code does — names for all three fields — and, because a reader who finds this in six months deserves the reason, it also records that closing two of them was attempted and withdrawn, with the constraint that stopped it and a pointer to #219.

ADR-012's typed-payload paragraph and its "validated names throughout" statement are accurate again as written, so neither needed amending after the revert; #219's work list includes them, since closing the sets is exactly what will make them stale.

The PR description's regulatory-impact section is the one place your point still stands: it claimed no behaviour change while the enums were in. I am updating it now to describe what this PR actually does after the withdrawal — typed payloads, and no change to which named values the compiler accepts.

AM-L (approved):

Issue #153, opened 2026-08-03
mdux.ml: constexpr package emitter, so a device build stops parsing package.json


ambroise-leclerc:
Follow-up to #18, deliberately left out of scope there. Shaped like the shader pipeline's #121, which did the same job for shader packages.

## The gap

`mdux.ml.runtime` takes a `ModelPackage` — a `constexpr`-constructible view over spans — and has never contained a parser. That is the design, and ADR-008 decision 2 leans on it: the runtime performs no I/O, which is what makes it usable on a device with no filesystem.

Today there is only one way to *get* such a package: `tools/ml/PackageLoad.cppm` parses the committed `package.json` at startup. It lives in the host-tools zone precisely so the governed runtime keeps having no parser in it, and `EcgClassifierExample` embeds the JSON as a blob so nothing is read from disk.

That leaves a device build parsing JSON at startup and linking a host-tools module to do it. It is honest but it is not the end state.

## What to build

A `constexpr` emitter turning a committed `package.json` into generated source:

```cpp
// generated
inline constexpr std::array<mdux::ml::LayerDesc, 5> ecgDemoLayers{ /* ... */ };
inline constexpr std::array<std::uint32_t, 180> ecgDemoGolden0Input{ /* ... */ };
inline constexpr mdux::ml::ModelPackage ecgDemoPackage{ /* ... */ };
static_assert(ecgDemoPackage.validate().has_value());
```

`mdux.ml.schema` is already header-only and fully `constexpr` for exactly this — `tests/ml/SchemaTests.cpp` already `static_assert`s a reference package validating at compile time, so the hard part is done.

**Weights stay out of it.** They remain a caller-supplied blob via `mdux_embed_blob()`. A multi-megabyte `constexpr` array takes minutes to compile and can exhaust MSVC — that is ADR-008 decision 2 and it does not change here. Only the *package* becomes source.

## Why it is worth doing

- A device build links no host-tools module and runs no parser at startup — the trust-zone story becomes structural rather than conventional.
- `validate()` runs at **compile time**, so a malformed package is a build error rather than a startup failure.
- The package lands in `.rodata` with no startup allocation at all.

## Acceptance

- [ ] An `emit` path producing a `.cppm` and a header fallback, mirroring #121's two-form arrangement
- [ ] `EcgClassifierExample` switched to the generated package; it should stop linking `MduX::MlBakeLib`
[...]

PR #221, opened 2026-08-21
Emit a compiled screen as constexpr C++ [#197]

### LINKED ISSUE #197: S8 — Canonical package and C++ emitters

ambroise-leclerc:
Second of two for S8, and the half that closes #197: a committed screen package becomes C++ a device
links, with the schema check happening at compile time rather than at startup.

## What it adds

- **`mdux.tools.medui.emit`** - renders a package as a module interface and a header, both from one
  in-memory body, so they cannot describe different screens.
- **`mdux-screenemit`** - its own executable rather than a mode of the compiler, for the reason
  `mdux-shaderemit` is its own: compiling a source and rendering a committed package are different
  jobs, and the second must stay runnable on an artifact whose source is not present.
- **`cmake/MduXScreenEmit.cmake`** - `mdux_emit_screen_package()` and `mdux_screen_identifier()`,
  following `mdux_emit_shader_package()` as ADR-012 decision 3 requires.

## The static_assert is the point

The generated source carries `static_assert(screen.validate().has_value())`. A screen that does not
satisfy its schema is a **build failure in whatever links it** - not a startup failure on a device,
and not a diagnostic from a tool somebody might not have run. That is the whole reason
`mdux.medui.schema` had to be header-only and fully `constexpr`, which #197 delivered rather than
inherited.

It is a genuine second gate rather than a restatement: `readPackage()` validated what it read, so
this is the *compiler* re-checking the *emitted* form. If the rendering ever dropped a required name
or mis-spelled a colour token, the emitted screen would stop validating and the build would say so
at the file carrying the mistake.

## The acceptance criterion, and what compiling proves on its own

`medui_spec` compiles both emitted forms into one binary and asserts they agree - id, surface,
budget, node for node. Two thin translation units per form plus a third that compares them, because
the module form is attached to a named module and the header form is not, so a single TU that both
imported and included would hold two distinct sets of entities with the same names.

Two properties are proved by those files existing at all, before a scenario runs:

- The `static_assert` is evaluated when they compile, so a malformed screen could not have got that
  far.
- `medui_spec` links `MduX::Core` and **no host-tools module**, so a device build reaching a compiled
  screen through generated code links no compiler and parses nothing at startup - the end state #153
  is pursuing for every package kind. That target's link line is the evidence.

## Two things I would draw your eye to

**The filename rule exists in two languages, and both halves are executed.** `mdux_screen_identifier()`
and `identifierForScreen()` must agree or the build looks for a file nobody wrote.
`medui-screen-identifier-parity` reads the answers CMake wrote at configure time and compares them
against the C++ ones, rather than asserting the C++ half and calling it consistent - which is exactly
how the shader pair drifted on the leading-digit rule.

It is also now a rule with **four implementations** across the two emitters and their two CMake
functions. Promoting it to `ToolsCommon` would leave one C++ definition and one CMake definition; I
did not do it here because it means editing merged, tested shader code inside a PR about screens.
Happy to do it as a follow-up if you want it.

**The emission reads a fixture, not an artifact under `generated/`.** Nothing bakes a screen there
until #198 registers one, so `tests/medui/fixtures/every-component-package.json` is committed as the
emitter's input. The obvious hazard is that a hand-maintained package drifts from what the compiler
would produce - so `medui-screen-emit-fixture-is-current` recompiles the fixture's `.medui` source
through parse, layout and `buildPackage()`, and compares bytes. If the layout solver ever changes a
rectangle, that scenario fails rather than the emitter quietly rendering a screen nobody wrote.

## Tests

- `tests/medui/GeneratedScreenTests.cpp` (in `medui_spec`) - the two forms agree; the four components
  a flat node could not have held read back out of generated code; five nodes carry a requirement, so
  a traceability export has something to walk.
- `tests/medui/EmitTests.cpp` (in `medui_tools_spec`) - the fixture is current, both halves of the
  filename rule agree, both outputs share one body, all eleven payloads are rendered, a package the
  schema refuses is not rendered and reports the reader's own `SCP005`, a missing package reports
  `SCE001`, and writing twice leaves the files unchanged.

> ### Review round: three ways a valid screen became invalid C++
>
> All three findings had one shape - the compiler accepted a screen, the schema called it valid, and
> the emitter rendered C++ that would not build - so the description above now describes what the
> emitter does after `8debe52`:
>
> - **A name carrying a control character.** The lexer decodes `\n` and `\t` inside a source string
>   and the dictionary accepts an unrestricted `String` for `requirement`, `source`, `stream_source`
>   and `template`, so `requirement: "REQ\nHALT"` is legal and its package valid. The old escaper
>   handled `"` and `\` only, so the decoded newline ended the literal and what followed it was read
>   as C++ tokens - a code-injection shape in a code generator. Control bytes are now escaped, as
>   three-digit octal rather than `\xNN`, because a hex escape is greedy.
> - **The empty screen.** `validate()` permits no nodes with an empty budget, and an empty C array
>   cannot deduce a zero bound, so that screen was unusable in both forms. It renders as an empty
>   span now, and a second emit target compiles it.
> - **Keyword ids.** `class`, `namespace`, `module` and `import` are legal slugs that mapped to
>   themselves, giving a namespace and module name no compiler accepts - and *both* implementations
>   agreed on that invalid answer, so the parity test reported agreement on something unusable.
>   Identifiers are prefixed unconditionally (`screen_class`), which also retires the leading-digit
>   rule, and keywords are in the parity list. An id mapping to a reserved `__` is refused on both
>   sides rather than rewritten, since collapsing the run would let two screens claim one filename.

## A GCC defect this turned up, and the one governed change it forced

The generated `static_assert` failed to compile on the sanitizer leg while every other configuration
accepted it. Reduced to a 30-line reproduction rather than guessed at, and the trigger is narrow:

| configuration | result |
|---|---|
| `inline constexpr` screen, no sanitizer | compiles |
| `inline constexpr` screen, `-fsanitize=address` | compiles |
| `inline constexpr` screen, `-fsanitize=undefined` | **rejected** |
| same screen, `constexpr` without `inline` | compiles |
| same screen, `inline` in an anonymous namespace | compiles |
| payload passed by value, `inline` kept | compiles |

GCC 16.1 under `-fsanitize=undefined` refuses `&object != nullptr` - the comparison inside
`std::get_if` - as a constant expression when the object is a subobject of an `inline` variable with
external linkage. That is precisely the shape generated code has. The standard is not ambiguous: the
address of an object with static storage duration is a core constant expression whatever instruments
the translation unit, so this is a compiler defect rather than a rule to design around.

**`validatePayload()` and `requirementOf()` therefore take the payload by value.** It is a governed
change, so it is documented at length where it lives, with what breaks without it. The copy costs
nothing that matters - every alternative is trivially copyable, so it neither allocates nor throws -
and it fixes the property for *any* consumer shape rather than only for the emitter's output, which
is why I preferred it to emitting non-`inline` data.

The pin for it is in a **named** namespace in `SchemaTests.cpp`, and that detail is the interesting
one: `inline` in an anonymous namespace does not reproduce the defect, so a pin written beside its
neighbours would have asserted nothing while looking like it asserted something.

## Regulatory impact

The change is where a malformed screen stops being possible. Before this PR a compiled screen was
JSON that something had to read correctly at startup; after it, the screen is `constexpr` data whose
validity is a compile-time property of the binary, and the tool that renders it refuses to emit a
screen the schema rejects. No generated code is committed - the reviewed artifact remains the JSON.

One governed file changes: `include/mdux/medui/Schema.cppm`, for the compiler defect above. The
change is a parameter passing mode, not a rule: the same names are required, the same colours
resolved, the same states paired, and the accepted and rejected screens are exactly what they were.
`governed.noThrow.symbolScan` and `mdux-governed-lint` cover it as before, and a variant copy over
trivially copyable alternatives introduces neither an allocation nor a throw.

Roadmap updated: #15 is S1-S8 done, #198 next.

## Summary by CodeRabbit

* **New Features**
  * Added the `mdux-screenemit` command-line tool to generate C++ module and header outputs from MedUI screen packages.
  * Added validation for malformed packages, invalid identifiers, reserved names, and output failures with readable or JSON diagnostics.
  * Added CMake integration for automatically generating and consuming screen outputs.
  * Generated files are written only when content changes.

* **Bug Fixes**
  * Improved compile-time validation and payload handling for generated screens.

* **Documentation**
  * Updated the Wave 5 roadmap to reflect completed screen emitters and upcoming integration work.


tools/medui/Emit.cpp
```diff
@@ -0,0 +1,356 @@
...
+            return std::nullopt;
+        }
+    }
+    return bytes;
+}
+
+/// A name as a C++ string literal body. Names in a compiled screen are validated identifiers, keys
+/// and dotted tokens, so this escapes the two characters that could still end a literal early
+/// rather than pretending to be a general escaper.
+[[nodiscard]] std::string escape(std::string_view text) {
```
ambroise-leclerc, line 75: P2: Escape every value accepted by the `.medui` string grammar before embedding it in C++. The lexer deliberately decodes `\n` and `\t` inside source strings, and semantic analysis accepts unrestricted `String` values for fields such as `requirement`, `source`, `stream_source`, and `template`; the governed schema then checks only whether required values are non-empty. For example, a source field containing `"REQ\nHALT"` can therefore produce a valid package and pass `readPackage()`, but this function emits the decoded newline directly into the generated string literal. Both the `.cppm` and `.hpp` then fail to compile, and text following the newline can be interpreted as C++ tokens rather than data. Please encode newline, tab, carriage return, and any other control bytes as valid C++ escapes, and add an emitter test covering at least a source-legal escaped newline.

tools/medui/Emit.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+    out += std::format("namespace mdux::medui::generated::{} {{\n\n", identifier);
+
+    out += "/// The package id this screen was generated from.\n";
+    out += std::format("inline constexpr std::string_view id = {};\n\n", quoted(package.id));
+
+    out += renderStateArrays(package);
+
+    out += "/// Every node the compiler resolved, in the order the package lists them.\n";
+    out += "inline constexpr mdux::medui::CompiledNode nodes[] = {\n";
```
ambroise-leclerc: P2: Preserve support for the schema's valid empty-screen case. `ScreenPackage::validate()` explicitly permits `nodes.empty()` with an empty budget, and `medui-schema-budget` pins that behavior, but this unconditional declaration renders such a package as `inline constexpr CompiledNode nodes[] = {};`. An unsized array cannot deduce a zero bound in standard C++; strict Clang rejects it as a zero-size-array extension, so an otherwise valid compiled screen cannot be consumed in either emitted form. Please branch for the empty case and emit an empty `std::span<const CompiledNode>` (matching the shader emitter's empty-collection handling), with an emission/compilation test for a blank screen.

tools/medui/Emit.cpp
```diff
@@ -0,0 +1,356 @@
...
+    out += "// artifact, regenerated on every build. The reviewed source of truth is the JSON beside\n";
+    out += "// the digests, not these initialisers. See tools/medui/Emit.cppm.\n";
+    out += "//\n";
+    out += std::format("// Screen: {}\n", packageId);
+    return out;
+}
+
+}  // namespace
+
+std::string identifierForScreen(std::string_view packageId) {
```
ambroise-leclerc, line 331: P2: Ensure the sanitized name is a C++ identifier, not merely identifier-shaped. Package artifact IDs are valid slugs under `^[a-z0-9][a-z0-9-]*$`, so IDs such as `class`, `namespace`, `module`, or `import` are legal. This mapping leaves each unchanged, after which `renderBody()` uses it as a namespace component and `renderScreen()` uses it as a module-name component; the generated source is ill-formed. The CMake and C++ implementations currently agree on this invalid result, so the parity test does not catch it. Please use a collision-free rule that avoids keywords—prefixing all generated identifiers is the simplest option—and add keyword IDs to the cross-language parity cases.

tests/medui/GeneratedScreenHeaderConsumer.cpp
```diff
@@ -0,0 +1,25 @@
...
+ * a named module - a device's entry point may include Vulkan and platform headers before importing
+ * anything, precisely because mixing the two orders has produced GCC ICEs in this repository.
+ *
+ * Note the include order: `import std;` and `import mdux.medui.schema;` first, then the generated
+ * header, which assumes both. That is the contract the shader emitter's header form has too.
+ */
+import std;
+import mdux.medui.schema;
+
+#include "every_component.hpp"
```
ambroise-leclerc: Non-blocking formatting note: `clang-format --dry-run --Werror` reports this blank line as a violation and joins the two adjacent project includes. Please run the repository formatter on this file before merge.

tools/medui/Emit.cpp
ambroise-leclerc: Confirmed and fixed in 8debe52, and it is the worst of the three: it is a code-injection shape in a code generator, not a formatting problem.

The chain is exactly as you describe - `Lexer.cpp:225` decodes `\n` and `\t` inside a source string, the dictionary accepts an unrestricted `String` for `requirement`, `source`, `stream_source` and `template`, and `validate()` then asks only whether a required name is non-empty. So `"REQ\nHALT"` is a legal screen with a valid package, and the old escaper - which handled `"` and `\` and called itself deliberately narrow - wrote the decoded newline straight into a C++ literal. It ends the literal, and `HALT` is then read as tokens.

One detail worth flagging because it changes the encoding you suggested: control bytes are emitted as three-digit **octal**, not `\xNN`. A hex escape in C++ is greedy, so `"\x0aBC"` is one enormous escape while `"\012BC"` is a newline followed by two letters. Bytes at or above 0x80 are left alone - they are continuation bytes of a name that is already valid UTF-8, and the generated file is UTF-8.

`medui-screen-emit-escapes` is the test you asked for: a source-legal `requirement: "REQ\nHALT"` and `template: "TPL\tSCORE"` through parse, layout, package, JSON and back out through the emitter, asserting both that the escapes appear and that no decoded newline reaches the generated source.

tools/medui/Emit.cpp
ambroise-leclerc: Confirmed and fixed in 8debe52. `validate()` permits `nodes.empty()` with an empty budget, `medui-schema-budget` pins it, and the emitter rendered exactly the form you name - so a screen this schema calls valid could not be consumed in either output.

It now branches to `inline constexpr std::span<const mdux::medui::CompiledNode> nodes{};`, matching the shader emitter's empty-descriptor handling; both branches leave `nodes` convertible to the span `ScreenPackage` holds, so the tail of the rendering is shared rather than duplicated.

Both halves of the test you asked for are there. `medui-screen-emit-empty` covers the emission, and `tests/medui/fixtures/empty-screen-package.json` is registered as a second emit target so `GeneratedEmptyScreenConsumer.cpp` **compiles** it - which is the half that would have caught this in the first place. Only the module form is consumed for the blank screen, since the two forms share one rendered body and `medui-screen-emit-two-forms` asserts that; I would rather say so than have a second consumer imply an independent check.

tools/medui/Emit.cpp
ambroise-leclerc: Confirmed and fixed in 8debe52. `class`, `namespace`, `module` and `import` are all legal slugs under `^[a-z0-9][a-z0-9-]*$`, the mapping left each unchanged, and the result was used as both a namespace component and a module-name component. Your last sentence is the part that stings: the two implementations agreed on the invalid answer, so the parity test reported agreement on something no compiler accepts. Agreement was never the whole property.

Taken as you suggest - `identifierForScreen()` and `mdux_screen_identifier()` now prefix unconditionally, so `class` becomes `screen_class`. The prefix also retires the leading-digit rule, which was a second rule existing only for a case the first now cannot produce. Generated filenames change accordingly (`screen_every_component.cppm`), which costs nothing because nothing is committed and #198 has not landed. `class`, `namespace`, `module`, `import` and `int` are in the parity id list.

One case the prefix does not fix, which I handled rather than left: two adjacent separators map to `__`, reserved to the implementation everywhere. Collapsing the run would produce a usable name and destroy injectivity - `a--b` and `a-b` would claim one filename - so both sides **refuse** it instead, `SCE003` in the emitter and a `FATAL_ERROR` at configure time.

tests/medui/GeneratedScreenHeaderConsumer.cpp
ambroise-leclerc: Fixed. I had run the formatter over the emitter and the scenario files but not over the three consumer translation units - the formatter joins and sorts the two project includes.

All files this PR touches are now `clang-format --dry-run --Werror` clean. Worth noting for anyone who repeats the check: `tools/medui/Diagnostics.cpp` reports violations on `develop` too, so a repository-wide run is not currently clean and I have left that alone.

tools/medui/Emit.cpp
```diff
@@ -0,0 +1,426 @@
...
+[[nodiscard]] std::string preamble(std::string_view packageId, std::string_view packagePath) {
+    std::string out;
+    out += std::format("// Generated by {} from {}.\n", emitToolName, packagePath);
+    out += "//\n";
+    out += "// Do not edit, and do not commit: this file is a mechanical rendering of a committed\n";
+    out += "// artifact, regenerated on every build. The reviewed source of truth is the JSON beside\n";
+    out += "// the digests, not these initialisers. See tools/medui/Emit.cppm.\n";
+    out += "//\n";
+    out += std::format("// Screen: {}\n", packageId);
+    return out;
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and fixed in ca089d7. This is the same defect as the string-literal one, one line further down - my previous round escaped the values that end up inside quotes and stopped there, which is exactly the kind of partial fix that leaves a reviewer thinking the class is closed.

Both values now go through the same escaper. Worth adding to your analysis: a **trailing backslash** is the mirror image of the newline case - it splices the *following* line into the comment rather than ending it early - and the same escaper removes that too, since it doubles every backslash.

`medui-screen-emit-comment-escapes` is the regression test, with an id of `a\nnamespace`: it renders, and the whole id has to stay on the `// Screen:` line. I picked that id rather than something more obviously hostile because an id with two adjacent separators is now refused outright by `SCE003`, so the interesting case had to avoid tripping that first.

I also left a note at `preamble()` saying why a comment is escaped at all, so the next reader does not simplify it back: this is a code generator, so every value it writes has to be unable to change the structure of what it writes.

docs/roadmap.md
```diff
@@ -9,11 +9,13 @@ backlog that closes the gap. Waves 1, 2 and 3 have shipped — the renderer draw
...
+that generates the screens it draws. Eight of its twelve children have landed — the ADRs,
 the diagnostic registry, the front end, semantic analysis, bounded layout, per-locale text budgets,
-and golden references — so the compiler now reads a `.medui` file, resolves it to a bounded box
-tree, refuses a box that cannot hold its widest approved translation, and says where safety-critical
-content must appear. #197 is next: the emitters that write all of it down.
+golden references, and the canonical package with its C++ emitters — so the compiler now reads a
+`.medui` file, resolves it to a bounded box tree, refuses a box that cannot hold its widest approved
+translation, says where safety-critical content must appear, and writes the result as both a
+byte-compared artifact and `constexpr` C++ a device links without a parser. #198 is next: the CMake
+integration and the `mdux-meduic` host tool that runs all of it from a recipe.
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in ca089d7. The summary at line 12 and the `#15` heading at line 242 disagreed, and a progress count that contradicts itself is worse than either number alone.

tests/medui/GeneratedScreenConsumers.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * #197's acceptance is that both emitted forms describe one screen, and asserting it needs both in
+ * one binary. They cannot share a translation unit: the module form is attached to a named module
+ * and the header form is not, so a TU that both imported and included would hold two distinct sets
+ * of entities with the same names.
+ *
+ * So there are two thin translation units, one per form, each returning a `ScreenPackage`, and a
+ * third that compares them. Each of these files compiling is itself half the acceptance - the
+ * generated module must be importable, and the generated header includable, by an ordinary
+ * consumer that links no host-tools module.
+ */
```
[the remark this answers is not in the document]
ambroise-leclerc: Declining this one: it asks for the opposite of the rule this repository settled, and following it would reintroduce a bug.

`CONTRIBUTING.md:49` is explicit - "open the block with `@file`, naming the file **with its extension and no path**, then `@brief`" - and the four files cited already do that. The section goes on to explain why, with a measurement: `@file` is what attaches the block to the *file*; without it the block is not ignored, it silently becomes the documentation of whatever declaration follows. On `tests/framework/RunRecords.hpp` with Doxygen 1.15, dropping the tag moved the file brief onto `namespace mdux::spec` - wrong documentation, no warning.

The rule did previously say `@brief` first, which is presumably where the guideline being cited comes from. `CONTRIBUTING.md:69` records the change and the reason: it "was never what the tree did - 82 files used `@file` against 56 that did not - and review bots cite this document, so the contradiction kept resurfacing on unrelated pull requests." #180 settled it.

One accurate observation buried in it: `tests/medui/` is not internally consistent - the scenario files put `@brief` first, following `GoldenTests.cpp`. That predates this PR and I have left it alone rather than reformat merged files in a PR about emitters; worth a separate sweep if you want the tree uniform.

AM-L (approved):

PR #225, opened 2026-08-21
Compile a screen from a recipe [#198]

ambroise-leclerc:
First of two for S9. This one is the compiler: a recipe in, three artifacts out. The
`mdux_compile_screen()` registration, the first committed screen under `generated/screen/` and its
`evidence.screen.<id>` test follow in the second.

## What it adds

`mdux.tools.medui.compile` plus the `mdux-meduic` executable, split library-and-executable like every
other baker so the scenarios drive `run()`, `write()` and `verify()` as calls. Spawning the process
per case would make a failing assertion report that an exit status was wrong, which is the least
useful thing it could say about a compiler.

**The order is the deliverable.** Every stage this epic built already existed as a call; what did not
exist was a statement of the order they run in. This driver is the only place that answer is written
down - parse, analyze, validate annotations, resolve layout, check text budgets, collect goldens,
build and write. Annotations before layout is not arbitrary: the shared conformance suite pins
`MEDUI-E070` on a screen with no `surface:`, which the solver cannot resolve at all, so a design that
could only report it after layout could not satisfy the contract it implements.

A stage that reports anything stops the compile, rather than accumulating across stages. A later
stage reading a screen an earlier one rejected reports consequences rather than causes, and the
surface-mismatch scenario asserts exactly one diagnostic for that reason.

## The rule I would most like you to check

`checkTextBudgets()` requires a font package and **exactly** the locales that font approves, so a
recipe declaring neither cannot run the stage at all. Skipping it silently would be a hole: a screen
that draws text and declares no approved locale would be certified against a set nobody approved, and
the omitted translation is precisely the one that overflows a box.

So the driver asks whether the screen has anything to measure, and refuses an incomplete recipe with
`MEDUI-E002` rather than compiling without the check. The question is asked through
`needsTextBudget()`, added to #195's module next to the dynamic-text table it already owns - the two
dynamic-text fields (`Clock.format`, `TextInput.charset`) are checked *only* by the budget stage, so
a driver that answered this itself would be the second place that knows what counts as text, and the
one that goes stale when a third field joins them.

The stage is therefore skipped only for a screen with no text key and no dynamic-text field, which is
what `accepted-textless.medui` is: a Row background, an Image, a VulkanViewport and a positioned
SignalTrace.

## Recipe shape

No defaults, anywhere. `MEDUI-E002`'s registry entry gives the reason and I have kept to it: a
defaulted knob does not appear in `report.json`, so changing that default later would leave every
report looking unchanged, which ADR-007 forbids.

The surface is declared by the recipe rather than read from the source. The solver then checks the
source's own `surface:` against it, so the panel a product ships and the screen an author drew for it
disagree loudly instead of silently rescaling.

## Two things deferred deliberately

- **No text package is committed anywhere in this repository yet**, so no screen that draws text can
  be compiled end to end today. That is #201's (S12) job, and it is why the first artifact this epic
  commits will be text-free. Flagging it because it surprised me: the tool exists and is tested, but
  nothing registers a `recipes/text/` bake.
- **`verify` derives `goldens.json` from the package's directory.** The shared CLI grammar carries
  two output paths and a screen has three; ADR-012 puts all three in one directory, so the sidecar's
  location is not an independent fact and naming it twice would let the two disagree.

## Tests

`tests/medui/CompileTests.cpp`, five scenarios in `medui_tools_spec`: every knob required (each
removal refused, one at a time), the three artifacts a compile produces, the locale rule above, a
rejected stage stopping at the cause, and write/verify agreeing - including that a hand-edited
artifact does not verify.

The package is read back through `readPackage()` rather than inspected as the value that produced it,
so the scenario exercises the round trip a device's build would take.

## Regulatory impact

No device code changes; everything here is host-tools zone. What changes is that the compile order
becomes a contract rather than a convention, and that an incomplete recipe is refused rather than
producing a screen whose text boxes nobody measured. The bake report records the recipe, the source
and every input's digest, with `toolVersion` as the project's semantic version and nothing that
varies per machine - ADR-007 decision 5's rule, which a byte-compared artifact depends on.

## Summary by CodeRabbit

- **New Features**
  - Added the `mdux-meduic` command-line compiler for `.medui` recipes.
  - Generates validated screen packages, golden artifacts, and evidence reports.
  - Supports artifact writing, verification, tamper detection, and text or JSON diagnostics.
  - Added validation for recipe completeness, locales, text budgets, Unicode ranges, and compilation stages.
  - Textless screens can compile without text-budget or locale configuration.

- **Tests**
  - Added comprehensive coverage for compilation, validation failures, diagnostics, artifact verification, and textless screens.

### DIFF
--- a/tools/medui/Compile.cpp
+++ b/tools/medui/Compile.cpp
@@ -36,6 +36,45 @@ void report(std::vector<cli::Diagnostic>& diagnostics, Code code, std::string fi
     diagnostics.push_back(diagnose(code, std::move(file), line, 0, std::move(message), std::move(fixHint)));
 }
 
+/// Whether `id` is the lowercase slug an artifact directory, an evidence test name and a generated
+/// C++ identifier are all derived from. `mdux_bake_artifact()` enforces the same shape at configure
+/// time; this is the half that also holds when the compiler is run directly.
+[[nodiscard]] bool isArtifactSlug(std::string_view id) noexcept {
+    const auto lower = [](char c) {
+        return c >= 'a' && c <= 'z';
+    };
+    const auto digit = [](char c) {
+        return c >= '0' && c <= '9';
+    };
+    if (id.empty() || (!lower(id.front()) && !digit(id.front()))) {
+        return false;
+    }
+    return std::ranges::all_of(id, [&](char c) {
+        return lower(c) || digit(c) || c == '-';
+    });
+}
+
+/// A screen name and an artifact slug reduced to the one word they must both spell: lowercase, with
+/// every separator removed.
+///
+/// A weaker check than deriving one from the other, and deliberately so. A slug cannot be derived
+/// injectively from a screen name - a name may carry `-` and `_`, and lowercasing is not injective
+/// across case - which is why ADR-012 has the recipe *record* the mapping rather than compute it.
+/// What must not happen is the two drifting apart: `generated/screen/foo/` holding a package that
+/// calls itself `bar`. Comparing the reduced forms catches exactly that, and leaves an author free
+/// to hyphenate a slug as they like.
+[[nodiscard]] std::string reduced(std::string_view text) {
+    std::string out;
+    out.reserve(text.size());
+    for (const char c : text) {
+        if (c == '-' || c == '_') {
+            continue;
+        }
+        out.push_back(static_cast<char>(std::tolower(static_cast<unsigned char>(c))));
+    }
+    return out;
+}
+
 [[nodiscard]] std::span<const std::byte> asBytes(const std::string& text) noexcept {
     return {reinterpret_cast<const std::byte*>(text.data()), text.size()};
 }
@@ -185,6 +224,20 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
         return std::nullopt;
     }
 
+    // The slug shape ADR-012 fixes, checked here rather than left to `mdux_bake_artifact()`'s
+    // FATAL_ERROR: the id names the directory, the evidence test and the emitted C++ identifier, and
+    // a compile run outside CMake would otherwise write `generated/screen/NeuroSense500/`, which no
+    // registration could ever match.
+    if (!isArtifactSlug(recipe.id)) {
+        report(diagnostics,
+               Code::RecipeMissingMember,
+               std::string{recipePath},
+               0,
+               std::format("the id '{}' is not a lowercase slug", recipe.id),
+               "an artifact id matches ^[a-z0-9][a-z0-9-]*$ - it names a directory, a test and a C++ identifier");
+        return std::nullopt;
+    }
+
     const toml::Table* budget = document.table("budget");
     if (budget == nullptr) {
         report(diagnostics,
@@ -198,9 +251,10 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
     }
 
     try {
-        const std::int64_t vertices = budget->require("maxVertices").asInteger();
-        const std::int64_t indices  = budget->require("maxIndices").asInteger();
-        const std::int64_t commands = budget->require("maxCommands").asInteger();
+        const std::int64_t vertices   = budget->require("maxVertices").asInteger();
+        const std::int64_t indices    = budget->require("maxIndices").asInteger();
+        const std::int64_t commands   = budget->require("maxCommands").asInteger();
+        const std::size_t  budgetLine = budget->require("maxVertices").line();
         for (const std::int64_t value : {vertices, indices, commands}) {
             if (value < 0 || value > std::numeric_limits<std::uint32_t>::max()) {
                 report(diagnostics,
@@ -210,6 +264,18 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
                        std::format("a draw budget entry is {}, which is outside the range a budget holds", value));
                 return std::nullopt;
             }
+        }
+        // The 16-bit index bound, refused while the recipe can still be told about it. Left to
+        // `ScreenPackage::validate()` it becomes `BudgetExceedsIndexWidth` reported to a caller that
+        // treats a schema failure as a compiler bug and throws - so an author's extra digit would
+        // terminate the tool instead of producing a diagnostic.
+        if (vertices > mdux::draw::maxIndexableVertices) {
+            report(diagnostics,
+                   Code::RecipeMissingMember,
+                   std::string{recipePath},
+                   budgetLine,
+                   std::format("maxVertices is {}, and a 16-bit index cannot address more than {}", vertices, mdux::draw::maxIndexableVertices));
+            return std::nullopt;
         }
         recipe.budget = mdux::draw::DrawBudget{.maxVertices = static_cast<std::uint32_t>(vertices),
                                                .maxIndices  = static_cast<std::uint32_t>(indices),
@@ -293,6 +359,21 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
                        std::format("[dynamicText] entry '{}' names the range {}..{}, which is outside Unicode", names[index], first, last));
                 return std::nullopt;
             }
+            // A descending range is refused rather than read as empty, and it is the one entry here
+            // that would be fail-open if it were wrong. The budget stage deliberately skips a range
+            // with `last < first` - what an author wrote is a name, and the shape of the table behind
+            // it is not theirs to fix - so an inverted range produces *no* charset check at all for a
+            // Clock or a TextInput using that rule, while this table is the governed upper bound on
+            // what the runtime can put on screen.
+            if (first > last) {
+                report(diagnostics,
+                       Code::RecipeMissingMember,
+                       std::string{recipePath},
+                       namesLine,
+                       std::format("[dynamicText] entry '{}' names the descending range {}..{}", names[index], first, last),
+                       "an inverted range checks nothing, so it is refused rather than read as empty");
+                return std::nullopt;
+            }
             DynamicText rule;
             rule.name = names[index];
             rule.produces.push_back(mdux::font::CharsetRange{.first = static_cast<char32_t>(first), .last = static_cast<char32_t>(last)});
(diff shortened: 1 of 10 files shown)


AM-L:
@coderabbitai review

tools/medui/Compile.cpp
```diff
@@ -0,0 +1,599 @@
...
+               "recipe has no [package] table",
+               "add a [package] table with 'id' and 'source' keys");
+        return std::nullopt;
+    }
+
+    // Every knob is required, defaults included - which is to say there are none. MEDUI-E002 states
+    // the reason: a defaulted knob does not appear in report.json, so changing the default later
+    // would leave every report looking unchanged, which ADR-007 forbids.
+    try {
+        recipe.id            = package->require("id").asString();
```
ambroise-leclerc, line 213: [P1] Bind the recipe ID to the parsed screen identity and the evidence slug. The value accepted here is copied directly into `ScreenPackage::id`, but there is no lowercase-slug validation and no comparison with `screen.name`. The later CMake registration can therefore place outputs in `generated/screen/foo/` while package.json identifies itself as `bar`; two differently cased screen names can also be mapped inconsistently. ADR-012 requires an injective recorded mapping from the CamelCase screen name to the artifact slug. Please validate that mapping in the driver and include `id` in `Recipe::toOptions()` as the shader and text bakers do.

tools/medui/Compile.cpp
```diff
@@ -0,0 +1,599 @@
...
+            localeTexts.push_back(LocaleText{.package = &locale.package, .sidecar = locale.sidecar});
+        }
+
+        std::vector<DynamicTextRule> dynamicRules;
+        dynamicRules.reserve(recipe.dynamicText.size());
+        for (const DynamicText& rule : recipe.dynamicText) {
+            dynamicRules.push_back(DynamicTextRule{.name = rule.name, .produces = rule.produces});
+        }
+
+        const TextBudgetResult budgets = checkTextBudgets(layout, recipe.source, {.font = &*font, .locales = localeTexts, .dynamicText = dynamicRules});
```
ambroise-leclerc, line 663: [P2] Convert recipe-controlled package-wiring failures into diagnostics before calling this throwing API. For example, a valid font package with `packages = []`, an omitted approved locale, a duplicate locale, or a text package referencing another font all pass the recipe parser and loaders. `checkTextBudgets()` classifies those as caller wiring errors and throws `std::logic_error`; the executable has no catch boundary, so an author mistake aborts the compiler. The driver is the caller that resolves these recipe inputs, so it should validate the cross-package invariants and report MEDUI-E002, or catch and translate the exception. Add CLI-level coverage to ensure malformed recipes exit normally with structured diagnostics.

tools/medui/Compile.cpp
```diff
@@ -0,0 +1,599 @@
...
+            return std::nullopt;
+        }
+
+        for (std::size_t index = 0; index < names.size(); ++index) {
+            const std::int64_t first = firstPoints[index];
+            const std::int64_t last  = lastPoints[index];
+            // Bounded here rather than left to the walk: the budget stage refuses a range outside
+            // Unicode, but a recipe naming one should be told which entry it was rather than which
+            // glyph the walk stopped at.
+            if (first < 0 || last < 0 || first > 0x10FFFF || last > 0x10FFFF) {
```
ambroise-leclerc, line 354: [P1] Reject descending dynamic-text ranges here. A recipe with `firstCodePoints = [58]` and `lastCodePoints = [48]` passes this bounds check and creates a descending `CharsetRange`. `checkDynamicText()` deliberately skips ranges where `last < first`, so a Clock or TextInput using this rule reaches package generation without checking any character against the restricted font charset. Because this table is the governed upper bound on runtime-produced text, treating an inverted range as empty is fail-open. Please diagnose `first > last` during recipe parsing and add an end-to-end regression showing that it cannot compile.

tools/medui/Compile.cpp
```diff
@@ -0,0 +1,599 @@
...
+               "add a [budget] table with 'maxVertices', 'maxIndices' and 'maxCommands'. The runtime "
+               "fails closed against these, so a product declares them rather than inheriting them");
+        return std::nullopt;
+    }
+
+    try {
+        const std::int64_t vertices = budget->require("maxVertices").asInteger();
+        const std::int64_t indices  = budget->require("maxIndices").asInteger();
+        const std::int64_t commands = budget->require("maxCommands").asInteger();
+        for (const std::int64_t value : {vertices, indices, commands}) {
```
ambroise-leclerc, line 258: [P2] Validate the package-level budget preconditions while the recipe can still receive a diagnostic. This loop accepts zero for any entry and accepts `maxVertices` above `mdux::draw::maxIndexableVertices` as long as it fits uint32_t. A nonempty screen then reaches `buildPackage()`, whose schema validation throws `std::logic_error` for `EmptyBudget` or `BudgetExceedsIndexWidth`; `mdux-meduic` does not catch it, so malformed author input terminates instead of returning MEDUI-E002. Please reject these values here and cover both boundaries.

ambroise-leclerc (commented):
Four blocking findings are detailed inline: two fail-open or traceability issues and two malformed-recipe termination paths. Although this compiler is host-only, it establishes safety-relevant text-budget, dynamic-text, and evidence controls. Please classify the change explicitly as safety-relevant and obtain maintainer or domain-expert review. GitHub CI is green on GCC 16 and MSVC, and git diff --check passes. I did not run a local build because this checkout is on unsupported macOS and AppleClang. GitHub does not permit the PR author to submit a request-changes review, so this is recorded as a non-approving comment review.

tools/medui/Compile.cpp
ambroise-leclerc: [P1] Confirmed and fixed. Two checks, because the finding names two different failures.

**Shape.** `isArtifactSlug()` refuses anything outside `^[a-z0-9][a-z0-9-]*$` at parse. `mdux_bake_artifact()` enforces it at configure time, but a compile run directly would otherwise write `generated/screen/NeuroSense500/`, which no registration could ever match.

**Identity.** The slug and the screen name must reduce to the same word - lowercased, separators removed - so `Screen TextlessScreen` may be `textless-screen` or `textlessscreen`, and nothing else. Hyphenation stays the author's; the word does not.

One place I did less than the finding asks, and I would rather say so than let it look settled. **A slug cannot be derived injectively from a screen name**: names may carry `-` and `_`, and lowercasing is not injective across case, so `NeuroSense` and `Neurosense` reduce alike. That is exactly why ADR-012 has the recipe *record* the mapping rather than compute it. What the driver can guarantee is that the two do not drift apart - `generated/screen/foo/` holding a package calling itself `bar` - and that is what the reduced comparison catches. Injectivity across a *set* of screens stays a registration-level property: two recipes claiming one slug collide on the directory.

`toOptions()` records the id, as the shader baker's does. `medui-compile-identity` covers a malformed slug, a well-formed slug naming another screen, and a differently hyphenated one compiling.

The check also caught my own fixture, which called itself `Screen Textless` while the artifact was `textless-screen`. I renamed the screen rather than the slug, so the fixture now shows the pairing it is meant to demonstrate.

tools/medui/Compile.cpp
ambroise-leclerc: [P1] Confirmed and fixed - and this is the one I am most glad you found, because it is fail-open on the governed upper bound for what a device can put on screen.

The chain is as you describe: `checkDynamicText()` deliberately skips a range with `last < first` - "a descending range produces nothing and is skipped rather than reported: what a `.medui` author wrote is a name, and the shape of the table behind it is not theirs to fix" - which is the right call there, and turns an inverted entry in the recipe into **no charset check at all** for every Clock or TextInput using that rule. Not a narrower check: none.

`first > last` is now refused at recipe parse, with the reasoning recorded at the check so the next reader does not relax it back. `medui-compile-inverted-range` covers it.

One caveat on the end-to-end regression you asked for: it cannot exist yet. A screen with a Clock needs the budget stage, which needs a font package and its approved locales, and **no text package is committed in this repository** - so no screen carrying a `format:` can be compiled at all today. The scenario is therefore at recipe level, and the end-to-end version belongs with #201, which is where the text half lands.

tools/medui/Compile.cpp
ambroise-leclerc: [P2] Confirmed and fixed, both boundaries.

`maxVertices` past `mdux::draw::maxIndexableVertices` is refused at parse - it is wrong regardless of what the screen turns out to contain. The empty budget is refused after layout instead, because that is where the fact that makes it invalid appears: `validate()` permits an empty budget on a screen with no nodes, so a parse-time refusal would have been a rule the schema does not have.

The mechanism you name is the part worth keeping in view: both values reached `ScreenPackage::validate()`, and `buildPackage()` treats a schema failure as a *compiler bug* and throws - correctly, since every other invariant it checks was already enforced upstream. That made an author's extra digit indistinguishable from a defect in the compiler. Keeping `logic_error` meaning "a gate was bypassed" is worth preserving, which is why these are diagnosed before they get there rather than caught after.

`medui-compile-budget-bounds` covers both.

tools/medui/Compile.cpp
ambroise-leclerc: [P2] Confirmed and fixed. `checkLocaleWiring()` in the driver now diagnoses each of the four cases you list - an empty package list, a package baked against another font's atlas, a repeated locale, and a locale the font approves that the recipe omits - before `checkTextBudgets()` can throw over them.

The distinction I have tried to keep sharp: the library's contract is right to call these caller errors, because its caller is normally code. Here the caller resolves paths an author typed, so the same conditions are author input and must produce a report. What stays a `std::logic_error` is a bypassed gate, which is a defect rather than a recipe.

`main()` also gains a catch boundary, as the parallel CodeRabbit finding suggested: an escaping exception now renders one diagnostic in the requested format rather than reaching the terminate handler, so `--format=json` always gets an envelope.

For the CLI-level coverage you asked for, `medui-compile-cli` spawns `mdux-meduic` with a malformed recipe and asserts it exits non-zero having printed an envelope carrying `MEDUI-E003`. It is the only scenario here that spawns the process, and it earned its place immediately: the first version passed on Linux and failed on MSVC because `cmd.exe` strips the outer quotes from a quoted program path, so the process never ran and the exit status was non-zero for entirely the wrong reason. The log is now asserted non-empty before its contents.

tools/medui/Compile.cpp
```diff
@@ -0,0 +1,599 @@
...
+    // 1. Parse. The lexer decides UTF-8 validity, so MEDUI-E004 comes from there rather than from a
+    //    second check here that could disagree with it.
+    ParseResult parsed = parse(textOf(*sourceBytes), recipe.source);
+    if (!parsed.diagnostics.empty() || !parsed.screen.has_value()) {
+        diagnostics.insert(diagnostics.end(), parsed.diagnostics.begin(), parsed.diagnostics.end());
+        if (diagnostics.empty()) {
+            report(diagnostics, Code::SourceUnreadable, recipe.source, 0, "the source produced no screen and no diagnostic");
+        }
+        return std::nullopt;
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Taken as suggested. `diagnostics` is an out-parameter, `CompileTests.cpp` does reuse one vector across `parseRecipe()` and `run()`, and the guard as written asked the wrong question - so a parse that produced neither a screen nor a reason would have returned nullopt with nothing of its own to say, and only when the vector happened to be non-empty. Guarding on `parsed.diagnostics` makes the answer depend on the parse rather than on what the caller brought with it.

tools/medui/Compile.cpp
```diff
@@ -0,0 +1,599 @@
...
+        font = loadFont(recipe, root, inputs, diagnostics);
+        if (!font.has_value()) {
+            return std::nullopt;
+        }
+        std::optional<std::vector<LoadedLocale>> loaded = loadLocales(recipe, root, inputs, diagnostics);
+        if (!loaded.has_value()) {
+            return std::nullopt;
+        }
+        locales = std::move(*loaded);
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Both halves taken, and they meet the maintainer's P2 on the same path from the other side.

The empty `packages` array is now one of four wiring conditions `checkLocaleWiring()` diagnoses in the driver - the others being a package baked against another font's atlas, a repeated locale, and an approved locale the recipe omits. Each is decided by paths an author typed, so each is a report rather than a `std::logic_error` from a library whose caller is supposed to be code.

`main()` gains the catch boundary too. Its comment states the narrow thing it is for: every recipe-controlled failure is already a diagnostic, so an escaping exception means a gate was bypassed - and even then a consumer that asked for `--format=json` is entitled to an envelope rather than a terminate handler's message on stderr.

AM-L (approved):

tools/medui/Compile.cpp
ambroise-leclerc: Correction to my reply above, found by the very test this PR set up.

I wrote that `toOptions()` records the id. It did not: the edit was lost in a batch that aborted before writing, and I reported it as done without re-reading the function. Nothing caught it until `evidence.screen.endoscope-monitor` in #231 compared a freshly baked `report.json` against the committed one and found it 31 bytes shorter - the width of the missing member.

Fixed in #231 (`f4a68ac`). The rest of that reply stands: the slug shape and the identity check are in and covered by `medui-compile-identity`.

PR #231, opened 2026-08-22
Register the first compiled screen as an evidence artifact [#198]

### LINKED ISSUE #198: S9 — CMake integration and the mdux-meduic host tool

ambroise-leclerc:
Closes #198. The second half of S9: a screen becomes an evidence artifact like every other kind.

## What it adds

- **`mdux_compile_screen()`** - a thin front for `mdux_bake_artifact()`, so a screen gets the same
  build-tree bake, the same `-update` target and the same `evidence.screen.<id>` byte comparison on
  both toolchain legs. Nothing about a screen justifies a second mechanism, and a second mechanism is
  how one artifact kind ends up with weaker evidence than its neighbours.
- **`generated/screen/endoscope-monitor/`** - the first screen this repository commits, with its
  recipe and `.medui` source under `recipes/screen/`.

## What the wrapper adds over calling the baker directly

Two things a call site could otherwise get wrong without anything failing:

- **The three outputs are fixed in the function.** ADR-012 makes `package.json`, `goldens.json` and
  `report.json` unconditional, so a screen pinning nothing writes `[]` rather than one fewer file. A
  call site listing its own `OUTPUTS` could drop one and still pass its own comparison.
- **The `.medui` source is read out of the recipe** and becomes a dependency automatically. A
  forgotten `SOURCES` entry does not fail loudly: it makes edits to the screen stop triggering a
  rebake, so the committed artifact quietly stops matching its source until someone runs the update
  target for an unrelated reason. A recipe whose `source` line the regex cannot find is a configure
  error instead.

## The screen, and the limit it is under

`EndoscopeMonitor` draws a Row background, a baked image, a live video surface and a live waveform.
The waveform is positioned, so `goldens.json` carries one entry rather than being empty and proving
nothing.

**It draws no text, and that is a limit rather than a design.** No text package is baked anywhere in
this tree, so the text-budget stage has nothing to measure against and a screen carrying `t("STR-KEY")`
is refused - correctly - for declaring no approved locale. #201 lands the text half; this screen
proves the rest of the pipeline end to end meanwhile. Both the recipe and the source say so where a
reader will meet the question.

## How the committed bytes were produced, since it matters here

I cannot build this repository locally - the GCC 16 floor means CI is the only place it compiles - so
the three artifacts were computed rather than baked: the JSON by the same canonical rules
`mdux.evidence.json` implements (sorted keys, two-space indent, trailing newline), and every digest
in `report.json` with SHA-256 over the exact committed bytes.

That is a claim, not evidence, and it is exactly what `evidence.screen.endoscope-monitor` exists to
check: CI bakes the screen with `mdux-meduic` and byte-compares. If any of my arithmetic is off - a
rectangle, a node order, a digest - that test fails on both legs and says so. Reviewing this PR is
partly reading whether the test passed.

## Regulatory impact

A screen joins the evidence pipeline on the same terms as a font, a shader and a model: reviewed as
committed text, byte-compared across MSVC and GCC, and rebuilt only through `mdux-bake-update`, the
one path allowed to write into the source tree. `report.json` records the recipe, the source and both
outputs by digest, with `toolVersion` as the project's semantic version and nothing that varies per
machine.

Roadmap: #15 is S1-S9 done, #199 next. `docs/architecture.md` lists six committed artifacts, and says
why the screen is the one whose payload is not opaque bytes.

## Summary by CodeRabbit

* **New Features**
  * Added an endoscope monitor screen with a themed header, live Vulkan viewport, and ECG signal trace.
  * Added a 1280×720 screen recipe with defined runtime drawing limits.
  * Integrated automatic screen compilation and evidence artifact generation.

* **Documentation**
  * Documented the new reviewable screen artifact and updated project roadmap progress.

* **Tests**
  * Added validation for screen recipe parsing, identity checks, and source-file handling.

### DIFF
--- a/cmake/MduXCompileScreen.cmake
+++ b/cmake/MduXCompileScreen.cmake
@@ -1,0 +1,81 @@
+# Registering a compiled screen as an evidence artifact (issue #198).
+#
+# `mdux_compile_screen()` is a thin front for `mdux_bake_artifact()`: a screen is baked like a font,
+# a shader or a model, so it gets the same build-tree output, the same `-update` target that stages
+# into `generated/`, and the same `evidence.screen.<id>` byte-comparison on every toolchain leg.
+# Nothing about a screen justifies a second mechanism, and a second mechanism is how one artifact
+# kind ends up with weaker evidence than its neighbours.
+#
+# What this wrapper adds over calling `mdux_bake_artifact()` directly is the two things a caller
+# could otherwise get wrong silently:
+#
+#   - **The three outputs are fixed here**, not per call site. ADR-012 makes `package.json`,
+#     `goldens.json` and `report.json` unconditional, so a screen that pins nothing writes `[]`
+#     rather than one fewer file. A call site listing its own OUTPUTS could drop one and produce a
+#     smaller artifact that still passed its own comparison.
+#   - **The `.medui` source becomes a dependency automatically**, read out of the recipe. A
+#     forgotten SOURCES entry does not fail: it makes edits to the screen stop triggering a rebake,
+#     so the committed artifact quietly stops matching its source until someone runs the update
+#     target for another reason.
+
+include_guard(GLOBAL)
+
+# mdux_compile_screen(ID <id> RECIPE <path> [SOURCES <path>...])
+#
+# SOURCES names further inputs the recipe references - the font and text packages a screen that
+# draws text needs. The `.medui` source itself is found in the recipe and does not have to be
+# repeated.
+function(mdux_compile_screen)
+    set(options "")
+    set(single ID RECIPE)
+    set(multi SOURCES)
+    cmake_parse_arguments(ARG "${options}" "${single}" "${multi}" ${ARGN})
+
+    if(ARG_UNPARSED_ARGUMENTS)
+        message(FATAL_ERROR "mdux_compile_screen: unexpected arguments: ${ARG_UNPARSED_ARGUMENTS}")
+    endif()
+    foreach(required ID RECIPE)
+        if(NOT ARG_${required})
+            message(FATAL_ERROR "mdux_compile_screen: ${required} is required")
+        endif()
+    endforeach()
+
+    set(recipe_path "${CMAKE_SOURCE_DIR}/${ARG_RECIPE}")
+    if(NOT EXISTS "${recipe_path}")
+        message(FATAL_ERROR "mdux_compile_screen: recipe '${ARG_RECIPE}' does not exist")
+    endif()
+
+    # The screen source, taken from the recipe rather than from the call site. A regex over one
+    # `key = "value"` line rather than a TOML parse: CMake has no TOML reader, the compiler is the
+    # authority on the recipe's meaning, and the only thing needed here is the dependency edge.
+    # A recipe whose `source` this cannot find is a configure error rather than a build that
+    # silently never rebakes.
+    file(READ "${recipe_path}" recipe_text)
+    string(REGEX MATCH "\n[ \t]*source[ \t]*=[ \t]*\"([^\"]+)\"" _match "\n${recipe_text}")
+    if(NOT CMAKE_MATCH_1)
+        message(FATAL_ERROR
+            "mdux_compile_screen: no `source = \"...\"` line in ${ARG_RECIPE}. The screen source has "
+            "to be a dependency of the bake, or editing it would not trigger one.")
+    endif()
+    set(screen_source "${CMAKE_MATCH_1}")
+
+    if(NOT EXISTS "${CMAKE_SOURCE_DIR}/${screen_source}")
+        message(FATAL_ERROR
+            "mdux_compile_screen: ${ARG_RECIPE} names the source '${screen_source}', which does not exist")
+    endif()
+
+    mdux_bake_artifact(
+        KIND screen
+        ID ${ARG_ID}
+        TOOL mdux-meduic
+        RECIPE ${ARG_RECIPE}
+        SOURCES
+            ${screen_source}
+            ${ARG_SOURCES}
+        # Fixed here rather than per call site: all three are unconditional (ADR-012, decision 1).
+        OUTPUTS
+            package.json
+            goldens.json
+            report.json
+    )
+endfunction()
--- a/cmake/MduXCompileScreen.cmake
+++ b/cmake/MduXCompileScreen.cmake
@@ -20,6 +20,80 @@
 
 include_guard(GLOBAL)
 
+# _mdux_screen_package_field(<out_var> <recipe_text> <key> <label>)
+#
+# The value of `key` in the recipe's `[package]` table, and nothing else's.
+#
+# Scoped to that table on purpose. `mdux-meduic` reads `document.table("package")->require("source")`
+# and ignores tables it does not know, so a recipe may legally carry an earlier `source` key in
+# another table - and a regex over the whole document would then hand CMake a different file from the
+# one the compiler compiles. The dependency edge would point at the wrong input, editing the real
+# screen would stop triggering a rebake, and `mdux-bake-update` would stage bytes nobody asked for.
+# That is precisely the silent drift this wrapper exists to prevent, so getting the scope wrong here
+# is worse than not having the wrapper.
+#
+# Two keys in one table is an error rather than a first-wins: the compiler's TOML reader decides that
+# case, this does not have to guess the same way, and a recipe that ambiguous should be fixed.
+#
+# Line-based rather than a single regex, because a value or a comment may contain `[` - this
+# repository's own recipes carry `^[a-z0-9]...` inside a comment in `[package]` - and a section match
+# that stopped at the next bracket would truncate the table.
+function(_mdux_screen_package_field out_var recipe_text key label)
+    string(REPLACE ";" "\\;" escaped "${recipe_text}")
+    string(REGEX REPLACE "\r?\n" ";" lines "${escaped}")
+
+    set(current_table "")
+    set(found "")
+    foreach(line IN LISTS lines)
+        # Both patterns are anchored after leading whitespace, so a commented-out `# source = "x"` or
+        # `# [package]` cannot match. That is why no comment stripping is needed, and why stripping
+        # would be wrong: a `#` inside a quoted value is part of the value.
+        if(line MATCHES "^[ \t]*\\[([A-Za-z0-9_.-]+)\\]")
+            set(current_table "${CMAKE_MATCH_1}")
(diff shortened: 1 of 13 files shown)


cmake/MduXCompileScreen.cmake
```diff
@@ -0,0 +1,81 @@
...
+        message(FATAL_ERROR "mdux_compile_screen: recipe '${ARG_RECIPE}' does not exist")
+    endif()
+
+    # The screen source, taken from the recipe rather than from the call site. A regex over one
+    # `key = "value"` line rather than a TOML parse: CMake has no TOML reader, the compiler is the
+    # authority on the recipe's meaning, and the only thing needed here is the dependency edge.
+    # A recipe whose `source` this cannot find is a configure error rather than a build that
+    # silently never rebakes.
+    file(READ "${recipe_path}" recipe_text)
+    string(REGEX MATCH "\n[ \t]*source[ \t]*=[ \t]*\"([^\"]+)\"" _match "\n${recipe_text}")
```
ambroise-leclerc: This regex takes the first line shaped like `source = "..."` anywhere in the document, whereas `mdux-meduic` specifically reads `document.table("package")->require("source")`. The recipe parser currently tolerates unrelated tables, so a valid recipe can contain an earlier `source` key that CMake selects while the compiler reads a later `[package].source`. On an incremental build, editing the actual `.medui` file then does not invalidate the custom command; `mdux-bake-update` can stage stale output and the comparison can keep checking the previously baked bytes. That is the exact silent dependency drift this wrapper is meant to prevent. Please scope extraction to the `[package]` table (and reject ambiguity), or expose the compiler parser’s resolved dependency information to CMake. A configure/incremental-build test with another table containing `source` would demonstrate the fix.

cmake/MduXCompileScreen.cmake
```diff
@@ -0,0 +1,81 @@
...
+    set(screen_source "${CMAKE_MATCH_1}")
+
+    if(NOT EXISTS "${CMAKE_SOURCE_DIR}/${screen_source}")
+        message(FATAL_ERROR
+            "mdux_compile_screen: ${ARG_RECIPE} names the source '${screen_source}', which does not exist")
+    endif()
+
+    mdux_bake_artifact(
+        KIND screen
+        ID ${ARG_ID}
```
ambroise-leclerc, line 134: `mdux_compile_screen()` does not verify that the registration `ID` matches `[package].id` in the recipe. As written, `mdux_compile_screen(ID foo RECIPE bar.toml)` can compile a recipe whose package id and `Screen` name reduce to `bar`, then place those bytes under `mdux_bake/screen/foo`, compare them through `evidence.screen.foo`, and stage them into `generated/screen/foo/`. The baker correctly checks the recipe id against the screen name, but it has no knowledge of the CMake `ID`, so every build and evidence comparison can pass while the directory/test identity disagrees with the package and report identity. That breaks the traceability invariant this wrapper is intended to centralize. Please extract `[package].id` alongside the source and fail configuration unless it equals `ARG_ID`; a configure-time negative test for a mismatched pair would keep this invariant from regressing.

cmake/MduXCompileScreen.cmake
ambroise-leclerc: Confirmed and fixed in 2a1bea0. The regex read the whole document while `mdux-meduic` reads `document.table("package")` and ignores tables it does not know, so the two could disagree about which file the screen even is - and the failure mode is the silent one you describe: the dependency edge points at the wrong input, editing the real screen stops triggering a rebake, and `mdux-bake-update` stages bytes nobody asked for. A wrapper that gets this wrong is worse than no wrapper, because it looks like the check is being made.

Extraction is scoped to `[package]` now, and it is line-based rather than a section regex: a value or a comment may contain `[`, and this repository's own recipe carries `^[a-z0-9]...` inside a comment in that table, so a section match ending at the next bracket would have truncated it. Two keys in one table is an error rather than a first-wins - which of them the compiler would use is its TOML reader's business, and this should not guess the same way.

`screen.recipe.scoped` is the test you asked for: a recipe with `[assets] source = "decoy.png"` *before* `[package]`, asserting the extraction returns the `[package]` value. `screen.recipe.ambiguous` and `screen.recipe.missing` cover the other two outcomes, matching the message rather than the exit status.

cmake/MduXCompileScreen.cmake
ambroise-leclerc: Confirmed and fixed in 2a1bea0, and this one is the sharper of the two: the driver's identity check - added in the previous PR for exactly this class of problem - was defeated one level up, because `mdux-meduic` never sees the CMake `ID`.

Your trace is right end to end: the bytes go to `mdux_bake/screen/foo`, the comparison runs as `evidence.screen.foo`, the staging lands in `generated/screen/foo/`, and every check passes while the directory, the test and the package disagree about which screen this is. Nothing downstream could notice, because each half is self-consistent.

`_mdux_screen_check_recipe()` now extracts `[package].id` and fails configuration unless it equals `ARG_ID`. `screen.recipe.identity` is the negative test - a recipe compiling `real-screen` registered as `another-screen` - matching the message, and `screen.recipe.identity-accepts` and `screen.recipe.committed` hold the positive side, the latter reading the recipe this repository actually commits.

Both checks live in script-mode functions so the tests are pure functions of the recipe text; configuring a fixture project to reach them would be slower and would report a failure two layers from its cause.

docs/architecture.md
```diff
@@ -145,6 +145,13 @@ Five artifacts are committed today:
 | `generated/model/ecg-demo/` | `mdux-mlbake` | `weights.bin` |
 | `generated/model/ecg-demo-alt/` | `mdux-mlbake` | `weights.bin` |
 | `generated/font/dejavu-ui/` | `mdux-textbake` | `atlas.bin` |
+| `generated/screen/endoscope-monitor/` | `mdux-meduic` | `goldens.json` |
+
+The screen is the one entry whose payload is not opaque bytes, and ADR-012 explains why: a screen
+cannot bake vertices, because four of its components draw from live data and their geometry does not
+exist until the frame does. What it bakes instead is layout - where each node is, how much it may
+draw, and which validated token and key it draws with - and a sidecar of golden references saying
+where safety-critical content must appear. Both are reviewable as text, which is the point.
```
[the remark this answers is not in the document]
ambroise-leclerc: Both halves taken in b0855a6, and both were wrong about the artifact this PR commits rather than merely imprecise.

The payload column named `goldens.json` alone, which conflates the compiled package with its sidecar - they have different consumers and different lifetimes, and keeping them apart is decision 4 of ADR-012. It now names both.

The golden-set sentence is the one that mattered: I wrote "where safety-critical content must appear" while ADR-011 puts **every** `@safety_critical` node *and* every explicitly positioned one in the set. The committed screen has exactly one entry, and it is there for the second reason - its `SignalTrace` is positioned, not annotated - so my own prose did not describe my own artifact.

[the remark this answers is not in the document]
ambroise-leclerc: Fair. The claim was about the *dictionary* - `NumericDisplay`, `SignalTrace`, `StatusIndicator` and `Clock` are the four components ADR-012 names - but it sat immediately under a table row about this screen, so it read as a claim about this screen, where it is not true. Fixed in b0855a6: the sentence names the four and says which set of eleven they are four of.

docs/roadmap.md
```diff
@@ -261,8 +262,8 @@ happened to read.
 - #195 S6 Text-budget validation against every approved locale · _closed_
 - #196 S7 Golden references for safety-critical nodes · _closed_
 - #197 S8 Canonical package and C++ emitters · _closed (PRs #220, #221)_
-- #198 S9 CMake integration and the `mdux-meduic` host tool · **next**
-- #199 S10 Allocation-free screen runtime
+- #198 S9 CMake integration and the `mdux-meduic` host tool · _closed (PRs #225, #228)_
```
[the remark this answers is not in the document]
ambroise-leclerc: Correct, and my error: I wrote the roadmap entry before opening the PR and guessed the next number. #228 is a Dependabot bump; this PR is #231. Fixed in b0855a6.

AM-L (approved):

Issue #200, opened 2026-08-11
S11 — mdux-medui-check


ambroise-leclerc:
Part of #15. Blocked by S6; independent of S8-S10.

## Work

Validate a single `.medui` file and print diagnostics, without a full build. This is the surface an author and an agent both use, and the `medui-authoring` skill already promises it.

- Text and JSON output, the JSON in the shared envelope (#118).
- Non-zero exit on any error-severity diagnostic.

## Acceptance

- Update the `medui-authoring` skill: it currently says "there is no tooling shortcut yet" and must stop saying so once this exists.
- A fixture drives the tool over a deliberately broken screen and asserts the codes.

PR #232, opened 2026-08-22
Draw a compiled screen without allocating [#199]

### LINKED ISSUE #199: S10 — Allocation-free screen runtime

ambroise-leclerc:
Closes #199. The governed half of the epic: a compiled screen becomes draw commands, with no
allocation and no work a device cannot bound before it runs.

## What it adds

`mdux.medui.screen` in `MduXCore`, which is what puts it under `mdux-governed-lint` and
`governed.noThrow.symbolScan` without either being told about it. The scratch is the caller's, the
bound is the screen's own `DrawBudget`, and every refusal is a `Result`.

## The decision I would most like you to overrule if you disagree

**It draws a `Panel`. Every other component is visited, counted, and left undrawn.** That is a
stated limit, not an omission, and the reason differs by component:

- `Label`, `Button`, `CriticalButton`, `TextInput` draw **text** - a compiled screen carries a
  `textKey`, not glyphs, so drawing them is a join with a baked text package. This repository has
  none; the join belongs with #201.
- `Clock`, `NumericDisplay`, `SignalTrace`, `StatusIndicator` draw from **live data**. Their geometry
  does not exist until the frame does, which is ADR-012's reason a screen bakes layout rather than
  vertices.
- `Image` needs a baked image package, which does not exist either.

What I deliberately did **not** do is decide that a `Label`'s box should be filled with its colour,
or that a `Button` has a face in its. Nothing in this project settles per-component appearance today
— not the ADRs, which stop at "where each node is and which validated token it draws with", and not
the sibling: TrustSC's `render_frame` returns `FrameStatistics` rather than geometry
(`crates/trustsc-ui/src/lib.rs:539`), so there is no parity source to follow either. Inventing it
here would make this module authoritative over a question it has no evidence for.

So the runtime reports what it could not decide in `FrameStats::deferred`, and an integrator sees
"eleven nodes, one drawn" instead of a screen that quietly renders less than it appears to. If you
would rather it painted something for the other kinds now, that is a call worth making explicitly -
and it needs a pixel test to settle it, which is #16/#17's ground.

## The bounded-work acceptance

`FrameStats::steps` exists for the two tests #199 specified:

- `medui-screen-work-is-pure` renders one screen twice and asserts the counts are identical, so the
  work is a function of the package and of nothing ambient. It also compares the recorded vertex
  bytes, which is what a cross-toolchain comparison rests on.
- `medui-screen-work-scales` renders screens of *n* and *2n* nodes and asserts the count doubles.
  Stated as an equality rather than a bound, because work growing with something other than the node
  count would land between the two and a bound would let it pass.

A component that looped until a condition breaks the first; one whose work grew with its data breaks
the second. That is the thing #199 said was missing: the bound held by argument and nothing held the
runtime to it.

## Failing closed

A refused frame rolls the list back to where the frame began, so a frame is whole or absent. A
half-drawn frame on a medical display is the worst outcome available, because it looks like a
reading. Two scenarios cover it: an unknown colour token after a panel was already recorded, and a
budget with room for one rectangle on a screen with two - both leaving the caller's own earlier
drawing untouched.

## No-heap, in three layers as #63 established

- `medui_noheap_spec` replaces the global `operator new` family and measures: eight frames, the
  refusal path, and a self-test that deliberately allocates so the suite cannot become a test that
  cannot fail.
- `screen.noheap.symbolScan` scans the emitted objects. The profile is spelled `ml-noheap` because
  that is its historical name; what it forbids is not ML-specific, so I reused it rather than copying
  it under a second name. Renaming it properly would touch the ML wiring in a PR about screens -
  worth doing separately if you want it.
- `governed.noThrow.symbolScan` already covers the module, by virtue of it being in `MduXCore`.

The interposition moved to `tests/framework/CountingAllocations.hpp`, shared with the ML suite.
Replacing a global operator is subtle enough - hand-rolled over-aligned allocation, a scoped GCC
diagnostic for a mismatched-new-delete the compiler cannot see through - that a second copy would
have been a second thing to keep right. `tests/ml/NoHeapTests.cpp` keeps its scenarios and includes
the header.

## Colour, and one determinism note

The governed table stores linear RGBA and the vertex colour is read as `R8G8B8A8_UNORM`, so
`quantise()` scales and rounds and applies no transfer function: whether the swapchain is sRGB is the
renderer's decision, and encoding in two places is how a colour gets encoded twice. The multiply and
the add are separate statements, and `mdux_enforce_fp_determinism(MduXCore)` already turns
contraction off for this target - a frame that is byte-compared across toolchains cannot afford a
last-bit difference in a colour.

## Regulatory impact

The first governed code in this epic that runs on a device. It allocates nothing, throws nothing,
performs no parsing, and does work bounded by two numbers known before the device starts - now with
tests that hold it to the second one rather than an argument that it holds. A frame that cannot be
completed is not shown in part.

Roadmap: #15 is S1-S10 done, #200 next.

## Summary by CodeRabbit

* **New Features**
  * Added governed screen rendering for supported panels.
  * Added frame statistics, deferred handling for unsupported content, and clear rendering errors.
  * Rendering avoids allocations and prevents partial output when limits or color values are invalid.

* **Documentation**
  * Updated roadmap progress, verification details, and upcoming release guidance.

* **Tests**
  * Added coverage for deterministic rendering, rollback behavior, draw budgets, and allocation-free execution.


CMakeLists.txt
```diff
@@ -198,6 +198,7 @@ target_sources(MduXCore
             include/mdux/text/Draw.cppm
             include/mdux/text/Schema.cppm
             include/mdux/medui/Schema.cppm
+            include/mdux/medui/Screen.cppm
```
ambroise-leclerc, line 201: This PR registers `mdux.medui.screen` as a new governed module, but the authoritative module table in `docs/architecture.md` still ends the MedUI entries at `mdux.medui.schema`. That document explicitly says every module named there is checked against `CMakeLists.txt`, and the repository definition of done requires architecture documentation to move with public/runtime architecture changes. Please add `mdux.medui.screen` with `include/mdux/medui/Screen.cppm` and `src/medui/Screen.cpp` to the governed module table (and update the root architecture summary if it is maintained in the same change), so the documented current state matches the build after merge.

tests/medui/ScreenRuntimeTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+                      checks.expect(*frameA == *frameB, std::format("identical statistics, got steps {} then {}", frameA->steps, frameB->steps));
+
+                      const std::span<const mdux::draw::UiVertex> a = listA.vertices();
+                      const std::span<const mdux::draw::UiVertex> b = listB.vertices();
+                      checks.expect(a.size() == b.size(), "the same number of vertices");
+                      const bool sameBytes = a.size() == b.size()
+                                             && std::equal(reinterpret_cast<const std::byte*>(a.data()),
+                                                           reinterpret_cast<const std::byte*>(a.data()) + a.size_bytes(),
+                                                           reinterpret_cast<const std::byte*>(b.data()));
+                      checks.expect(sameBytes, "and the same bytes, which is what a cross-toolchain comparison rests on");
```
ambroise-leclerc, line 158: This compares two renders produced by the same executable, so it proves only repeatability within one toolchain invocation. GCC and MSVC can emit different vertex bytes and both CI legs will still pass because each compares its own output with itself. That does not satisfy #199 acceptance that a fixed input screen produces byte-identical draw output across toolchains, and the runtime is not yet exercised by any committed evidence or pixel artifact that would provide that comparison indirectly. Please check a frozen expected byte sequence/digest on both legs, or publish one leg output and compare it with the other in CI. Field-level expected assertions can also work if they completely pin every emitted byte, including colour quantisation and padding.

tests/medui/ScreenRuntimeTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+                                                                               checks.expect(false, "both frames are recorded");
+                                                                               checks.raise();
+                                                                               return;
+                                                                           }
+
+                                                                           // #199's second criterion. An equality rather than a bound: work that grew
+                                                                           // with something other than the node count would land between the two, and a
+                                                                           // bound would let it pass.
+                                                                           checks.expect(
+                                                                               twoN->steps == n->steps * 2,
```
ambroise-leclerc, line 188: These assertions do not mechanically hold future per-node work to the claimed bound. `steps` is self-reported by the implementation: a future inner loop can simply perform work without incrementing it, leaving both tests green. Even when every iteration increments `steps`, rendering identical input twice remains identical, and duplicating identical nodes makes any deterministic node-data-dependent work double exactly, so both the purity and 2n tests still pass. For example, work proportional to `node.bounds.width` passes both scenarios as written. Please either enforce a per-node cap structurally/type-level, instrument the actual bounded operations, or add same-node-count cases with deliberately varied payload/data dimensions and narrow the claim to what those tests can establish. As written, the verification evidence does not support the safety-relevant statement that a future data-dependent loop would be detected.

src/medui/Screen.cpp
```diff
@@ -0,0 +1,106 @@
...
+            ++stats.deferred;
+            continue;
+        }
+
+        const auto colour = resolveColorToken(panel->colorToken);
+        if (!colour.has_value()) {
+            return refuse(ScreenError::UnknownColorToken);
+        }
+
+        if (const auto recorded = list.addSolidRect(toRect(node.bounds), toColor(*colour)); !recorded.has_value()) {
```
ambroise-leclerc, line 112: The runtime never reads `screen.budget`; `addSolidRect()` enforces only the budget used to create `list`. Consequently, a screen declaring room for one rectangle can render two successfully whenever the caller passes a list created with a larger budget. That contradicts the module/API claim that the compiled screen own `DrawBudget` is the frame bound and that `DrawList` will fail closed if the list was not created against it. It also means a compiler or integration mistake in the baked budget is silently bypassed instead of being observable. Please validate the list budget against `screen.budget` at entry (or explicitly cap writes against the screen budget) and add the inverse-mismatch test: a two-panel screen with a one-rectangle screen budget rendered into a larger list must be refused and rolled back.

ambroise-leclerc: Confirmed - this is a real defect, not a documentation gap. `render()` never read `screen.budget`, so the module comment claimed a bound the code did not apply, and your consequence is exact: a screen declaring room for one rectangle drew two whenever the caller passed a roomier list, and a mistake in a baked budget was bypassed rather than observed.

Fixed in d91d125. Both budgets are enforced now and for different reasons: `DrawList` fails closed against the budget it was created with, and `render()` additionally holds the frame to the screen's own by measuring what it added since entry. Measured rather than predicted - a rectangle costs four vertices, six indices, and either extends the current command or starts one - so this does not carry a second copy of arithmetic `DrawList` already owns.

A larger list stays legitimate, since one list may carry several screens; what is refused is a *frame* exceeding the ceiling its own screen declares. `medui-screen-own-budget` is the inverse test you asked for: a two-panel screen with a one-rectangle screen budget, rendered into a list sized for 128, refused and rolled back.

tests/medui/ScreenRuntimeTests.cpp
ambroise-leclerc: You are right, and the example settles it: work proportional to `node.bounds.width` passes both scenarios as written. Rendering one screen twice proves determinism and nothing more; duplicating identical nodes doubles any per-node cost whatever it depends on. I had written two tests that could not fail for the reason I claimed they existed.

Three changes in d91d125, taking your third option and part of the first:

- `medui-screen-work-is-data-independent` renders two screens of **equal node count** sharing nothing else - 2x2 rectangles against 400x150 ones, different colour tokens, different payload mix - and asserts the step counts match. That is the case that would catch width-proportional work.
- `frame->rects <= frame->nodes` is asserted where a frame is read: the per-node cap this implementation actually has, stated rather than described.
- The module comment now says what the three tests establish **and what they cannot**: `steps` is self-reported, so a future inner loop that performed work without incrementing it would leave all three green. What they do establish is that the work does not vary with a node's payload or geometry at equal node count, and scales linearly with the count.

On the type-level cap: I agree it is the instrument that would make the per-node half a fact rather than a measurement, and TrustSC's `TextRuntime::<MAX_GLYPH_COMMANDS_PER_RUN>` is the shape. I have not added one here because every component this runtime draws emits exactly one rectangle, so the cap would be `1` and would constrain nothing that could vary. It belongs with the first component whose geometry is variable - #17's ground - and the module comment records that rather than leaving it as something I considered and dropped.

tests/medui/ScreenRuntimeTests.cpp
ambroise-leclerc: Correct, and it is the defect class this repository has a note about: a parity claim tested on one side only. My scenario compared two renders *inside one binary*, so each CI leg compared its own output with itself and both passed while potentially emitting different bytes. That is repeatability, not the cross-toolchain identity #199 asked for.

Taken as your third option, in d91d125. `medui-screen-vertex-bytes` freezes the expected values: all four vertices field by field - position, uv, packed colour, mode - and all six indices, checked by MSVC and GCC against the same constants. Every byte is covered, because `UiVertex` is statically asserted to be exactly 24 bytes at fixed offsets with no padding.

The colour constant is the part worth reading: `Theme.Colors.TopbarBackground` is `{0.82, 0.84, 0.86, 1.0}` linear, and this runtime quantises without a transfer function, so `{209, 214, 219, 255}` is pinned. A change to the governed table, to the rounding, or to a compiler's float conversion now fails on the leg that differs rather than producing a slightly different frame nobody compares.

I chose field-level over a digest deliberately: a failure names which vertex and which value, where a digest mismatch says only that something moved.

CMakeLists.txt
ambroise-leclerc: Fixed in d91d125. `mdux.medui.screen` is in the governed module table with both its files.

While there I corrected a second row the same finding applies to: `MduXMeduiLib` still said "emitters (#197) and `mdux-meduic` (#198) land on this same target", which was true when it was written and stopped being true two PRs ago. It now describes what the target holds.

tests/framework/CountingAllocations.hpp
```diff
@@ -0,0 +1,169 @@
...
+    if (pointer == nullptr) {
+        throw std::bad_alloc{};
+    }
+    return pointer;
+}
+
+void* operator new[](std::size_t size, std::align_val_t alignment) {
+    return ::operator new(size, alignment);
+}
+
```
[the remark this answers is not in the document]
ambroise-leclerc: Correct, and worth more than a softened comment: the two aligned nothrow forms - `operator new(size, align_val_t, nothrow_t)` and its array variant - were genuinely absent, so there was an uncounted path in the one file whose entire job is that no path is uncounted. A `std::nothrow` over-aligned allocation would have gone unnoticed and every scenario would have stayed green.

Added in d91d125, with the matching deletes, and the comment at the new pair records why they were missing rather than just that they now exist.

src/medui/Screen.cpp
```diff
@@ -0,0 +1,106 @@
...
+            // Visited and left undrawn. The module comment says which components these are and why
+            // each one's appearance is not decidable from a compiled screen alone.
+            ++stats.deferred;
+            continue;
+        }
+
+        const auto colour = resolveColorToken(panel->colorToken);
+        if (!colour.has_value()) {
+            return refuse(ScreenError::UnknownColorToken);
+        }
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed, and the schema makes the case for you: `checkColor()` distinguishes `MalformedToken` from `UnknownToken` and its comment says why - "a malformed name is an emitter defect while an absent one is a table that does not define it". Collapsing them here threw away a distinction the layer below deliberately preserves, and sent an integrator to the wrong place.

`ScreenError::MalformedColorToken` added in d91d125, mapped from `colour.error()`, with its own `describe()` line and a scenario covering a token that is not of the form at all.

include/mdux/medui/Screen.cppm
```diff
@@ -0,0 +1,116 @@
...
+import mdux.draw;
+import mdux.medui.schema;
+
+export namespace mdux::medui {
+
+/// Why a frame was refused. Every one leaves the draw list exactly as it was found.
+enum class ScreenError : std::uint8_t {
+    UnknownColorToken,  ///< a node names a colour the governed table does not define
+    BudgetExhausted,    ///< a write would exceed the `DrawBudget` the screen declares
+};
```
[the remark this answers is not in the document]
ambroise-leclerc: Same finding from the interface side, and taken the same way: the enum is extended rather than the documentation widened. `UnknownColorToken` now means what it says - well-formed and absent from the governed table - and `MalformedColorToken` covers the other half, with a note at the enum explaining that the two are kept apart because the schema keeps them apart.

AM-L (dismissed):

docs/roadmap.md
```diff
@@ -1,22 +1,24 @@
...
+that generates the screens it draws. Ten of its twelve children have landed — the ADRs,
 the diagnostic registry, the front end, semantic analysis, bounded layout, per-locale text budgets,
 golden references, the canonical package with its C++ emitters, and the `mdux-meduic` compiler with
 its CMake registration — so the compiler now reads a `.medui` file, resolves it to a bounded box
 tree, refuses a box that cannot hold its widest approved translation, says where safety-critical
 content must appear, and writes the result as a byte-compared artifact and as `constexpr` C++ a
 device links without a parser. The first compiled screen is committed under
-`generated/screen/endoscope-monitor/`. #199 is next: the allocation-free runtime that draws one.
+`generated/screen/endoscope-monitor/`, and the governed runtime draws one without allocating. #200
+is next: `mdux-medui-check`, the authoring-side checker.
```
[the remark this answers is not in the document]
ambroise-leclerc: Taken in 8b0e54e. The introduction said the emitters and the runtime had landed while the divergence table two sections down still called them "the back end" that was ahead - the document contradicting itself, which is worse than either version alone.

The row now says what is in the tree and what actually remains: #200's authoring checker and #201's first end-to-end screen, the latter being what the missing text package blocks.

tests/CMakeLists.txt
```diff
@@ -971,6 +1003,16 @@ file(GENERATE
...
 
+# The same scan over the screen runtime's objects (issue #199). The profile is named `ml-noheap` for
+# historical reasons and what it forbids is not ML-specific - the allocation family and the throwing
+# operators - so it is reused rather than copied under a second name. Renaming it would touch the ML
+# wiring in a PR about screens; worth doing, separately.
+set(mdux_screen_object_list "${CMAKE_CURRENT_BINARY_DIR}/screen_objects_$<CONFIG>.txt")
+file(GENERATE
+    OUTPUT "${mdux_screen_object_list}"
+    CONTENT "$<JOIN:$<FILTER:$<TARGET_OBJECTS:MduXCore>,INCLUDE,/medui/>,\n>"
+)
```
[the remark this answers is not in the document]
ambroise-leclerc: Declining this one: the premise is wrong, and the evidence is in this PR's own CI run.

**`$<TARGET_OBJECTS:tgt>` is not restricted to object libraries.** The manual shipped with the CMake this project requires says it "would **typically** be used on object library targets" - a recommendation, not a constraint. The sources quoted in the analysis are the 3.20 documentation; this repository sets `cmake_minimum_required(VERSION 4.0.0)`.

**And it demonstrably works here.** `$<TARGET_OBJECTS:MduXCore>` has driven `ml.noheap.symbolScan` since #63 and `governed.noThrow.symbolScan` since #116, both of which have been green on every run since. If the expression were invalid, generation would fail and nothing in the repository would build - rather than 587 tests passing.

From the Linux leg of `d91d125`:

```
584/587 Test #584: screen.noheap.symbolScan ... Passed
585/587 Test #585: ml.noheap.symbolScan ...... Passed
```

There is a real hazard adjacent to this one, and it is worth saying that it is already covered rather than leaving the impression nobody thought about it: a scan over an **empty** object list would pass by checking nothing. `cmake/MduXNoHeapScan.cmake` refuses that with a `FATAL_ERROR` whose message names the `$<TARGET_OBJECTS:...>` expression as the thing to check. So `screen.noheap.symbolScan` passing is positive evidence that the expression resolved to a non-empty set of objects and that the `/medui/` filter matched them - which is the property that would have been worth doubting.

AM-L (approved):

PR #233, opened 2026-08-22
Check one .medui file, and say what a file alone cannot answer [#200]

### LINKED ISSUE #200: S11 — mdux-medui-check

ambroise-leclerc:
Closes #200. The surface an author and an agent reach for first: point it at a file, read the
diagnostics, fix them.

```
mdux-medui-check path/to/screen.medui [--format=json|text]
```

Writes nothing. Exits non-zero when anything of error severity was found, so it drops into a
pre-commit hook or an agent loop as it stands, and `--format=json` emits the envelope every MduX tool
emits (#118).

## The design question, and the answer I would like checked

Two stages cannot be fed by a single file. **Text keys** are checked against the approved locales a
recipe names, and **text budgets** need the font those locales were baked into. With no recipe there
are neither.

Running the key check anyway would report every `t("STR-KEY")` as absent from every approved locale —
a vacuous truth dressed as a finding, and the fastest way to teach an author that this tool's output
is noise. So the check does not run, and **the run says so**: a `Severity::Note` carrying `MDC001`,
travelling in the same envelope as the findings. A screen with no `surface:` gets `MDC002` for the
same reason — layout, overflow and golden bounds went unchecked, and that is a gap in what can be
checked rather than a finding about the file.

Notes do not fail a run. What they buy is that a clean result is visibly **partial** rather than
silently so, and that an agent can key off the code to decide whether to re-run through
`mdux-meduic` with a recipe. A checker whose clean run is silently partial is worse than none,
because it converts "I did not look" into "nothing is wrong".

## One change below the tool

`analyze()` no longer reports a text key as absent when it was given **no** approved locales, because
there is nothing for it to be absent from. That is arguably what it should always have done; what
made it safe to change now is that the caller which must not tolerate the gap — the compiler driver —
already refuses a recipe whose screen draws text and declares no locales (#198). A scenario pins the
new behaviour, and the existing three-locale scenario is untouched.

## Tests

`tests/medui/CheckTests.cpp` drives the same rejection corpus the parser and analyzer suites use by
call — `rejected-unterminated-string`, `rejected-nested-row`, `rejected-duplicate-id`,
`rejected-safety-without-requirement` — so a code that stops surfacing to an author stops a test.
That corpus being real files rather than string literals is exactly what makes this possible, which
is what #192 said it was for.

Two scenarios spawn the executable, for the reason #198's does: nothing at call level shows that the
tool exits non-zero on a finding and **zero on a run whose only output is notes**. The second half is
the one that would have been easy to get wrong — an exit status keyed off the presence of diagnostics
rather than their severity fails every clean run.

## The skill

`medui-authoring` said "there is no tooling shortcut yet", which #200's acceptance requires it to stop
saying. It now documents the command, what it covers, and — in the same breath — the two things it
cannot, with the instruction to compile through a recipe for those.

## Regulatory impact

No device code changes; the tool is host-only and writes nothing. What improves is where an authoring
error is found: at the file, by the author, in the envelope an agent can act on, rather than at a
build that compiles the whole screen. What has not changed is what counts as an error — the checker
runs the same stages with the same codes, and reports strictly less than a compile because it knows
strictly less.

Roadmap: #15 is S1-S11 done, #201 next and last.

## Summary by CodeRabbit

* **New Features**
  * Added the `mdux-medui-check` command to validate standalone `.medui` files.
  * Supports text and JSON diagnostic output, clear success or failure status, and actionable validation messages.
  * Checks syntax, semantic rules, safety annotations, and layout when a surface is provided.

* **Bug Fixes**
  * Improved locale and text-key validation, including clearer handling of skipped checks and source locations.

* **Documentation**
  * Updated authoring guidance and roadmap details to reflect the available compiler and validation workflow.

* **Tests**
  * Expanded coverage for checker behavior, diagnostics, command-line execution, and semantic validation.

### DIFF
--- a/.agents/skills/medui-authoring/SKILL.md
+++ b/.agents/skills/medui-authoring/SKILL.md
@@ -10,7 +10,7 @@ governs the *authoring* of a `.medui` screen; for the baking mechanics behind it
 `evidence-pipeline`, and for the compliance framing of a safety-critical node see
 `regulatory-citations`.
 
-## Status: the front end through golden references is implemented; emission and runtime are not
+## Status: the compiler is complete front to back; one end-to-end screen remains
 
 **A `.medui` lexer, AST, parser, semantic analyzer, integer-only bounded layout solver, text-budget
 check, and golden-reference pass exist** in the host-tools zone (`tools/medui/`). The diagnostic
@@ -23,9 +23,22 @@ against the font package's restricted charset (#195); and the golden pass applie
 `@safety_critical` rules and derives the reference set #16's verifier will consume (#196). The AST
 keeps names unresolved.
 
-**There is still no compiler, no emitter and no runtime.** The HTML/CSS path that used
-to stand in for one - `UiFileWatcher::loadContent()`, which sniffed a file extension and stored
-the file as a string, with no parsing, layout or rendering behind it - was deleted by
+**The back end exists now**, and this paragraph used to say the opposite. `mdux-meduic` compiles a
+screen from a recipe into `generated/screen/<id>/` — the compiled package, the golden sidecar and a
+bake report, byte-compared across toolchains (#198). Two emitters render that package as a `.cppm`
+and a `.hpp` carrying `static_assert(screen.validate().has_value())`, so a malformed screen is a
+build error rather than a startup failure (#197). A governed runtime draws one without allocating
+(#199), and `mdux-medui-check` validates a single file (#200).
+
+Two limits are worth knowing before you write a screen. **No text package is baked in this
+repository yet**, so a screen carrying `t("STR-KEY")` cannot be compiled end to end — that is #201,
+the last child of the epic. And the runtime draws a `Panel`; every other component is visited,
+counted in `FrameStats::deferred` and left undrawn, because text needs that package and live-data
+components have no geometry until the frame does.
+
+The HTML/CSS path that used to stand in for all of this - `UiFileWatcher::loadContent()`, which
+sniffed a file extension and stored the file as a string, with no parsing, layout or rendering
+behind it - was deleted by
 [issue #127](https://github.com/ambroise-leclerc/MduX/issues/127).
 
 What exists now is the layer a `.medui` compiler will target: `mdux.draw` describes a frame as
@@ -36,7 +49,9 @@ bounded vertex, index and command buffers with a compiler-computed budget, and
 contract in `medui-conformance.toml`. This skill records MduX status and integration; canonical
 grammar, component semantics, diagnostics, and portable guidance live in
 [`Compliatory/MedUI` at `d5136a8`](https://github.com/Compliatory/MedUI/tree/d5136a8518bd499760ecff2aad215d3721329f20).
-Do not write a `.medui` file expecting it to build anything in MduX until the compiler waves land.
+A `.medui` file builds something in MduX today: register it with `mdux_compile_screen()` and it
+becomes a committed, byte-compared artifact plus generated C++ a device links. What it cannot yet do
+is carry text, for the reason above.
 
 ## Grammar shape
 
--- a/.agents/skills/medui-authoring/SKILL.md
+++ b/.agents/skills/medui-authoring/SKILL.md
@@ -112,11 +112,27 @@ verifier checks against. Rules:
 
 ## Checking a file without a full build
 
-`mdux-medui-check path/to/screen.medui` (issue #200) will validate a single file and print
-diagnostics — once it exists. The parser, analyzer, solver, budget check, and golden pass it will
-call already detect syntax errors, structural violations, unknown dictionary/theme names, hardcoded
-text in localizable fields, wrong field value domains, incomplete locale keys, layout overflow, a
-box that cannot contain its widest approved translation, a dynamic-text source that could escape
-the baked charset, a `@safety_critical` node with no `requirement:`, and an unknown `cv_check`;
-what is missing is the command that exposes them. Until it lands, review a `.medui`
-file by hand against this grammar and the component table above.
+```
+mdux-medui-check path/to/screen.medui [--format=json|text]
+```
+
+Validates one file and writes nothing. Exits non-zero when anything of error severity was found, so
+it drops into a pre-commit hook or an agent loop as it stands. `--format=json` emits the same
+envelope every MduX tool emits (#118), so a finding is `{file, line, column, code, severity,
+message, fixHint}` rather than prose to parse.
+
+**What it checks:** the grammar, the closed component dictionary, each field's value domain,
+hardcoded strings in localizable fields, theme tokens against the governed table, the
+`@safety_critical` rules including an unknown `cv_check` — and, when the file declares a `surface:`,
+bounded layout with its overflow and containment rules plus the golden set that follows.
+
+**What it cannot check, and says so.** Text keys and text budgets need the approved locales a recipe
+names, and a single file names none. Rather than reporting every `t("STR-KEY")` as absent from every
+locale — a vacuous truth dressed as a finding — the checker skips those and emits a note, `MDC001`.
+A screen with no `surface:` gets `MDC002` for the same reason: layout, overflow and golden bounds
+went unchecked. Notes do not fail the run; they are what makes a clean result visibly *partial*
+rather than silently so.
+
+For the two it cannot cover, compile the screen through `mdux-meduic` with a recipe carrying a
+`[text]` table: that checks every key against every approved locale, and every box against the
+widest translation of the text it holds.
--- a/.agents/skills/medui-authoring/SKILL.md
+++ b/.agents/skills/medui-authoring/SKILL.md
@@ -127,7 +127,7 @@ verifier checks against. Rules:
 
 ## Checking a file without a full build
 
-```
+```console
 mdux-medui-check path/to/screen.medui [--format=json|text]
 ```
 
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -10,7 +10,7 @@ backlog that closes the gap. Waves 1, 2 and 3 have shipped — the renderer draw
 pixel, zero-SOUP ML inference is in the tree, and the documentation has been rebuilt from
 what the build actually produces. Track C's authoring story is what remains: #14 closed
 Wave 4 with the font and text pipeline, and #15 is underway in Wave 5 with the compiler
-that generates the screens it draws. Ten of its twelve children have landed — the ADRs,
+that generates the screens it draws. Eleven of its twelve children have landed — the ADRs,
 the diagnostic registry, the front end, semantic analysis, bounded layout, per-locale text budgets,
 golden references, the canonical package with its C++ emitters, and the `mdux-meduic` compiler with
 its CMake registration — so the compiler now reads a `.medui` file, resolves it to a bounded box
@@ -62,7 +62,7 @@ Wave 1 · shipped v0.2.0     #7 (done)   #11 (done)  #19 (S4–S6 open)
 Wave 2 · shipped v0.3.0     #8 (done)   #9 (done)   #12 (done)
 Wave 3 · shipped v0.4.0     #10 (done)  #13 (done)  #18 (done)
 Wave 4 · shipped v0.5.0     #14 (done)
-Wave 5 · in progress        #15 (S1–S10 done · S11 next)
+Wave 5 · in progress        #15 (S1–S11 done · S12 next)
 Wave 6                      #16  #17
 ```
 
@@ -263,7 +263,7 @@ charset table — and the compiler rejects any format that could escape it, whic
 
 _Unblocks #15_
 
-#### #15 — `.medui` compiler & build integration · **In progress · Wave 5 · 10/12**
+#### #15 — `.medui` compiler & build integration · **In progress · Wave 5 · 11/12**
 
 The schema module is imported by both the device runtime and the host compiler — one
 definition, shared. The runtime never sees the parser, which lives in a host-only tool.
@@ -287,8 +287,8 @@ happened to read.
 - #197 S8 Canonical package and C++ emitters · _closed (PRs #220, #221)_
 - #198 S9 CMake integration and the `mdux-meduic` host tool · _closed (PRs #225, #231)_
 - #199 S10 Allocation-free screen runtime · _closed (PR #232)_
-- #200 S11 `mdux-medui-check` · **next**
-- #201 S12 First end-to-end screen
+- #200 S11 `mdux-medui-check` · _closed (PR #233)_
+- #201 S12 First end-to-end screen · **next**
 
 _Blocks #16, #17_
 
@@ -395,5 +395,5 @@ lint — is real, but it is narrower. The wording is fixed in #40 and #38:
 ---
 
 _Epic status verified at `develop` @ `45eecbe` · 22 August 2026_
-_13 epics · 9 delivered · Waves 1–4 shipped · Wave 5 in progress (#15 at 10/12) · no enforcement gaps outstanding_
+_13 epics · 9 delivered · Waves 1–4 shipped · Wave 5 in progress (#15 at 11/12) · no enforcement gaps outstanding_
 _All epics on GitHub_
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,7 +1,7 @@
 # MduX → TrustSC parity roadmap
 
 > Backlog · ambroise-leclerc/MduX · updated 22 August 2026
-> Epic status verified against `develop` @ `45eecbe` · 22 August 2026 · `#15` current through `#199`.
+> Epic status verified against `develop` @ `a5cc41f` · 22 August 2026 · `#15` current through `#200`.
 > The divergence table below was last re-verified on 17 August 2026 and is not re-checked here.
 
 MduX (C++23 / Vulkan) and TrustSC (Rust) target the same problem — a medical-device UI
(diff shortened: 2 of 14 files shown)


tools/medui/Semantic.cpp
```diff
@@ -204,6 +204,15 @@ class Analyzer {
     }
 
     void analyzeTextKey(const ast::Value& value) {
+        // No approved locale to check against means there is nothing to check, and saying "absent
+        // from every approved locale" when there are none is a vacuous truth dressed as a finding.
+        // The caller that must not tolerate the gap is the compiler driver, which refuses a recipe
+        // whose screen draws text and declares no locales (#198); `mdux-medui-check` runs without a
+        // recipe on purpose and reports the gap as a note (#200).
+        if (inputs_.textPackages.empty()) {
```
ambroise-leclerc: **[P2] Make skipped locale validation an explicit analyzer mode.** An empty `textPackages` span previously represented an approved set containing no locales; this change silently reinterprets it as permission to skip key validation. That affects every caller of the semantic module, while the compensating `MDC001` note is emitted only by `checkScreen()`. A direct caller can therefore receive a successful semantic result for arbitrary `t("...")` keys with no indication that the documented locale-completeness check did not run. The separate `needsTextBudget()` gate currently protects `mdux-meduic`, but it does not preserve the analyzer contract or protect future callers if the component dictionary and predicate diverge. Please add an explicit validation policy/mode to `SemanticInputs` (for example, required versus intentionally skipped), keep the normal analyzer path fail-closed, and have only `mdux-medui-check` select the partial-validation mode. Add a test that distinguishes these two modes.

.agents/skills/medui-authoring/SKILL.md
```diff
@@ -112,11 +112,27 @@ verifier checks against. Rules:
...
-`mdux-medui-check path/to/screen.medui` (issue #200) will validate a single file and print
-diagnostics — once it exists. The parser, analyzer, solver, budget check, and golden pass it will
-call already detect syntax errors, structural violations, unknown dictionary/theme names, hardcoded
-text in localizable fields, wrong field value domains, incomplete locale keys, layout overflow, a
-box that cannot contain its widest approved translation, a dynamic-text source that could escape
-the baked charset, a `@safety_critical` node with no `requirement:`, and an unknown `cv_check`;
-what is missing is the command that exposes them. Until it lands, review a `.medui`
-file by hand against this grammar and the component table above.
+```
+mdux-medui-check path/to/screen.medui [--format=json|text]
```
ambroise-leclerc, line 131: **[P2] Reconcile the skill status with this new checker section.** This command is now documented as implemented, but the same skill still says at lines 13, 26, and 39 that emission/runtime are absent, that there is no compiler/emitter/runtime, and that a `.medui` file cannot build anything. Those statements were superseded by #197, #198, and #199 and now directly contradict this added section. Repository instructions route agents through this skill as an operational source of truth, so stale status text can cause incorrect implementation and review decisions. Please update the heading and introductory status paragraphs to describe the compiler, emitters, governed runtime, and checker that exist today, leaving only #201 as pending.

docs/roadmap.md
```diff
@@ -10,7 +10,7 @@ backlog that closes the gap. Waves 1, 2 and 3 have shipped — the renderer draw
 pixel, zero-SOUP ML inference is in the tree, and the documentation has been rebuilt from
 what the build actually produces. Track C's authoring story is what remains: #14 closed
 Wave 4 with the font and text pipeline, and #15 is underway in Wave 5 with the compiler
-that generates the screens it draws. Ten of its twelve children have landed — the ADRs,
+that generates the screens it draws. Eleven of its twelve children have landed — the ADRs,
```
ambroise-leclerc, line 13: **[P2] Update all current-state prose when advancing this count to 11/12.** The paragraph immediately below still says `#200 is next`, the UI-authoring table still lists the checker as future work, and the verification banner says the epic is current only through `#199`. Those statements conflict with this changed count and with the correctly updated backlog entry naming #200 closed and #201 next. Please update all four locations together so this current-state roadmap is internally consistent.

ambroise-leclerc (commented):
Review result: changes needed before merge. I found one behavioral contract issue and three documentation/traceability gaps. The change is potentially safety-relevant: it is host-only and does not alter device runtime behavior, but it changes which authoring validation failures are surfaced for governed screen inputs. Please record that classification and rationale in the PR impact statement.

One additional documentation item cannot be attached inline because the file is unchanged: `docs/architecture.md` calls itself the build-verified current-state inventory, but its `MduXMeduiLib` row does not include the new `mdux.tools.medui.check` module or `mdux-medui-check` executable. Please add #200 to that host-tool entry.

Verification: all current GitHub checks are green, and `git diff --check origin/develop...HEAD` passes. I could not run a supported local build because this workspace is macOS/AppleClang, outside the repository supported toolchains.

tools/medui/Semantic.cpp
ambroise-leclerc: [P2] Confirmed, and this is the finding with teeth. I reinterpreted an existing meaning - an empty `textPackages` span *is* an approved set containing no locale - and did it for every caller of the module, while the compensating `MDC001` note is emitted by `checkScreen()` alone. A direct caller would have got a clean semantic result for arbitrary keys with nothing in the result to say the documented check had not run.

Your point about the gate is the part I had not thought through: `mdux-meduic` is protected by `needsTextBudget()`, which is a *different* predicate over a different list of fields. It happens to cover the same screens today, and nothing keeps the two aligned if the dictionary grows a third dynamic-text field or a new key-valued domain.

Fixed in f115d3b as you suggest. `SemanticInputs` carries a `LocalePolicy`, defaulting to `Required` - which restores the previous behaviour exactly, so a caller who wanted the check skipped and forgot to say so gets a finding rather than a clean result. `Skipped` is selected by `mdux-medui-check` and by nothing else, and the enum documents why the mode is asked for rather than inferred.

`medui-semantic-locale-policy` analyzes one screen under both policies and asserts the difference, so the two modes cannot quietly become one.

.agents/skills/medui-authoring/SKILL.md
ambroise-leclerc: [P2] Confirmed, and worse than stale - actively contradicting the section this PR adds. Lines 13, 26 and 39 said the front end was implemented "and emission and runtime are not", that there is "still no compiler, no emitter and no runtime", and that a `.medui` file will not build anything "until the compiler waves land". All three were superseded by #197, #198 and #199.

That matters more here than in a document: repository instructions route agents through this skill as an operational source of truth, so an agent reading it today would have been told to hand-review a file against the grammar while the tool sat in the tree.

Rewritten in f115d3b. The heading and both paragraphs describe what exists - the compiler, the two emitters, the governed runtime, the checker - and name the two limits that remain rather than leaving them to be discovered: no text package is baked in this repository, so a screen carrying `t("STR-KEY")` cannot be compiled end to end, and the runtime draws a `Panel` while counting every other component as deferred.

docs/roadmap.md
ambroise-leclerc: [P2] Confirmed, all four. Fixed in f115d3b: the banner reads through #200, the introduction says #201 is next and last, the UI-authoring row names the checker as landed, and the backlog entry was already correct.

Worth recording how two of those got past me rather than just that they did. They were edits I wrote in the same commit as the count, using string replacements with no assertion - so when the surrounding prose had changed under them, they matched nothing, changed nothing, and reported nothing. That is the defect class this repository keeps producing in a different costume, and the replacements now assert. It is also the second time in this epic that a silent no-op edit reached a review; the first was `toOptions()` in #231.

tools/medui/Check.cpp
```diff
@@ -0,0 +1,120 @@
...
+    //    the note recorded below for what that leaves uncovered.
+    std::vector<std::string_view> themeTokens;
+    themeTokens.reserve(ms::themeColors.size());
+    for (const ms::ThemeColor& colour : ms::themeColors) {
+        themeTokens.push_back(colour.token);
+    }
+
+    // The one caller that asks for the partial mode, and the reason the mode is a request rather
+    // than something inferred from an empty package list: this file may belong to no recipe at all.
+    const SemanticResult semantic = analyze(screen, file, {.themeTokens = themeTokens, .textPackages = {}, .locales = LocalePolicy::Skipped});
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed, and the way this one arose is worth naming: the comment was true when I wrote it and false by the end of the same commit. `LocalePolicy` is what made an empty package list stop meaning "skip", and the comment three lines above the call still described the behaviour the commit had just removed - a comment describing behaviour the code does not have, produced in the act of fixing the behaviour it described.

Corrected in c0f0310, and it now says which of the two facts does the work: an empty package list still means every key is absent, and the skip happens because the call asks for `LocalePolicy::Skipped`.

tests/medui/SemanticTests.cpp
```diff
@@ -340,14 +334,60 @@ const mdux::spec::Register semanticHardcodedTextList{
...
+                                                                      "policy.medui",
+                                                                      md::SemanticInputs{.themeTokens = themes, .textPackages = none});
+                      checks.expect(count(required, md::Code::UnknownTextKey) == 1, "the default policy reports a key with no locale to resolve in");
+
+                      const md::SemanticResult skipped = md::analyze(
+                          *md::parse(source, "policy.medui").screen,
+                          "policy.medui",
+                          md::SemanticInputs{.themeTokens = themes, .textPackages = none, .locales = md::LocalePolicy::Skipped});
+                      checks.expect(skipped.ok(), "and the explicit mode skips it");
+                      checks.expect(count(skipped, md::Code::UnknownTextKey) == 0, "reporting no key findings at all");
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed. `*md::parse(...).screen` on a `nullopt` is undefined behaviour, so a fixture that stopped parsing would crash rather than report a failed expectation - and the helper at the top of the same file already guards exactly this, which makes it an inconsistency as well as a hazard. Parsing twice was the other half of it.

Fixed in 54d7dd2: parsed once, checked, reused by both calls.

That correction missed my first push, and the reason is worth a line since it is the same shape as the finding above. clang-format had reflowed those two calls, so my edit matched nothing - the assertion caught it, but the commit had already gone out with three of the four corrections. Hence the second commit.

docs/roadmap.md
```diff
@@ -1,7 +1,7 @@
 # MduX → TrustSC parity roadmap
 
 > Backlog · ambroise-leclerc/MduX · updated 22 August 2026
-> Epic status verified against `develop` @ `45eecbe` · 22 August 2026 · `#15` current through `#199`.
+> Epic status verified against `develop` @ `a5cc41f` · 22 August 2026 · `#15` current through `#200`.
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed - the document reported two verification snapshots, header and footer disagreeing. Fixed in c0f0310, both now name `a5cc41f`.

That is the third self-inconsistency this roadmap has produced in as many PRs. The pattern is always the same: a status line updated in one place and not the two others that restate it. Worth considering whether the verification banner should be generated rather than hand-maintained, but that is a separate change from this PR.

.agents/skills/medui-authoring/SKILL.md
```diff
@@ -112,11 +127,27 @@ verifier checks against. Rules:
...
 
-`mdux-medui-check path/to/screen.medui` (issue #200) will validate a single file and print
-diagnostics — once it exists. The parser, analyzer, solver, budget check, and golden pass it will
-call already detect syntax errors, structural violations, unknown dictionary/theme names, hardcoded
-text in localizable fields, wrong field value domains, incomplete locale keys, layout overflow, a
-box that cannot contain its widest approved translation, a dynamic-text source that could escape
-the baked charset, a `@safety_critical` node with no `requirement:`, and an unknown `cv_check`;
-what is missing is the command that exposes them. Until it lands, review a `.medui`
-file by hand against this grammar and the component table above.
+```
```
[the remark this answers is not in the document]
ambroise-leclerc: Taken in c0f0310 - the fence carries `console` now.

AM-L (dismissed):

tools/medui/Semantic.cppm
```diff
@@ -68,9 +68,28 @@ struct ComponentRule {
  * analyzer reads only each package's locale and run IDs. Theme tokens are full names such as
  * `Theme.Colors.Title`; their concrete colour values belong to the later emitter stage.
  */
+/**
```
ambroise-leclerc, line 64: **[P3] Keep each Doxygen block adjacent to the declaration it documents.** The `SemanticInputs` block at lines 64–70 is now followed by this separate `LocalePolicy` block and enum, so it is detached from `SemanticInputs`; Doxygen associates only the immediately preceding documentation block with a declaration. Move the external-name-tables block below `LocalePolicy`, immediately before `struct SemanticInputs`, while leaving this policy block directly above the enum. That preserves API documentation for both newly separated concepts.

tests/medui/CheckTests.cpp
```diff
@@ -0,0 +1,281 @@
...
+        }
+        rendered += diagnostic.code.empty() ? std::string{"<none>"} : diagnostic.code;
+    }
+    return rendered.empty() ? std::string{"<nothing>"} : rendered;
+}
+
+/// A directory this scenario owns, removed when it is done with it.
+class TemporaryDirectory {
+public:
+    explicit TemporaryDirectory(std::string_view name) : path_{std::filesystem::temp_directory_path() / name} {
```
ambroise-leclerc: **[P3] Give each CLI scenario a unique temporary directory.** This constructor maps a fixed name directly under the system temporary directory, removes any existing directory at construction, and removes it again at destruction. Two concurrent `medui_tools_spec` executions therefore share a path such as `mdux-medui-check-cli`; either instance can delete files while the other process is using them, producing intermittent subprocess failures unrelated to the checker. Append a process-specific or otherwise collision-resistant suffix and retain the existing RAII cleanup. The caller-visible interface can remain unchanged.

ambroise-leclerc (commented):
Follow-up review after the correction commits: the main behavioral fix is correct. `LocalePolicy::Required` restores fail-closed semantic analysis, only `mdux-medui-check` selects `Skipped`, and the new scenario distinguishes both modes with a guarded parse result. The skill status, roadmap prose, roadmap snapshot, command-fence language, and stale implementation comment were also corrected.

Changes are still needed in three places that cannot be attached inline:

1. **[P2] Update the PR description to match the corrected contract.** The section titled `One change below the tool` still says `analyze()` no longer reports a key as absent when passed zero approved locales. That is now false: under the default `LocalePolicy::Required`, zero packages still report `UnknownTextKey`; only an explicit `Skipped` policy suppresses locale diagnostics. Please replace that explanation with the explicit-policy design. The regulatory section should also record the requested classification verbatim as **potentially safety-relevant**, with the existing rationale that this changes host-side authoring diagnostics but not device runtime behavior.

2. **[P2] Update `docs/architecture.md`.** Its authoritative `MduXMeduiLib` inventory still stops at #198 and omits the new `mdux.tools.medui.check` module and `mdux-medui-check` executable. This was part of the previous overall review and remains unresolved.

3. **[P3] Update ADR-011.** Its consequences section still says authoring feedback requires a build until #200 provides a single-file path. With this PR landing that path, the present-tense limitation is stale; revise it to describe the implemented checker and its locale/text-budget limits.

All current substantive CI checks pass on `54d7dd2`, including GCC 16, MSVC, sanitizers, documentation/compliance lint, CodeQL, and security analysis. `git diff --check origin/develop...HEAD` also passes. I did not run a local build because this workspace is macOS/AppleClang, outside the repository supported toolchains.

tools/medui/Semantic.cppm
ambroise-leclerc: [P3] Confirmed. Inserting `LocalePolicy` between the `SemanticInputs` block and the struct left that block documenting the enum - Doxygen associates only the immediately preceding block with a declaration, so `SemanticInputs` lost its documentation and `LocalePolicy` acquired a description of external name tables.

Same failure mode as the missing `@file` that `CONTRIBUTING.md` measures: not ignored documentation, but documentation silently attached to the wrong thing. Fixed in 79e3c90 - the policy block stays above the enum, the external-name-tables block sits above its struct again.

tests/medui/CheckTests.cpp
ambroise-leclerc: [P3] Confirmed. A fixed name straight under the system temporary directory, removed at construction *and* destruction, so two concurrent `medui_tools_spec` runs share the path and either can delete files the other process is using - producing an intermittent subprocess failure that looks like the checker and is not.

Fixed once rather than three times, because your finding is on the third copy: `EmitTests`, `CompileTests` and `CheckTests` were each carrying that class, and each carrying the defect with it. It now lives in `tests/framework/TemporaryDirectory.hpp` beside `CountingAllocations.hpp`, with a random suffix on the name - `std::random_device` rather than a process id, since that needs no platform header in a tree that reaches the standard library through `import std`.

One consequence worth mentioning: the shared class creates the directory it owns, where the `EmitTests` copy did not, so four `create_directories` calls its scenarios were making by hand went away with it.

AM-L (approved):

Issue #235, opened 2026-08-22
Bake a text package, so a screen can carry text


ambroise-leclerc:
No text package is baked in this repository, so **no screen carrying `t("STR-KEY")` can be compiled
end to end** — and nothing tracks that. It surfaced repeatedly through #197–#201 as "the text half",
always attributed to whichever issue was open at the time. It belongs to none of them.

## What exists and what does not

- `mdux-textbake` produces text packages, and is tested (#14, shipped in v0.5.0).
- A **font** package is baked and committed: `generated/font/dejavu-ui/`, approving `en-US`.
- No **text** package is committed. Nothing registers a `recipes/text/<id>.toml`, so
  `generated/text/` does not exist.

## What that blocks, concretely

- `mdux-meduic` refuses a recipe whose screen draws text and declares no approved locale — correctly,
  since a budget checked against no locale is a claim nobody made. So a screen with a `Label`, a
  `Button`, a `CriticalButton`, a `StatusIndicator` or a `TextInput` cannot be compiled at all.
- The committed screen (`endoscope-monitor`) is therefore text-free, and its safety-critical node is
  a `NumericDisplay` — chosen because it is the one traceable component carrying no text key.
- `mdux-medui-check` reports `MDC001` on every file, because a single file names no approved locale.
  That note is correct, but it will keep being the only outcome for the text checks until a package
  exists to check against.
- The text-budget stage (#195) — measuring every resolved box against the widest approved
  translation — has never run over a committed artifact. It is unit-tested against packages a test
  builds; it has no end-to-end exercise.

## Work

- A `recipes/text/<id>.toml` and the runs it names, baked into `generated/text/<id>/` through
  `mdux_bake_artifact()` like every other artifact kind.
- A screen recipe carrying a `[text]` table that names it, and a screen that draws a `Label`.
- The locale set has to match `generated/font/dejavu-ui`'s approved list exactly, which today is
  `en-US` alone — so either one locale, or the font recipe grows its charset and locale list first.

## Acceptance

- A committed screen carrying at least one `t("STR-KEY")`, compiled by `mdux-meduic` and
  byte-compared on both toolchain legs.
- `checkTextBudgets()` runs over it, so #195 gains an end-to-end exercise rather than only unit
  coverage.
- The "no text package is baked" caveats in `README.md`, `docs/architecture.md`, the
  `medui-authoring` skill and `tests/render/ScreenPixelTests.cpp` are removed rather than reworded.

PR #236, opened 2026-08-22
Record how a release is cut, and what it ships

ambroise-leclerc:
Wave 5 closed with #201, and `docs/roadmap.md` recommends cutting v0.6.0 now. It also names two
things worth settling before the tag rather than during it. This closes the first.

## What it adds

`CHANGELOG.md`, with the 0.6.0 entry written from the merges it actually contains — the twelve
children of #15, the two enforcement gaps #11 closed, and the smaller changes — plus summaries of
0.2.0 through 0.5.0.

## The part I want to flag rather than bury

**The earlier entries are reconstructed from the roadmap's own wave record, not derived from git
history, and the file says so at the top.**

That is not laziness. `v0.2.0` and `v0.5.0` are ancestors of neither `develop` nor `master`, while
`v0.3.0` and `v0.4.0` *are* ancestors of `develop`. A reader checking an entry with
`git log v0.4.0..v0.5.0` gets something that does not mean what they expect, and I would rather a
changelog say which account it rests on than imply a precision the tags cannot support.

The roadmap's note on the release ritual is updated with that finding, since "v0.5.0 is not an
ancestor of develop" turned out to be half the picture — the pattern is mixed, which is worse for a
reader than either consistent answer.

## The known limits are in the entry, deliberately

Three things about 0.6.0 are easy to mistake for defects, so the entry names them: no baked text
package (#235), a runtime that draws a `Panel` while counting every other component as deferred, and
a golden sidecar whose consumer is static rather than rendered. A release note that omitted these
would be describing a different release.

## Since review: the second gap is closed too

The roadmap named two things to settle before tagging. The changelog was the first; the second was
the release procedure, and `docs/release-process.md` now records it — a release branch off `develop`,
the version moved in `CMakeLists.txt`, the artifacts re-baked, the diff reviewed, a merge to `master`,
an annotated tag there, a back-merge.

**The step that would have made the first release red** is the reason this was worth writing rather
than doing from memory: `report.json` records `toolVersion`, so moving the version invalidates all
six committed artifacts and every `evidence.*` comparison until they are re-baked. The ordering is
bump → re-bake → review → tag, and the review has a rule that can be checked instead of felt — every
changed file must differ in `toolVersion` and nothing else, with the command to prove it.

IEC 62304 §5.8 gains the link it should have had: it asks a manufacturer to know what is being
released *including its known anomalies*, which is exactly what the changelog's known-limits section
is for. That clause also claimed "there is no product release yet for this to apply to", which
conflated a library tag with a manufacturer's product release; it now keeps the two apart, and notes
that MduX has cut five of the first kind.

## What this does not do

**It does not cut the tag.** That is yours: the ritual is not readable from the repository, and I do
not know whether 0.6.0 belongs on `develop`, on `master`, or on a branch this history does not show.
Tell me which and I will prepare it; the `unreleased` marker and the date line are the only things
that need to change.

## Regulatory impact

None to the code. A changelog is a traceability surface: for an auditor holding two artefacts, it is
the document that says what differs between them. Its usefulness depends entirely on not overstating
what it knows, which is why the reconstruction note is at the top rather than in a footnote.

## Summary by CodeRabbit

- **Documentation**
  - Added a comprehensive changelog covering releases 0.2.0 through the upcoming 0.6.0 release.
  - Documented compiler capabilities, licensing and tooling updates, known limitations, and historical release context.
  - Updated the release-readiness roadmap with current release workflow and version history details.
  - Marked the Wave 5 interface pipeline as complete, while documenting current text limitations and deferred runtime components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,106 @@
+# Changelog
+
+Written for a reader deciding whether a version does what they need, and for an auditor asking what
+changed between two artefacts they hold. Entries name the issue that owns the work, because the
+issue carries the reasoning and this file carries the outcome.
+
+## How to read the earlier entries
+
+This file starts at 0.6.0. **The entries for 0.2.0 through 0.5.0 are reconstructed from
+[`docs/roadmap.md`](docs/roadmap.md)'s own wave-by-wave record rather than derived from git history**,
+and that distinction is worth one paragraph because a reader may try to check them.
+
+Two of the four earlier tags — `v0.2.0` and `v0.5.0` — are not ancestors of `develop`, so
+`git log v0.4.0..v0.5.0` does not describe what a reader would expect it to. What the roadmap
+records is what each wave closed, and that is the account these entries summarise. Anything more
+precise would need the release ritual written down, which
+[`docs/roadmap.md`](docs/roadmap.md#when-v060-gets-cut) flags as the second thing to settle before
+tagging 0.6.0.
+
+---
+
+## 0.6.0 — unreleased
+
+Wave 5: the `.medui` compiler, end to end. An authored screen becomes a committed artifact, becomes
+`constexpr` C++, and reaches compared pixels.
+
+### The compiler (#15, all twelve children)
+
+- **The boundary and the artifacts** — ADR-011 fixes what a compiled screen is and is not; ADR-012
+  fixes the three files a screen emits and which are committed (#190, amended by #203 to keep the
+  compiled screen locale-free).
+- **Diagnostics** — the shared `MEDUI-E` registry, in the envelope every MduX tool emits (#191).
+- **The front end** — lexer, parser, AST and a fixture corpus of real files (#192); semantic
+  analysis over the closed component dictionary, field value domains, theme tokens and locale-checked
+  strings (#193).
+- **Bounded layout** — integer-only resolution to absolute rectangles, with `Row` flattening and a
+  synthesised background node (#194).
+- **Text budgets** — every resolved box measured against the *widest approved translation*, not the
+  authoring one (#195).
+- **Golden references** — one entry per `@safety_critical` node and per explicitly positioned node,
+  written as a sidecar (#196).
+- **The canonical package and two C++ emitters** — a compiled screen as canonical JSON, and as a
+  `.cppm` and `.hpp` carrying `static_assert(screen.validate().has_value())`, so a malformed screen
+  is a build error rather than a startup failure (#197).
+- **`mdux-meduic`** — the compiler driver and `mdux_compile_screen()`, with the first committed
+  screen under `generated/screen/endoscope-monitor/`, byte-compared on both toolchain legs (#198).
+- **The governed runtime** — a compiled screen becomes draw commands with no allocation, no parsing
+  and work bounded by numbers known before the device runs (#199).
+- **`mdux-medui-check`** — validates one file without a build, and names the two checks a file on its
+  own cannot cover (#200).
+- **The first end-to-end screen** — an authored `.medui` file drawn through Vulkan into an offscreen
+  target and compared pixel by pixel under lavapipe (#201).
+
+### Enforcement that had been asserted but not run (#11)
+
+- `mdux-governed-lint` over governed source, and `governed.noThrow.symbolScan` over the emitted
+  objects. ADR-005 had asserted a lint that did not exist while `src/text/Raster.cpp` — governed,
+  shipped in 0.5.0 — contained `try` and three `catch` clauses. The rasteriser moved to the
+  host-tools zone, and ADR-004 and ADR-005 were rewritten to describe only mechanisms CI runs (#116).
+- A stack-safe PR and post-merge policy (#117).
+
+### Also
+
+- Dual licensing under EUPL-1.2.
+- A file-header lint, after ten files were found in a documentation order `CONTRIBUTING.md` had
+  retired twice (#223).
+- SpecLab pinned to its first tagged release.
+
+### Known limits, stated because they are easy to mistake for defects
+
+- **No text package is baked**, so a screen carrying `t("STR-KEY")` cannot be compiled end to end
+  ([#235](https://github.com/ambroise-leclerc/MduX/issues/235)). The committed screen is text-free
+  and its safety-critical node is a `NumericDisplay`, the one traceable component with no text key.
+- **The runtime draws a `Panel`**; every other component is visited, counted in
+  `FrameStats::deferred`, and left undrawn. Text needs the package above; live-data components have
+  no geometry until a frame exists (#17).
+- **The golden sidecar has a static consumer, not a rendered one.** A test cross-checks it against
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,106 @@
...
+issue carries the reasoning and this file carries the outcome.
+
+## How to read the earlier entries
+
+This file starts at 0.6.0. **The entries for 0.2.0 through 0.5.0 are reconstructed from
+[`docs/roadmap.md`](docs/roadmap.md)'s own wave-by-wave record rather than derived from git history**,
+and that distinction is worth one paragraph because a reader may try to check them.
+
+Two of the four earlier tags — `v0.2.0` and `v0.5.0` — are not ancestors of `develop`, so
+`git log v0.4.0..v0.5.0` does not describe what a reader would expect it to. What the roadmap
```
ambroise-leclerc: **P2 — This is the wrong tag range to use as evidence that release history is non-linear.**

`v0.4.0` is an ancestor of `v0.5.0`: `git merge-base --is-ancestor v0.4.0 v0.5.0` succeeds, and `git rev-list --left-right --count v0.4.0...v0.5.0` reports `0 72`. Therefore `git log v0.4.0..v0.5.0` does describe the commits added between those two releases in the ordinary way. The fact that `v0.5.0` is not on `develop` does not invalidate the range between the tags themselves.

The genuinely non-linear boundary is `v0.2.0...v0.3.0`: `v0.2.0` is not an ancestor of `v0.3.0`, and the symmetric difference is `462 475`. Please use that boundary as the concrete example, or describe the mixed ancestry without claiming the valid `v0.4.0..v0.5.0` range is misleading. This matters because the paragraph is the changelog's provenance disclaimer; its supporting example must itself be reproducible.

docs/roadmap.md
```diff
@@ -91,12 +91,15 @@ until the content filled in would be waiting on #17, which is a wave of its own.
...
-- **`v0.5.0` is an ancestor of neither `develop` nor `master`**, and `develop` is 668 commits ahead of
-  `master`. Whatever the release ritual is, it is not readable from the repository - worth knowing
-  before 0.6.0 rather than at the moment of cutting it.
+- **There is now a [CHANGELOG](../CHANGELOG.md), and still no release workflow.** The 0.6.0 entry is
+  written from the merges it contains; the four earlier ones are summarised from this document's own
+  wave record, and the file says so, because two of the four tags are not ancestors of `develop` and
+  a reader checking them with `git log` would find that out the hard way.
+- **The release ritual is not readable from the repository.** `v0.2.0` and `v0.5.0` are ancestors of
+  neither `develop` nor `master`, while `v0.3.0` and `v0.4.0` are ancestors of `develop`; `develop`
+  is 668 commits ahead of `master`. So there is no rule a reader can infer about where a tag is cut,
```
ambroise-leclerc: **P2 — Both repository-history facts in this sentence are incorrect.**

First, `v0.5.0` is not an ancestor of neither branch: its peeled commit is the current `origin/master` tip, so it is necessarily an ancestor of `master`. Direct checks give: `v0.2.0 -> develop: no`, `v0.2.0 -> master: no`, `v0.5.0 -> develop: no`, and `v0.5.0 -> master: yes`.

Second, `develop` is not 668 commits ahead of `master`. `git rev-list --left-right --count origin/master...origin/develop` reports `1 32`: master has one unique commit and develop has 32. The 668 figure appears to have been a total-history count mislabeled as an ahead count.

Please correct the ancestry statement and use the actual symmetric divergence. Because this paragraph is presented as the evidence for why the release ritual cannot be inferred, factual errors here undermine the traceability rationale of the new changelog.

ambroise-leclerc (commented):
I found two release-provenance inaccuracies. The implementation and release bullets otherwise match reachable commits and current tests, and all PR checks are green. These are documentation-only findings, but they affect the auditability this changelog explicitly sets out to provide, so I recommend correcting them before merge. GitHub does not permit the PR author to submit a formal request-changes review, so these are published as review comments.

docs/roadmap.md
ambroise-leclerc: [P2] Both wrong, and I have checked your figures rather than taking them: `v0.5.0^{commit}` is `a92643bf6`, which is exactly `origin/master`, and `git rev-list --left-right --count origin/master...origin/develop` gives `1 32`.

The 668 came from the **local** `master` ref, which is 462 commits divergent from `origin/master` - a stale branch I never fetched against. So it was a real number about the wrong pair of refs, published twice, in the paragraph that exists to justify not trusting inference. I also repeated it to you outside the PR.

What the corrected facts change is not just the bullet but its conclusion. **Tags land on `master`** - that is inferable, and I said it was not. Fixed in b97d1c9, and the bullet now claims only what is actually missing: the *procedure*. Nothing records how develop reaches master, who moves the version in `CMakeLists.txt`, or what a tag must be accompanied by.

The one non-linear boundary is `v0.2.0` against `v0.3.0`, and it is not a mystery either - #23 purged normative text from git history, so everything before that rewrite sits on a line the current one does not contain. That is now stated as the cause rather than left as an anomaly.

CHANGELOG.md
ambroise-leclerc: [P2] Confirmed. `git merge-base --is-ancestor v0.4.0 v0.5.0` succeeds and the symmetric count is `0 72`, so that range is perfectly ordinary and my disclaimer's only supporting example did not support it. A provenance note whose evidence does not reproduce is worse than no note - it teaches a reader to distrust the parts that are true.

Rewritten in b97d1c9 around the boundary that is genuinely non-linear, `v0.2.0...v0.3.0` at `462 475`, with the cause named: #23 purged reproduced normative text from git history, so the pre-rewrite line is not in the current one.

The hedge narrows accordingly. `v0.3.0..v0.4.0` and `v0.4.0..v0.5.0` are linear (`0 93`, `0 72`), so the file now says commit-level detail *is* available for those two and that the entries stay at wave granularity for consistency with 0.6.0's - which is a choice, where before I was implying an obstacle.

CHANGELOG.md
```diff
@@ -0,0 +1,106 @@
...
+## 0.6.0 — unreleased
+
+Wave 5: the `.medui` compiler, end to end. An authored screen becomes a committed artifact, becomes
+`constexpr` C++, and reaches compared pixels.
+
+### The compiler (#15, all twelve children)
+
+- **The boundary and the artifacts** — ADR-011 fixes what a compiled screen is and is not; ADR-012
+  fixes the three files a screen emits and which are committed (#190, amended by #203 to keep the
+  compiled screen locale-free).
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and fixed in b97d1c9. `AGENTS.md` reported seven of twelve children with five remaining, and its skill table told agents that "a `.medui` file still builds nothing".

That second one is the one that mattered: `AGENTS.md` is where an agent is told what today looks like, so it was actively routing work away from tooling that exists. Both now describe the wave as closed, and name what it left behind - the text package (`#235`) and the components' geometry (`#17`) - rather than what it was still expected to deliver.

AM-L (approved):

PR #237, opened 2026-08-23
Release v0.6.0

ambroise-leclerc:
Wave 5, cut by [`docs/release-process.md`](docs/release-process.md) — written last week, executed
here for the first time.

## Read the diff in two parts

**This PR's diff is everything since v0.5.0**, because it merges `develop` into `master`. That is
32 commits: the whole of Wave 5. An earlier version of this description said "ten files", which
described the *release commit* and was misleading for anyone applying the release-process checks to
the PR in front of them.

- **The release commits** — `e5a480a`, `3b13f5e`, `bab7adc` — are what the procedure governs: the
  version, the six `report.json` files that record it, the changelog date, the roadmap, and the
  corrections review found in the procedure itself.
- **Everything else** is Wave 5, already reviewed and merged into `develop` PR by PR:
  [`CHANGELOG.md`](CHANGELOG.md) is the summary of it.

## The release-only diff, and the rule it satisfies

```console
$ git diff --stat develop...HEAD          # what this release adds over what was already reviewed
$ git diff generated/ | grep '^[-+]' | grep -v toolVersion | grep -v '^[-+][-+]'
```

The second prints nothing: **every changed artifact differs in `toolVersion` and nothing else**,
which is the review rule step 4 states.

## The deviation, now defined rather than recorded

Step 4 says run `mdux-bake-update`. I cannot build this tree — the GCC 16 floor means CI is the only
place it compiles — so the six fields were written directly.

Review was right that a documented ad-hoc deviation should not become the next release's precedent,
so the procedure now **defines the fallback**: bounded to a pure version bump, where `toolVersion` is
the only field that can move and the expected bytes are derivable by inspection, and gated on
`evidence.<kind>.<id>` passing on both toolchain legs — which re-bakes every artifact from its recipe
and byte-compares. It does not extend to a release that changes a recipe, an input or an option.

Worth stating precisely, because it changes what kind of deviation this is: ADR-007 does not forbid
this. It lists a hand-edited file under `generated/` as a **risk**, whose mitigation is that
"re-baking overwrites the edit while the byte-comparison fails the PR". That mitigation is exactly
what ran here, on two toolchains, and it passed.

## What executing the procedure corrected in it

Three things, all found by following it rather than by reading it:

- **Step 3 was wrong.** It said `CMakeLists.txt` is the only file to edit. Three tests hard-code the
  version — including `IEC 62304 Version Traceability` — and went red after every evidence comparison
  had passed. They should keep hard-coding it: that case exists to check the version a build *reports*
  is the version the project *declares*, and deriving it from the same macro would assert nothing.
- **Step 9 was in the wrong place.** The roadmap update belongs on the release branch, and the text
  said so while the numbering put it after the tag. It is step 6 now.
- **The roadmap contradicted itself**, in the tree this tag will name: the introduction still said
  Waves 1–3 had shipped and #15 was underway. That is the precise failure the reordering exists to
  prevent, and it happened in the same PR.

## After this merges

```console
$ git switch master && git pull
$ git tag -a v0.6.0 -m "MduX v0.6.0"
$ git push origin v0.6.0
$ git switch develop && git merge --no-ff master && git push
```

## Regulatory impact

The first release cut against a written procedure, with one deviation that is now defined rather than
improvised. §5.8's three requirements are answered mechanically: what is included is the composition
every `report.json` identifies by digest, the known anomalies are the changelog's limits section, and
the configuration it was built from is re-derived and byte-compared on two toolchains before the tag
exists.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,6 +197,7 @@ target_sources(MduXCore
             include/mdux/font/Schema.cppm
             include/mdux/text/Draw.cppm
             include/mdux/text/Schema.cppm
+            include/mdux/medui/Schema.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include(cmake/PreventInSourceBuilds.cmake)
 
-project(MduX VERSION 0.5.0 DESCRIPTION "Modern C++23 Modules UI Library for Medical Devices" HOMEPAGE_URL "https://github.com/ambroise-leclerc/MduX" LANGUAGES CXX)
+project(MduX VERSION 0.6.0 DESCRIPTION "Modern C++23 Modules UI Library for Medical Devices" HOMEPAGE_URL "https://github.com/ambroise-leclerc/MduX" LANGUAGES CXX)
 
 set(CMAKE_CXX_EXTENSIONS OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,6 +198,7 @@ target_sources(MduXCore
             include/mdux/text/Draw.cppm
             include/mdux/text/Schema.cppm
             include/mdux/medui/Schema.cppm
+            include/mdux/medui/Screen.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
@@ -209,6 +210,7 @@ target_sources(MduXCore
         src/font/Schema.cpp
         src/text/Draw.cpp
         src/text/Schema.cpp
+        src/medui/Screen.cpp
         src/governance/Governance.cpp
         src/governance/Justification.cpp
         src/governance/Program.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -465,6 +465,17 @@ mdux_compile_screen(
     RECIPE recipes/screen/endoscope-monitor.toml
 )
 
+# ...and the generated C++ for it (issue #201). Build-tree only, never committed. This is the last
+# link in the chain the epic set out to build: a `.medui` file becomes a committed artifact, the
+# artifact becomes `constexpr` C++, and a pixel test links that C++ and renders it - so the first
+# pixel drawn from an authored screen is compared against an expectation rather than described.
+include(cmake/MduXScreenEmit.cmake)
+mdux_emit_screen_package(
+    ID endoscope-monitor
+    OUT_MODULE MDUX_ENDOSCOPE_SCREEN_MODULE
+    OUT_DIR MDUX_GENERATED_SCREEN_DIR
+)
+
 # Generated C++ for that package (issue #121). Build-tree only, never committed - see
 # cmake/MduXShaderEmit.cmake. The renderer (#124) and the triangle example (#122) consume the
 # module form; the header form exists for a translation unit that cannot import a named module.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,6 @@ target_sources(MduXCore
             include/mdux/font/Schema.cppm
             include/mdux/text/Draw.cppm
             include/mdux/text/Schema.cppm
-            include/mdux/text/Raster.cppm
     PRIVATE
         src/evidence/Digest.cpp
         src/evidence/Json.cpp
@@ -209,7 +208,6 @@ target_sources(MduXCore
         src/font/Schema.cpp
         src/text/Draw.cpp
         src/text/Schema.cpp
-        src/text/Raster.cpp
         src/governance/Governance.cpp
         src/governance/Justification.cpp
         src/governance/Program.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -449,6 +449,20 @@ mdux_bake_artifact(
         shaders.spv
 )
 
+# The first compiled screen (issue #198). Registered through mdux_compile_screen(), which is a thin
+# front for mdux_bake_artifact() - a screen earns the same build-tree bake, the same -update target
+# and the same evidence.screen.<id> byte comparison as a font or a shader, because nothing about a
+# screen justifies weaker evidence than its neighbours.
+#
+# It draws no text on purpose: no text package is baked in this tree yet, so the text-budget stage
+# has nothing to measure against and a screen carrying t("STR-KEY") would be refused for declaring no
+# approved locale. #201 lands the text half.
+include(cmake/MduXCompileScreen.cmake)
+mdux_compile_screen(
+    ID endoscope-monitor
+    RECIPE recipes/screen/endoscope-monitor.toml
+)
+
 # Generated C++ for that package (issue #121). Build-tree only, never committed - see
 # cmake/MduXShaderEmit.cmake. The renderer (#124) and the triangle example (#122) consume the
 # module form; the header form exists for a translation unit that cannot import a named module.
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -132,10 +132,15 @@ Skipping this is how `master` accumulates commits `develop` does not have. As of
 `origin/master...origin/develop` is `1 32` — master carries one commit develop does not, which is
 exactly the drift this step prevents.
 
-### 9. Update the roadmap
+### 9. Update the roadmap — on the release branch, with the rest
 
-The wave line moves from "in progress" to "shipped vX.Y.Z", and the next wave opens. That file is the
-programme's own account of what shipped when, and the CHANGELOG's earlier entries rest on it.
+The wave line moves from "in progress" to "shipped vX.Y.Z", the counts move, and the next wave opens.
+That file is the programme's own account of what shipped when, and the CHANGELOG's earlier entries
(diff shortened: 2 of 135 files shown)


ambroise-leclerc (commented):
I found two release-blocking traceability contradictions and two process defects. The Linux and Windows jobs are green: all six evidence comparisons passed on GCC and MSVC, the exact version tests passed, and the full suites completed (594 Linux / 593 Windows). Those results validate the compiled artifacts but do not resolve the stale controlled documents or procedure ordering.

One P1 cannot be attached inline because the line is unchanged relative to `master`: `software_development_file/regulatory/IEC_62304/SAD.md` lines 11–12 still identify the current project version as `0.5.0`. The 0.6.0 tag would therefore contain contradictory document control. Following the SDF summarize-don't-duplicate rule, remove the duplicated literal and point solely to `CMakeLists.txt` rather than bumping another copy that will become stale again.

GitHub does not permit the PR author to submit a formal request-changes review, so these are published as review comments. I recommend resolving both P1 findings and the procedure defects before merging or tagging.

docs/roadmap.md
```diff
@@ -1,60 +1,115 @@
...
+> The divergence table below was last re-verified on 17 August 2026 and is not re-checked here.
 
 MduX (C++23 / Vulkan) and TrustSC (Rust) target the same problem — a medical-device UI
 SDK with IEC 62304 Class B/C compliance modelling built in. This is the dependency-ordered
 backlog that closes the gap. Waves 1, 2 and 3 have shipped — the renderer draws its first
 pixel, zero-SOUP ML inference is in the tree, and the documentation has been rebuilt from
 what the build actually produces. Track C's authoring story is what remains: #14 closed
-Wave 4 with the font and text pipeline, and #15 opens Wave 5 with the compiler that
-generates the screens it draws.
+Wave 4 with the font and text pipeline, and #15 is underway in Wave 5 with the compiler
```
ambroise-leclerc: **P1 — The roadmap still describes Wave 5 as underway while this same release marks it shipped.**

This introductory status says that only Waves 1–3 have shipped and that `#15 is underway in Wave 5`. Lower in the same file, the release changes the metrics to five shipped waves, labels Wave 5 `shipped v0.6.0`, and marks #15 done. The provenance header and footer compound the contradiction by claiming verification at `develop @ 06b2a59`, which is the #200 commit before #201 landed, while also claiming #15 is current through #201 and Wave 5 shipped.

This is exactly the stale-tag problem that the new release-process rationale says moving the roadmap into the release branch prevents. Please update the introductory status and the verification SHA/date to the actual release baseline so every section of the tagged roadmap describes the same release state.

docs/release-process.md
```diff
@@ -0,0 +1,180 @@
...
+
+```console
+$ git switch develop && git merge --no-ff master && git push
+```
+
+Skipping this is how `master` accumulates commits `develop` does not have. As of this writing
+`origin/master...origin/develop` is `1 32` — master carries one commit develop does not, which is
+exactly the drift this step prevents.
+
+### 9. Update the roadmap — on the release branch, with the rest
```
ambroise-leclerc: **P2 — This release-branch step is still ordered after merge, tag, and back-merge.**

The text correctly says the roadmap must be updated on the release branch so the tag includes it, but this remains step 9 after step 7 (`Merge to master and tag there`) and step 8 (`Back-merge`). A maintainer following the procedure sequentially would create the tag before performing this update—the precise failure the explanation says to avoid. Returning to the release branch afterward would require rewriting the tag and repeating the merges.

Move the roadmap update before the merge/tag step and renumber the following steps. The current PR happened to perform the update at the right time; the documented procedure would not reproduce that result.

generated/font/dejavu-ui/report.json
```diff
@@ -39,5 +39,5 @@
   },
   "schemaVersion": 1,
   "tool": "mdux-textbake",
-  "toolVersion": "0.5.0"
+  "toolVersion": "0.6.0"
```
ambroise-leclerc, line 42: **P2 — This and the other five generated reports were hand-edited instead of produced through the only authorized update path.**

The evidence-pipeline doctrine says files under `generated/` are never edited manually and that `mdux-bake-update` is the sole source-tree update path. Step 4 of the new release procedure says the same. The PR description explicitly records a deviation: these six `toolVersion` fields were changed by hand because no supported local toolchain was available.

Both CI legs re-baked the artifacts and byte-compared successfully, so the committed bytes are correct. However, that verification does not make the statement that this is the first release cut against the written procedure accurate. Please either obtain the generated update from a supported build and commit those outputs, or amend the procedure to define a controlled CI-verified fallback and explain how the resulting files are transferred without hand-editing. A documented ad-hoc deviation should not silently become the precedent for the next release.

docs/roadmap.md
ambroise-leclerc: [P1] Confirmed, and the timing is the sharp part: this is the failure the step-9 rationale in this same PR claims to prevent, happening in this same PR. The introduction still said Waves 1 to 3 had shipped and `#15` was underway, while the metrics twenty lines below said five waves and done, and the banner named `06b2a59` - the #200 commit, before #201 landed.

A tagged roadmap that contradicts itself is worse than a stale one: a reader consulting the one document that says what shipped gets two answers and no way to choose.

Fixed in bab7adc. The introduction, the metrics, the wave block, the banner and the footer all name the v0.6.0 baseline, `develop` @ `acbe102`.

docs/release-process.md
ambroise-leclerc: [P2] Confirmed. I added the reasoning and left the numbering, so the document explained why the roadmap belongs on the release branch and then instructed a sequential reader to update it after tagging - and re-tagging is exactly what that would cost.

It is step 6 now, before the merge, and the following steps renumber. The explanation moved with it and names the failure mode rather than the rule alone.

This PR happened to do it in the right order because I was working from the reasoning rather than the numbers, which is precisely why it needed catching: the next person will work from the numbers.

generated/font/dejavu-ui/report.json
ambroise-leclerc: [P2] The substance is right and I have acted on it: an ad-hoc deviation must not become the next release's precedent by being recorded once in a PR body. The procedure now **defines** the fallback rather than describing what I happened to do.

One correction to the premise, though, and I would rather state it than let it stand: **ADR-007 does not say `generated/` is never edited manually.** It lists "Someone hand-edits a file under `generated/`" among its risks, with the mitigation "the strict reader rejects permissive JSON, and re-baking overwrites the edit while the byte-comparison fails the PR". `MduXBake.cmake`'s "only path that writes into the source tree" is a statement about what a *build* may do, not about what a human may commit. So the doctrine anticipated this case and specified how it is caught, which is what ran here - on two toolchains, and it passed.

The fallback as defined is bounded to where the expected bytes are derivable by inspection: a pure version bump, where `toolVersion` is the only field that can move. It is gated on `evidence.<kind>.<id>` on **both** legs, which re-bakes from the recipes and byte-compares - stronger verification than one unverified local bake, not weaker. And it explicitly does not extend to a release that changes a recipe, an input or a resolved option: there the bytes are not derivable and a supported toolchain is required.

The release PR is required to say which of the two paths produced the bytes, so a reviewer never has to infer it.

docs/release-process.md
```diff
@@ -0,0 +1,166 @@
...
+precisely what is being released — every included item, its known anomalies, and the configuration it
+was built from — and §8.3 asks that changes to that composition be controlled. This repository can
+answer both mechanically, because every baked artifact carries a `report.json` recording its inputs
+by digest and CI re-derives it on two toolchains. What follows is the procedure that turns those
+mechanisms into a release.
+
+## The branch topology
+
+```text
+  feature/NNN-slug ──▶ develop ──▶ release/vX.Y.Z ──▶ master  (tagged vX.Y.Z)
```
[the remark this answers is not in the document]
ambroise-leclerc: Correct, and the failure is the silent kind: a PR from `feature/...` matches no workflow's `branches:` filter, so it gets **no checks at all** rather than failing ones. Fixed in bab7adc - the diagram uses `<issue-number>-slug`, which is what GitHub's "create a branch for this issue" button generates and what the filters match, with `feat/**` noted as the predecessor convention.

CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include(cmake/PreventInSourceBuilds.cmake)
 
-project(MduX VERSION 0.5.0 DESCRIPTION "Modern C++23 Modules UI Library for Medical Devices" HOMEPAGE_URL "https://github.com/ambroise-leclerc/MduX" LANGUAGES CXX)
+project(MduX VERSION 0.6.0 DESCRIPTION "Modern C++23 Modules UI Library for Medical Devices" HOMEPAGE_URL "https://github.com/ambroise-leclerc/MduX" LANGUAGES CXX)
```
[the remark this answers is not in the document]
ambroise-leclerc: Fair, and the description was misleading rather than merely incomplete: it described the release *commit* while the PR merges `develop` into `master`, so the diff in front of a reviewer is all 32 commits of Wave 5.

Rewritten to say so, and to split what the release-process checks apply to - three release commits - from what was already reviewed PR by PR on `develop`, with the commands for looking at each. Splitting the bump into its own PR is the other option you name; it would not help here, because the thing being merged to `master` is the wave, and the bump has to travel with it.

AM-L (approved):

PR #238, opened 2026-08-23
Record what cutting v0.6.0 taught the procedure

ambroise-leclerc:
v0.6.0 is tagged and published. Three things the procedure got wrong, each found by following it.

**Tag `origin/master`, never a checked-out `master`.** The documented `git switch master && git pull`
would have attempted to merge 462 divergent commits in this clone - the same stale local branch that
produced a wrong divergence figure in the roadmap last week. Tagging the remote ref needs no checkout
and cannot pick up a stale branch.

**Publish a GitHub release, not only a tag.** Every release before this one has one, titled
`vX.Y.Z — Wave N: <subject>`. The tag alone is not what a reader lands on, and the notes are where
IEC 62304 §5.8 known limits actually get read.

**Merge the release PR rather than squashing it.** #237 was squashed, which gives the release commit
one parent - so the back-merge saw every file the release branch touched as an `add/add` conflict.
Four of them. They resolve to `master`'s side, because that side *is* the release content, but the
procedure now says to avoid creating them and how to recover if the button was pressed anyway,
including the per-file check that a resolution kept the right side.

The release itself is complete: tag `v0.6.0` on `efdd55a`, GitHub release published, back-merge done
and `origin/master...origin/develop` back to `0 34`.

## Summary by CodeRabbit

* **Documentation**
  * Updated release process guidance to use the `master` branch for release pull requests and annotated tagging.
  * Added instructions for publishing GitHub releases with changelog highlights and known limitations.
  * Clarified back-merge procedures, including branch alignment checks, merge requirements, and conflict-resolution verification.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
 The `evidence` label is the one that matters here: it is the mechanical answer to "is the committed
 artifact the one this source produces".
 
-### 8. Merge to `master` and tag there
+### 8. Merge to `master`, tag there, and publish the release
 
 ```console
-$ gh pr create --base master --head release/v0.6.0 --title "Release v0.6.0"
-# after review and a green run:
-$ git switch master && git pull
-$ git tag -a v0.6.0 -m "MduX v0.6.0"
-$ git push origin v0.6.0
+$ gh pr create --base master --head release/vX.Y.Z --title "Release vX.Y.Z"
+# after review and a green run, merge it - then, without checking master out:
+$ git fetch origin
+$ git tag -a vX.Y.Z origin/master -m "MduX vX.Y.Z"
+$ git push origin vX.Y.Z
+$ gh release create vX.Y.Z --verify-tag --title "vX.Y.Z — Wave N: <what it is>" --notes-file <notes>
 ```
 
-The tag is annotated, and it names the commit on `master`. A lightweight tag records no author and no
-date of its own, which is a poor fit for a document that is meant to identify a configuration.
+Three things this spells out because cutting v0.6.0 found each of them the hard way.
+
+**Tag `origin/master`, not a checked-out `master`.** A local `master` in a long-lived clone can be
+hundreds of commits from the real one — 462, in the clone this procedure was written in — so
+`git switch master && git pull` attempts a merge rather than a fast-forward. Tagging the remote ref
+needs no checkout and cannot pick up a stale branch.
+
+**The tag is annotated.** A lightweight tag records no author and no date of its own, which is a poor
+fit for an object meant to identify a configuration.
+
+**Publish a GitHub release too.** Every release before v0.6.0 has one, titled
+`vX.Y.Z — Wave N: <subject>`, and the tag alone is not what a reader lands on. Its notes are the
+changelog entry's highlights *and its known-limits section* — §5.8's anomalies belong where someone
+reads them, not only in a file they might open.
 
 ### 9. Back-merge, so the histories do not drift
 
 ```console
-$ git switch develop && git merge --no-ff master && git push
+$ git switch develop && git pull && git merge --no-ff origin/master && git push
+$ git rev-list --left-right --count origin/master...origin/develop   # left column must be 0
 ```
 
-Skipping this is how `master` accumulates commits `develop` does not have. As of this writing
-`origin/master...origin/develop` is `1 32` — master carries one commit develop does not, which is
-exactly the drift this step prevents.
+Skipping this is how `master` accumulates commits `develop` does not have.
+
+**Merge the release PR, do not squash it.** A squash gives the release commit a single parent, so the
+release-branch commits are not in `master`'s ancestry and every file the release branch touched
+arrives at the back-merge as an `add/add` conflict — four of them in v0.6.0. The conflicts are
+resolvable, because `master`'s side *is* the release content, but they are noise the merge button
+creates and a real merge avoids.
+
+If it was squashed anyway: resolve every conflict to `master`'s side, then verify rather than trust —
+`git diff --quiet origin/master -- <path>` for each one, so a resolution that silently kept the wrong
+side is caught before it is pushed.
 
 ## What is deliberately not automated
 
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -59,7 +59,7 @@ moves.
 ### 1. Confirm `develop` is releasable
 
 ```console
-$ git switch develop && git pull
+$ git switch develop && git pull --ff-only    # fails rather than merging, if local has drifted
 $ gh pr list --state open --base develop      # nothing half-landed
 $ gh run list --branch develop --limit 1      # green on both toolchain legs
 ```
@@ -164,14 +164,21 @@ artifact the one this source produces".
 
 ```console
 $ gh pr create --base master --head release/vX.Y.Z --title "Release vX.Y.Z"
-# after review and a green run, merge it - then, without checking master out:
+# after review and a green run - --merge, never --squash, see step 9:
+$ gh pr merge <pr> --merge
+# then, without checking master out:
 $ git fetch origin
-$ git tag -a vX.Y.Z origin/master -m "MduX vX.Y.Z"
+$ git tag -a vX.Y.Z -m "MduX vX.Y.Z" origin/master
 $ git push origin vX.Y.Z
 $ gh release create vX.Y.Z --verify-tag --title "vX.Y.Z — Wave N: <what it is>" --notes-file <notes>
 ```
 
-Three things this spells out because cutting v0.6.0 found each of them the hard way.
+Four things this spells out because cutting v0.6.0 found each of them the hard way.
+
+**Merge with `--merge`, mechanically.** The reason is in step 9, but the choice is made *here*, and
+the web UI remembers whichever method was used last — which is precisely how v0.6.0 was squashed
+against the intended topology. Naming the method on the command line does not depend on anyone
+recalling a warning further down this page.
 
 **Tag `origin/master`, not a checked-out `master`.** A local `master` in a long-lived clone can be
 hundreds of commits from the real one — 462, in the clone this procedure was written in — so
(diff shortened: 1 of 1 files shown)


ambroise-leclerc (commented):
I verified the factual release state: `v0.6.0` is an annotated tag on `origin/master` at `efdd55a`, the GitHub release is published with its known limits, `origin/master...origin/develop` is `0 34`, and PR #237 was squash-merged. The graph supports the documented explanation of the conflicts. I found one potentially destructive recovery rule and three reproducibility/accuracy issues in the revised procedure. All PR checks are green and `git diff --check` passes, but I recommend correcting these process findings before this becomes the precedent for v0.7.0. GitHub does not permit the PR author to submit a formal request-changes review, so these are published as review comments.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
-exactly the drift this step prevents.
+Skipping this is how `master` accumulates commits `develop` does not have.
+
+**Merge the release PR, do not squash it.** A squash gives the release commit a single parent, so the
+release-branch commits are not in `master`'s ancestry and every file the release branch touched
+arrives at the back-merge as an `add/add` conflict — four of them in v0.6.0. The conflicts are
+resolvable, because `master`'s side *is* the release content, but they are noise the merge button
+creates and a real merge avoids.
+
+If it was squashed anyway: resolve every conflict to `master`'s side, then verify rather than trust —
```
ambroise-leclerc: **P1 — Resolving every conflict to `master` can silently delete work already merged to `develop`.**

That resolution happened to be correct for the four conflicts in the v0.6.0 back-merge, but the procedure does not freeze `develop` while the release PR is open. Suppose another PR lands on `develop` during release preparation and edits `CHANGELOG.md`, `docs/roadmap.md`, `CMakeLists.txt`, or another shared file also present in the squashed release delta. The back-merge can then conflict for two reasons at once. Taking `master` unconditionally discards the post-branch `develop` change, and the proposed `git diff --quiet origin/master -- <path>` check actively requires that loss.

This also contradicts the repository-wide union-by-default policy for shared registries. Please scope the recovery to paths proven to differ only because of the squash—for example by comparing each side with the release branch point/release head—and retain or union independent `develop` changes. If that proof is not available, stop for an explicit conflict decision rather than prescribing one side globally.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
+### 8. Merge to `master`, tag there, and publish the release
 
 ```console
-$ gh pr create --base master --head release/v0.6.0 --title "Release v0.6.0"
-# after review and a green run:
-$ git switch master && git pull
-$ git tag -a v0.6.0 -m "MduX v0.6.0"
-$ git push origin v0.6.0
+$ gh pr create --base master --head release/vX.Y.Z --title "Release vX.Y.Z"
+# after review and a green run, merge it - then, without checking master out:
```
ambroise-leclerc: **P2 — The procedure still leaves the merge method unspecified at the action point.**

`gh pr create` creates the release PR, but the next instruction is only ‘merge it.’ GitHub can retain the last-selected/default merge method, which is exactly how #237 was squashed despite the desired topology. The warning at line 198 explains the consequence after the fact but does not make the action reproducible.

Please prescribe an explicit merge-commit operation, such as `gh pr merge --merge <pr-number-or-url>` after the checks are green, or an equally unambiguous UI requirement. A release procedure written to prevent the squash failure should select the safe method mechanically rather than depend on remembering prose further down the page.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
+**Publish a GitHub release too.** Every release before v0.6.0 has one, titled
+`vX.Y.Z — Wave N: <subject>`, and the tag alone is not what a reader lands on. Its notes are the
+changelog entry's highlights *and its known-limits section* — §5.8's anomalies belong where someone
+reads them, not only in a file they might open.
 
 ### 9. Back-merge, so the histories do not drift
 
 ```console
-$ git switch develop && git merge --no-ff master && git push
+$ git switch develop && git pull && git merge --no-ff origin/master && git push
```
ambroise-leclerc: **P2 — This unrestricted `git pull` can reproduce the local-branch contamination the preceding section avoids.**

The tag step correctly avoids a stale local `master`, but the back-merge now trusts local `develop`. If that branch has diverged, tracks the wrong upstream, or contains an unpublished commit, plain `git pull` may create an additional merge before `origin/master` is merged. That contaminates the controlled back-merge and makes the final `0 ...` ancestry count insufficient to explain where the extra history came from.

Use `git pull --ff-only` at minimum, or fetch and explicitly fast-forward a clean local branch to `origin/develop`, then merge `origin/master`. The same guard should be considered for step 1. The procedure should fail on unexpected local divergence instead of resolving it implicitly.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
+
+**Tag `origin/master`, not a checked-out `master`.** A local `master` in a long-lived clone can be
+hundreds of commits from the real one — 462, in the clone this procedure was written in — so
+`git switch master && git pull` attempts a merge rather than a fast-forward. Tagging the remote ref
+needs no checkout and cannot pick up a stale branch.
+
+**The tag is annotated.** A lightweight tag records no author and no date of its own, which is a poor
+fit for an object meant to identify a configuration.
+
+**Publish a GitHub release too.** Every release before v0.6.0 has one, titled
```
ambroise-leclerc: **P2 — ‘Every release before v0.6.0’ is broader than the repository evidence supports.**

GitHub has published releases for v0.2.0, v0.3.0, v0.4.0, and v0.5.0. The repository also has a `v0.1` tag, but GitHub has no corresponding release. Therefore the literal claim that every earlier release has one is false unless this document defines v0.1 as something other than a release.

Please narrow the statement to ‘every wave release from v0.2.0 onward’ or explicitly explain why the v0.1 tag is excluded. That preserves the useful naming convention without creating a release-history claim a reader cannot reproduce.

docs/release-process.md
ambroise-leclerc: Agreed, and this was the finding worth stopping for. Fixed in 3a878ed.

You are right that the rule was only sound because of a fact this procedure does not establish. I checked what actually made it work: `git diff --stat acbe102 efdd55a` is 12 files, all of them release-only, because nothing landed on `develop` between the branch point and the merge. Nothing in the document arranges that, so the rule as written would have required deleting any `develop` change that collided.

The recovery is now per path, and it proves the condition instead of assuming it. A squashed PR's branch is usually deleted, but GitHub keeps `refs/pull/<pr>/head`, so the branch point is always recoverable:

```console
$ git fetch origin refs/pull/<pr>/head
$ base=$(git merge-base FETCH_HEAD origin/develop)
$ git log --oneline $base..origin/develop -- <path>
```

Empty means the collision is the squash alone, and only then is `master`'s side the whole content — with the check now `git diff --quiet --cached origin/master -- <path>`, staged rather than working-tree, as you asked. Non-empty means two real changes met and it is an ordinary merge: keep both, union for the shared registries, and stop for an explicit decision if they cannot be reconciled by reading them.

Run against the four v0.6.0 conflicts, all four return empty — so the resolution I made is the one the rule now derives rather than the one it prescribes.

docs/release-process.md
ambroise-leclerc: Fixed in 3a878ed. `gh pr merge <pr> --merge` is now in the step 8 block at the point the choice is made, with `--merge, never --squash, see step 9` beside it, and a paragraph naming the mechanism: the web UI remembers the last method used, which is how #237 was squashed against the intended topology. Prose in a later section cannot prevent a button press in an earlier one.

docs/release-process.md
ambroise-leclerc: Fixed in 3a878ed — `--ff-only` on both the step 9 pull and the step 1 pull, which you were right to flag as the same exposure.

The reason is now written down too, because it is specific to this step: an extra merge created by an implicit pull is indistinguishable, in the `rev-list --left-right --count` that closes the step, from the drift the count exists to measure. The check would still read `0` and mean less than it appears to.

docs/release-process.md
ambroise-leclerc: Correct, and I overstated it. `gh release list` shows four: v0.2.0 through v0.5.0. The `v0.1` tag has none.

Narrowed in 3a878ed to "every wave release from v0.2.0 onward has one; the `v0.1` tag predates the convention and has none" — which keeps the naming convention as the reproducible part and drops the history claim a reader would have found false on the first check.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
 ```console
-$ gh pr create --base master --head release/v0.6.0 --title "Release v0.6.0"
-# after review and a green run:
-$ git switch master && git pull
-$ git tag -a v0.6.0 -m "MduX v0.6.0"
-$ git push origin v0.6.0
+$ gh pr create --base master --head release/vX.Y.Z --title "Release vX.Y.Z"
+# after review and a green run, merge it - then, without checking master out:
+$ git fetch origin
+$ git tag -a vX.Y.Z origin/master -m "MduX vX.Y.Z"
```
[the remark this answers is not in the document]
ambroise-leclerc: Reordered in 3a878ed, but the reason given here is not right, so recording the check rather than the conclusion.

`git tag` uses git's parse-options, which interleaves options and positional arguments; `-m` after the ref parses and applies normally:

```console
$ git tag -a ztest-parse HEAD -m "probe"
$ git tag -l -n1 ztest-parse
ztest-parse     probe
```

So the documented form was executable as written. `-a vX.Y.Z -m "MduX vX.Y.Z" origin/master` is the conventional order and reads better against the rest of the document, which is why it changed — not because the previous form could fail.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
+$ git push origin vX.Y.Z
+$ gh release create vX.Y.Z --verify-tag --title "vX.Y.Z — Wave N: <what it is>" --notes-file <notes>
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in 3a878ed, together with @ambroise-leclerc's [P2](https://github.com/ambroise-leclerc/MduX/pull/238#discussion_r3837968813) on the same point. `gh pr merge <pr> --merge` is now in the step 8 block where the choice is actually made, annotated `never --squash, see step 9`, with the mechanism named: the web UI remembers the last method used, which is how #237 was squashed.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
+`vX.Y.Z — Wave N: <subject>`, and the tag alone is not what a reader lands on. Its notes are the
+changelog entry's highlights *and its known-limits section* — §5.8's anomalies belong where someone
+reads them, not only in a file they might open.
 
 ### 9. Back-merge, so the histories do not drift
 
 ```console
-$ git switch develop && git merge --no-ff master && git push
+$ git switch develop && git pull && git merge --no-ff origin/master && git push
+$ git rev-list --left-right --count origin/master...origin/develop   # left column must be 0
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in 3a878ed, by showing the output rather than dropping the prompts — the `$` convention is used throughout this document, and stripping it from one block would have made that block the odd one.

The `rev-list` line now prints its expected result:

```console
$ git rev-list --left-right --count origin/master...origin/develop
0	34
```

Which resolves MD014 for the block and is the more useful form anyway: the reader now sees what a passing check looks like instead of being told about it in the following sentence.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
-Skipping this is how `master` accumulates commits `develop` does not have. As of this writing
-`origin/master...origin/develop` is `1 32` — master carries one commit develop does not, which is
-exactly the drift this step prevents.
+Skipping this is how `master` accumulates commits `develop` does not have.
+
+**Merge the release PR, do not squash it.** A squash gives the release commit a single parent, so the
+release-branch commits are not in `master`'s ancestry and every file the release branch touched
+arrives at the back-merge as an `add/add` conflict — four of them in v0.6.0. The conflicts are
+resolvable, because `master`'s side *is* the release content, but they are noise the merge button
+creates and a real merge avoids.
```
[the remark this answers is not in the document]
ambroise-leclerc: Correct on both counts, and the second one I had not noticed. Fixed in 3a878ed.

`git merge-tree` on the real back-merge parents:

```
CONFLICT (add/add):  CHANGELOG.md
CONFLICT (add/add):  docs/release-process.md
CONFLICT (content):  docs/roadmap.md
CONFLICT (add/add):  generated/screen/endoscope-monitor/report.json
Auto-merging CMakeLists.txt
```

Three `add/add`, one content — and `CMakeLists.txt` merged cleanly, so "every file the release branch touched" was wrong as well as the conflict type. The text now states both, and explains the mechanism that actually produces them: a squash leaves the back-merge with the *previous release* as its common commit, so everything both sides changed since then meets.

docs/release-process.md
```diff
@@ -160,28 +160,50 @@ $ ctest --preset <your-preset>
...
+
+If it was squashed anyway: resolve every conflict to `master`'s side, then verify rather than trust —
+`git diff --quiet origin/master -- <path>` for each one, so a resolution that silently kept the wrong
+side is caught before it is pushed.
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in 3a878ed — see the [detailed reply](https://github.com/ambroise-leclerc/MduX/pull/238#discussion_r3837968812) on the maintainer's thread for the same finding. Short version: the recovery is now per path and proves the condition rather than assuming it, the comparison is `--cached` against `origin/master`, and a path `develop` also changed is treated as the ordinary merge it is — union for the shared registries, explicit decision when the two sides cannot be reconciled by reading them.

AM-L (approved):

PR #239, opened 2026-08-23
Position a text run from a string [#235]

ambroise-leclerc:
First of two for #235. This is the tool half; the committed artifact and the screen that carries a `Label` follow on top of it.

## What was missing

`mdux-textbake` has produced text packages since #157, but never a *run*. The text path assembled a package with an empty run list and a comment pointing elsewhere for the runs — and the two docs disagreed about where: `TextBake.cppm` said the `.medui` compiler produced them, `TextBudget.cppm` said "the text package already holds the runs `mdux-textbake` positioned". Neither was true. Nothing could position a run until a font package existed to position it against, and one has since #160.

## What this adds

```toml
[package]
id     = "label-welcome"
atlas  = "roboto-ui"
locale = "en-US"
font   = "generated/font/roboto-ui/package.json"

[strings]
keys   = ["STR-TITLE", "STR-SUBTITLE"]
values = ["Welcome",   "Select a patient"]
```

A key becomes a run id, which is what `t("STR-KEY")` resolves against. A value becomes v1 records: look each code point up in the font package, emit a record at the pen, advance by the advance width plus whatever kerning the package baked for the pair.

**It is a pen walk, not shaping** — no reordering, substitution, ligatures or bidi. A script needing any of those is not writable by this baker, and the honest form of that limit is producing no package rather than a plausible wrong one. ADR-010 leaves the runtime no fallback, so a code point the font cannot draw is `TXT024` rather than a substitution.

**The pen accumulates in font units** and converts per glyph, rather than converting each advance and accumulating pixels — only the first keeps a rounding error from compounding along a line. Integer throughout: the sidecar is committed bytes compared across toolchains, and a float would make the artifact depend on the host.

**A record per code point, blanks included.** Neither consumer draws them, so the six bytes buy nothing at run time; they buy something at review time, since a run' record count is its character count.

## Evidence

The font package is the bake' one input, digested from the bytes that were *parsed* rather than from a second read — so re-baking the font invalidates the text package rather than leaving it describing an atlas that moved, and a file that changed mid-bake cannot produce a report attesting to a bake that never happened.

## Tests

A three-glyph fixture font written through `mdux::font::FontPackage::write()` (so a fixture that stops being a legal package fails here rather than teaching the baker to accept one), with `unitsPerEm` 1000 against `pixelSize` 10 — one pixel is exactly 100 font units, so the expected x of every record is readable by hand. The expected bytes are derived from the fixture' advances rather than from the implementation' formula: a test that recomputed the formula would agree with a wrong one.

Ten rejection cases, each pinned to its stable code, covering the six new diagnostics plus the two path escapes.

## Not verified locally

Still no local build: Homebrew now ships GCC 16, and the configure now succeeds, but its `bits/std.cc` is a **1-byte stub**, so `import std` has nothing to scan. CI is the compiler.

## Summary by CodeRabbit

* **New Features**
  * Added localized string entries to text recipes.
  * Added font-package loading and validation for glyph availability, locale, path safety, and package integrity.
  * Generated deterministic positioned glyph data, including advances and kerning, for baked text.
  * Added font-package details to bake reports and package metadata.

* **Bug Fixes**
  * Added diagnostics for unsupported scripts, combining marks, negative pen positions, malformed recipes, invalid packages, atlas mismatches, and excessive run widths.
  * Strengthened repository path protection, including symlink and cross-platform path escapes.
  * Preserved empty-text and legacy recipe support.

### DIFF
--- a/tools/text/TextBake.cpp
+++ b/tools/text/TextBake.cpp
@@ -66,6 +66,16 @@ constexpr std::string_view recipePixelSizeInvalid = "TXT018";
 // the fix is a different font, not a different charset.
 constexpr std::string_view tabularFiguresMismatch = "TXT019";
 
+// #235's text path. A font recipe's own "cannot read the font" is TXT012, which is about a
+// TrueType file; these are about a baked font *package*, which fails in different ways and sends
+// an author somewhere else - so they get their own codes rather than overloading TXT012/TXT013.
+constexpr std::string_view recipeStringsMalformed  = "TXT020";
+constexpr std::string_view recipeFontPackageUnread = "TXT021";
+constexpr std::string_view fontPackageUnparsed     = "TXT022";
+constexpr std::string_view localeNotApproved       = "TXT023";
+constexpr std::string_view stringNotDrawable       = "TXT024";
+constexpr std::string_view runTooWide              = "TXT025";
+
 void report(std::vector<cli::Diagnostic>& diagnostics, std::string file, std::size_t line,
              std::string_view code, std::string message, std::string fixHint = {}) {
     diagnostics.push_back(cli::Diagnostic{.file = std::move(file),
@@ -91,6 +101,36 @@ void report(std::vector<cli::Diagnostic>& diagnostics, std::string file, std::si
     return std::as_bytes(std::span{text.data(), text.size()});
 }
 
+/**
+ * @brief Resolves `relativePath` under `root`, or nullopt when it would escape.
+ *
+ * `root / relativePath` is not the containment it looks like: std::filesystem's operator/
+ * *replaces* the left operand when the right is absolute, so `source = "/etc/shadow"` reads that
+ * file and ignores `root` entirely. A relative path with `..` components escapes the same way by a
+ * different route. The recipe is repository content rather than network input, so this is not a
+ * live attack surface today - but a baker that will one day run over a recipe somebody else wrote
+ * should not have "the recipe is trusted" as its only defence.
+ *
+ * Shared by the font path's `source` and the text path's `font`, so the two cannot drift into
+ * disagreeing about what is inside the repository.
+ */
+[[nodiscard]] std::optional<std::filesystem::path> confine(const std::filesystem::path& root,
+                                                            std::string_view relativePath) {
+    if (relativePath.empty() || std::filesystem::path{relativePath}.is_absolute()) {
+        return std::nullopt;
+    }
+    const std::filesystem::path resolved   = (root / relativePath).lexically_normal();
+    const std::filesystem::path normalRoot = root.lexically_normal();
+    // lexically_relative() gives the path from root to the target; a result that starts with ".."
+    // means the target is outside root. Comparing the normalised forms is what makes `a/../../b`
+    // fail rather than pass on a string prefix check.
+    const auto relative = resolved.lexically_relative(normalRoot);
+    if (relative.empty() || *relative.begin() == "..") {
+        return std::nullopt;
+    }
+    return resolved;
+}
+
 }  // namespace
 
 // ---------------------------------------------------------------------------
@@ -103,6 +143,15 @@ json::Value Recipe::toOptions() const {
     static_cast<void>(options.set("atlas", json::Value::string(atlas)));
     static_cast<void>(options.set("locale", json::Value::string(locale)));
     static_cast<void>(options.set("sidecar", json::Value::string(sidecar)));
+    if (!font.has_value()) {
+        // The resolved font package path, and how many runs were asked for. Not the strings
+        // themselves: they are the recipe's own content, the recipe's digest is recorded a few
+        // lines above this in every report, and their keys are already `package.json`'s run ids.
+        // Copying them here would make one edited translation appear in three places of the same
+        // artifact diff, which makes the diff harder to read rather than more complete.
+        static_cast<void>(options.set("font", json::Value::string(fontPackage)));
+        static_cast<void>(options.set("strings", json::Value::integer(strings.size())));
+    }
     if (font.has_value()) {
         // Every resolved knob, because ADR-007's rule is that a silently changed default must not
         // leave every report looking unchanged. The charset is recorded as its ranges rather than
@@ -314,6 +363,86 @@ namespace {
     return spec;
 }
 
+/**
+ * @brief Reads a text recipe's `[strings]` table and its `[package] font` key into `recipe`.
+ *
+ * Both are optional together and required together. A recipe with neither is the S1 shape - a
+ * package with no runs - which stays valid because a locale can legitimately have nothing to say
+ * yet, and because every text package in this repository was that shape until #235. A recipe with
+ * one and not the other is a half-written recipe, and saying so here is more useful than baking an
+ * empty package and leaving the author to wonder where their strings went.
+ */
+[[nodiscard]] bool parseStrings(const toml::Document& document, const toml::Table& package, Recipe& recipe,
+                                std::string_view recipePath, std::vector<cli::Diagnostic>& diagnostics) {
+    const toml::Table* strings = document.table("strings");
+    const toml::Value* font    = package.find("font");
+
+    if (strings == nullptr && font == nullptr) {
+        return true;
+    }
+    if (strings == nullptr || font == nullptr) {
+        report(diagnostics, std::string{recipePath}, 0, recipeStringsMalformed,
+               strings == nullptr ? "recipe names a font package and has no [strings] table"
+                                  : "recipe has a [strings] table and names no font package",
+               "A run is a string positioned against a font package: neither is meaningful alone.");
+        return false;
+    }
+
+    std::vector<std::string> keys;
+    std::vector<std::string> values;
+    try {
+        recipe.fontPackage = font->asString();
+        keys               = strings->require("keys").asStringArray();
+        values             = strings->require("values").asStringArray();
+    } catch (const toml::TomlError& error) {
+        report(diagnostics, std::string{recipePath}, error.line(), recipeStringsMalformed, error.what(),
+               "[strings] needs parallel 'keys' and 'values' string arrays.");
+        return false;
+    }
+
+    if (keys.size() != values.size()) {
+        report(diagnostics, std::string{recipePath}, 0, recipeStringsMalformed,
+               std::format("[strings] has {} keys and {} values", keys.size(), values.size()),
+               "The two arrays are parallel: one value per key, in the same order.");
+        return false;
+    }
+    if (keys.empty()) {
+        report(diagnostics, std::string{recipePath}, 0, recipeStringsMalformed,
+               "[strings] is present and empty",
+               "Remove the table to bake a package with no runs, or give it a key.");
+        return false;
+    }
+
+    for (std::size_t i = 0; i < keys.size(); ++i) {
+        // Empty on either side, refused here rather than at schema validation. An empty key cannot
+        // be named by `t("")`; an empty value would bake a zero-length run, which validates and
+        // draws nothing, so a missing translation would ship as a blank label rather than as a
+        // build failure. Both are recipe-authoring mistakes, and a diagnostic naming the index is
+        // more actionable than a schema error at line 0.
+        if (keys[i].empty()) {
+            report(diagnostics, std::string{recipePath}, 0, recipeStringsMalformed,
+                   std::format("[strings] key {} is empty", i));
+            return false;
+        }
+        if (values[i].empty()) {
+            report(diagnostics, std::string{recipePath}, 0, recipeStringsMalformed,
+                   std::format("[strings] value for '{}' is empty", keys[i]),
+                   "An empty translation bakes a run that draws nothing, which ships as a blank label.");
+            return false;
+        }
+        for (std::size_t j = 0; j < i; ++j) {
+            if (keys[j] == keys[i]) {
+                report(diagnostics, std::string{recipePath}, 0, recipeStringsMalformed,
+                       std::format("[strings] names the key '{}' twice", keys[i]),
+                       "A key becomes a run id, which is unique within a package.");
+                return false;
+            }
+        }
+        recipe.strings.push_back(TextString{.key = keys[i], .text = values[i]});
+    }
+    return true;
+}
+
 }  // namespace
 
 std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipePath,
@@ -367,6 +496,8 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
             return std::nullopt;
         }
         recipe.font = std::move(*spec);
+    } else if (!parseStrings(document, *package, recipe, recipePath, diagnostics)) {
+        return std::nullopt;
     }
 
     if (!isFontRecipe && recipe.atlas.empty()) {
@@ -678,32 +809,16 @@ using SlotIndex = std::vector<const atlas::GlyphSlot*>;
                                                      std::vector<cli::Diagnostic>& diagnostics) {
     const FontSpec& spec = *recipe.font;
 
-    // Confine the source path to the repository before opening it.
-    //
-    // `root / spec.source` is not the containment it looks like: std::filesystem's operator/
-    // *replaces* the left operand when the right is absolute, so `source = "/etc/shadow"` reads
-    // that file and ignores `root` entirely. A relative path with `..` components escapes the
-    // same way by a different route. The recipe is repository content rather than network input,
-    // so this is not a live attack surface today - but a baker that will one day run over a
-    // recipe somebody else wrote should not have "the recipe is trusted" as its only defence.
-    if (spec.source.empty() || std::filesystem::path{spec.source}.is_absolute()) {
+    // Confine the source path to the repository before opening it. See `confine()` for why the
+    // obvious `root / spec.source` is not containment.
+    const auto sourcePath = confine(root, spec.source);
+    if (!sourcePath.has_value()) {
         report(diagnostics, std::string{recipePath}, 0, recipeFontUnreadable,
-               std::format("font path '{}' must be relative to the repository root", spec.source));
-        return std::nullopt;
-    }
-    const std::filesystem::path sourcePath = (root / spec.source).lexically_normal();
-    const std::filesystem::path normalRoot = root.lexically_normal();
-    // lexically_relative() gives the path from root to source; a result that starts with ".."
-    // means source is outside root. Comparing the normalised forms is what makes `a/../../b`
-    // fail rather than pass on a string prefix check.
-    const auto relative = sourcePath.lexically_relative(normalRoot);
-    if (relative.empty() || *relative.begin() == "..") {
-        report(diagnostics, std::string{recipePath}, 0, recipeFontUnreadable,
-               std::format("font path '{}' resolves outside the repository root", spec.source));
+               std::format("font path '{}' must be relative to the repository root and stay inside it", spec.source));
         return std::nullopt;
     }
 
-    auto fontBytes = readFile(sourcePath);
+    auto fontBytes = readFile(*sourcePath);
     if (!fontBytes.has_value()) {
         report(diagnostics, std::string{recipePath}, 0, recipeFontUnreadable,
                std::format("cannot read font '{}'", spec.source),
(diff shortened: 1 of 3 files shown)


tools/text/TextBake.cpp
```diff
@@ -784,6 +899,212 @@ using SlotIndex = std::vector<const atlas::GlyphSlot*>;
...
+        return std::nullopt;
+    }
+
+    std::vector<std::byte> records;
+    records.reserve(points->size() * text::recordSize);
+    std::int64_t pen = 0;  // font units
+
+    for (std::size_t i = 0; i < points->size(); ++i) {
+        const char32_t point = (*points)[i];
+        const fontpkg::GlyphRecord* glyph = font.find(point);
```
ambroise-leclerc, line 1196: **[P1] Reject text that needs shaping even when its glyphs exist.**

The only eligibility check here is `font.find(point)`. That proves that an atlas slot exists, but it does not prove that an identity-mapped, left-to-right pen walk is a valid rendering of the string. A valid `FontPackage` can contain Arabic code points, combining marks such as U+0301, or another non-v1 repertoire. In those cases this loop succeeds and commits a plausible but incorrectly ordered/positioned run, despite the module contract and ADR-010 saying such input must produce no package.

For example, if a font recipe includes Arabic glyphs, every `find()` succeeds, but this code emits isolated glyphs in logical LTR order with no joining or bidi resolution. Likewise, a baked combining acute accent is placed after the base advance rather than attached to the base.

Please make the supported v1 shaping model mechanically enforceable before positioning—either by rejecting code points/sequences outside the explicitly supported repertoire here, or by adding validated font-package metadata that attests the allowed shaping model and checking it here. Add rejection tests where all glyphs exist, so the test proves the failure is about unsupported shaping rather than a missing glyph.

tools/text/TextBake.cpp
```diff
@@ -784,6 +899,212 @@ using SlotIndex = std::vector<const atlas::GlyphSlot*>;
...
+        }
+
+        // The record's index is into the *package's* glyph list, not the font's own glyph id -
+        // `mdux::text::draw::GlyphPlacement` documents the distinction, and swapping the two draws
+        // the wrong character rather than failing. `find()` returns a pointer into `font.glyphs`,
+        // so the index is that pointer's offset.
+        const auto packageIndex = static_cast<std::uint16_t>(glyph - font.glyphs.data());
+
+        const std::int64_t x = toPixels(pen, font.pixelSize, font.unitsPerEm);
+        if (x > std::numeric_limits<std::int16_t>::max()) {
```
ambroise-leclerc, line 1230: **[P1] Handle negative pen positions before narrowing to `int16_t`.**

`FontPackage::validate()` allows any `int16_t` kerning adjustment and does not require `advanceWidth + kerning >= 0`. A valid package can therefore drive `pen` negative. Two failures follow:

1. `toPixels()` adds half a unit before division, which is only correct for the documented non-negative case; negative values are biased toward zero.
2. This bounds check covers only `INT16_MAX`. An `x < INT16_MIN` reaches the `static_cast<std::int16_t>` below and is encoded as an implementation-defined/wrapped coordinate, potentially drawing the glyph far to the right.

Please either reject a pen transition that becomes negative, preserving the current non-negative contract, or implement sign-correct rounding and check `INT16_MIN <= x <= INT16_MAX` before the cast. A test with zero/small advances and a large negative valid kerning pair should pin both the diagnostic and the absence of output.

tools/text/TextBake.cpp
```diff
@@ -819,10 +1167,15 @@ std::optional<BakeOutputs> run(const Recipe& recipe, std::string_view recipePath
...
-    // TrueType and rasteriser inputs here; recording an empty list now keeps the report's inputs
-    // array present and structured rather than omitted.
+    // The font package is this bake's one input, and recording it is what makes the evidence chain
+    // hold: re-baking the font changes its package.json, which changes this report's input digest,
+    // which fails `evidence.text.<id>` until the runs are re-positioned against the new atlas. A
+    // package with no runs reads nothing and records an empty list - present and structured rather
+    // than omitted, so a reader distinguishes "no inputs" from "inputs not recorded".
     std::vector<evidence::FileRecord> inputs;
+    if (font.has_value()) {
+        inputs.push_back(fileRecord(recipe.fontPackage, font->bytes));
```
ambroise-leclerc, line 1320: **[P2] Normalize or reject platform-native separators before recording the evidence path.**

This records `recipe.fontPackage` verbatim. On Windows, a recipe value containing backslashes can successfully resolve and be read, but `BakeReport::validate()` rejects backslashes in recorded paths. The same recipe therefore fails later with generic `TXT005` on Windows and generally fails as an unreadable literal filename on Linux, violating the cross-toolchain evidence contract.

Please require repository paths to use `/` during recipe parsing, or normalize the accepted path to a repository-relative generic string before both resolution and report recording. A Windows-style path test should assert one stable result across toolchains.

tools/text/TextBake.cpp
```diff
@@ -367,6 +496,8 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
             return std::nullopt;
         }
         recipe.font = std::move(*spec);
+    } else if (!parseStrings(document, *package, recipe, recipePath, diagnostics)) {
```
ambroise-leclerc, line 553: **[P2] Reject recipes that mix the font and text forms.**

The document is classified as a font recipe solely by the presence of `[charset]`. Because `parseStrings()` is called only in this `else` branch, a recipe containing the complete font fields plus `[charset]`, `[package].font`, and `[strings]` succeeds as a font bake and silently discards every string.

For an evidence-producing tool, silently choosing one interpretation is risky: a copy/paste or merge error can produce a valid artifact of the wrong kind rather than a stable authoring diagnostic. Please make the forms explicitly mutually exclusive and add a test proving a mixed recipe is rejected.

tools/text/TextBake.cpp
```diff
@@ -91,6 +101,36 @@ void report(std::vector<cli::Diagnostic>& diagnostics, std::string file, std::si
...
+ *
+ * Shared by the font path's `source` and the text path's `font`, so the two cannot drift into
+ * disagreeing about what is inside the repository.
+ */
+[[nodiscard]] std::optional<std::filesystem::path> confine(const std::filesystem::path& root,
+                                                            std::string_view relativePath) {
+    if (relativePath.empty() || std::filesystem::path{relativePath}.is_absolute()) {
+        return std::nullopt;
+    }
+    const std::filesystem::path resolved   = (root / relativePath).lexically_normal();
```
ambroise-leclerc, line 134: **[P2] Lexical normalization does not confine paths through symlinks.**

This rejects absolute paths and lexical `..` escapes, but a repository entry such as `assets/external -> /outside/root` still produces a resolved path lexically under `root`; `readFile()` then follows the symlink outside it. The report records the repository-relative spelling while digesting external bytes, so the artifact can depend on uncommitted host state even though the API says inputs are confined to the repository.

Please compare canonical/weakly-canonical paths after resolving symlinks (taking care to canonicalize the root as well), or explicitly reject symlink components. Add a test with an in-root symlink targeting a readable file outside the fixture root.

ambroise-leclerc (commented):
Review summary: this is a safety-relevant change because it modifies ADR-010's build-time control against misrendered visible text. The implementation and tests are thoughtful, but the inline findings identify paths to plausible wrong packages or non-reproducible evidence and should be resolved before merge.

Please also record the explicit safety-impact classification, affected requirements/design statements and software items, verification evidence, and residual uncertainty in the PR description, as required by the repository's regulated-change workflow.

Current automated evidence is green on GCC 16, MSVC, sanitizers, and repository lints. I did not run a local build; `git diff --check develop...HEAD` passed.

tools/text/TextBake.cpp
ambroise-leclerc: You are right, and this is the finding that changes the design rather than the code. Fixed in 6dfe3ec.

`find()` proved an atlas slot existed and I treated that as proving the walk was valid, which it is not. A package baking U+0627 or U+0301 would have let this baker commit an artifact that validates, byte-compares on both toolchains, and renders wrongly — isolated unjoined letters running the wrong way, an accent parked after its base rather than over it. That is precisely the outcome ADR-010 exists to prevent, produced by the tool that is supposed to enforce it.

I took the first of your two options. Attesting the shaping model in the font package is the better long-term answer, but it is a governed-schema change that would re-bake every committed font package, and it would put the attestation in a file the baker cannot verify — a package could claim a model it does not satisfy. The repertoire lives with the code that implements it instead:

```cpp
constexpr std::array<RepertoireRange, 4> v1Repertoire{{
    {U

File: tools/text/TextBake.cpp (hunk 2 above)
Comment (ambroise-leclerc) [reply tools/text/TextBake.cpp:1230]: Confirmed and fixed in 6dfe3ec. I checked `FontPackage::validate()` rather than take it on faith: it bounds a kerning adjustment to an `int16_t` and never mentions `advance + kerning`, so a *valid* package can drive the pen negative — which is why the baker rather than the schema has to be the one to refuse.

I took your first option: reject the transition, preserving the non-negative contract. Sign-correct rounding plus a two-sided bound would work, but it would make the format express something it has no meaning for — a runs origin is its own left edge, and a glyph before it is an author meaning something a v1 record cannot say. Rejecting keeps one invariant instead of two behaviours.

That invariant is what makes `INT16_MAX` the only bound worth checking, and `toPixels()` now states the precondition rather than an assumption:

> **Precondition: `pen >= 0`.** The `+ unitsPerEm / 2` before the division is a half-*up* rule, which is only half-away-from-zero while the input is non-negative; on a negative pen it would bias toward zero and place a glyph a pixel right of where the same magnitude places it left.

The test uses your shape: `A` advances 700, the pair kerns -900, so the pen lands 200 units left of the origin. It asserts the package `validate()`s first — the point being that this is a legal artifact, not a corrupt one — then that the bake is `TXT027` and produces nothing.

File: tools/text/TextBake.cpp (hunk 3 above)
Comment (ambroise-leclerc) [reply tools/text/TextBake.cpp:1320]: Fixed in 6dfe3ec, by requiring `/` rather than normalising — your first option.

Normalising would accept a recipe that is wrong and quietly repair it, and the repair is only unambiguous on one platform: a backslash is a legal filename character on Linux, so `sub\font.json` there names a single file rather than a path, and translating it would invent a directory the author did not write. Refusing gives the same answer on both.

The rule went into `confine()`, which means the font recipes `source` gets it too — it recorded `spec.source` verbatim into `report.json` with the same exposure, and that was a latent defect this PR did not introduce but sits directly on.

The test asserts the cross-toolchain claim rather than the local one: `sub\font.json` is `TXT021` on Linux and on Windows, which the green Windows leg now shows. Without it the case would have passed on Linux for the accidental reason that the file does not exist.

File: tools/text/TextBake.cpp (hunk 4 above)
Comment (ambroise-leclerc) [reply tools/text/TextBake.cpp:553]: Confirmed and fixed in 6dfe3ec as `TXT028`.

I checked the exact failure mode you describe: `[charset]` alone classified the recipe, `parseStrings()` ran only in the `else`, so a recipe with both baked a valid font package and discarded every string in silence. A merge or a copy/paste is all it takes.

Your framing is the one I put in the code, because it is the general rule for this tool rather than a note about this branch:

> For a tool whose output is evidence, an artifact of the wrong kind that passes every check is worse than a stopped build.

Detection is unambiguous rather than heuristic: the two forms use different keys for their font input — `source` for a TrueType file, `font` for a baked package — so `[strings]` or `package.font` on a `[charset]` recipe can only be a mixed recipe. The test feeds a complete font recipe carrying both and asserts `TXT028`.

File: tools/text/TextBake.cpp (hunk 5 above)
Comment (ambroise-leclerc) [reply tools/text/TextBake.cpp:134]: Confirmed and fixed in 6dfe3ec. Lexical containment was not containment, and the consequence is the one you name: the report would record the repository-relative spelling while digesting bytes from outside, so the artifacts reproducibility would depend on the machine.

`confine()` now re-checks after `weakly_canonical`, keeping the lexical test first so an absolute path or a `..` escape is still refused without touching the filesystem.

Your parenthesis about canonicalising the root was not a detail — it is the whole thing. This repository is built under paths that are themselves symlinks (macOS resolves `/tmp` to `/private/tmp`, CIs workspace is a mount), so canonicalising only the target would have put every legitimate path outside a root spelled differently, and the font bake would have started failing for no reason. Both sides are canonicalised, and the reason is in the comment so it does not get "simplified" later.

The test is the one you asked for: an in-root symlink to a readable file outside the fixture root, refused as `TXT021`, with the same file read in place as the control so the scenario proves the symlink is what was rejected. It asserts only when `create_symlink` actually succeeded — Windows needs a privilege it usually lacks — because a test that silently passed on a platform it never exercised would be worse than one that admits it did not run.

File: tools/text/TextBake.cpp [hunk 6]
```diff
@@ -784,6 +899,212 @@ using SlotIndex = std::vector<const atlas::GlyphSlot*>;
...
+
+/// Font units to pixels, rounded half up, in integers throughout.
+///
+/// `pen` is non-negative for every run this baker produces - the pen starts at zero and advances -
+/// so half-up and half-away-from-zero are the same rule here. Written as half-up because that is
+/// what the expression says, rather than implying a negative case that cannot arise.
+[[nodiscard]] std::int64_t toPixels(std::int64_t pen, std::uint32_t pixelSize, std::uint16_t unitsPerEm) noexcept {
+    const auto scaled = pen * static_cast<std::int64_t>(pixelSize) + static_cast<std::int64_t>(unitsPerEm) / 2;
+    return scaled / static_cast<std::int64_t>(unitsPerEm);
+}
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in 6dfe3ec, together with [the maintainer thread](https://github.com/ambroise-leclerc/MduX/pull/239#discussion_r3838077754) on the same defect. The pen can indeed go negative — `FontPackage::validate()` bounds a kerning adjustment to an `int16_t` and says nothing about `advance + kerning` — and both consequences you name were real: the half-up rounding biases toward zero on a negative input, and the bounds check covered only `INT16_MAX` so the cast wrapped.

Resolved by making the non-negative pen an enforced precondition (`TXT027`) rather than by widening the rounding, since a glyph left of the runs own origin is not something a v1 record can express. `toPixels()` now documents the precondition instead of assuming it.

tools/text/TextBake.cpp
```diff
@@ -103,6 +143,15 @@ json::Value Recipe::toOptions() const {
...
     static_cast<void>(options.set("sidecar", json::Value::string(sidecar)));
+    if (!font.has_value()) {
+        // The resolved font package path, and how many runs were asked for. Not the strings
+        // themselves: they are the recipe's own content, the recipe's digest is recorded a few
+        // lines above this in every report, and their keys are already `package.json`'s run ids.
+        // Copying them here would make one edited translation appear in three places of the same
+        // artifact diff, which makes the diff harder to read rather than more complete.
+        static_cast<void>(options.set("font", json::Value::string(fontPackage)));
+        static_cast<void>(options.set("strings", json::Value::integer(strings.size())));
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Kept as-is, with the reasoning now in the code — one of six findings on this PR I did not act on.

The pair is not ambiguous: `parseStrings()` accepts `[strings]` and `font` **together or neither**, so an empty `font` always means "this package positions nothing", and `strings: 0` says the same thing a second time. Neither can occur without the other.

Emitting them conditionally would give text reports two structural shapes, and a reader comparing two of them would have to work out which shape each was before reading either. ADR-007s reason for recording resolved options is that a silently changed default must not leave every report looking unchanged; a field that appears and disappears works against that.

The comment now says so, so the next reader meets the argument rather than the empty string.

tools/text/TextBake.cpp
```diff
@@ -819,10 +1167,15 @@ std::optional<BakeOutputs> run(const Recipe& recipe, std::string_view recipePath
...
-    // array present and structured rather than omitted.
+    // The font package is this bake's one input, and recording it is what makes the evidence chain
+    // hold: re-baking the font changes its package.json, which changes this report's input digest,
+    // which fails `evidence.text.<id>` until the runs are re-positioned against the new atlas. A
+    // package with no runs reads nothing and records an empty list - present and structured rather
+    // than omitted, so a reader distinguishes "no inputs" from "inputs not recorded".
     std::vector<evidence::FileRecord> inputs;
+    if (font.has_value()) {
+        inputs.push_back(fileRecord(recipe.fontPackage, font->bytes));
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and fixed in 6dfe3ec — see [the maintainer thread](https://github.com/ambroise-leclerc/MduX/pull/239#discussion_r3838077755) for the detail.

Refused rather than normalised, and the rule went into the shared path check, so the font recipes `source` — which had the same latent exposure — is covered too. A backslash is a legal filename character on Linux, so normalising would invent a directory separator the author did not write; refusing gives one answer on both toolchains, which the green Windows leg now demonstrates.

tools/text/TextBake.cpp
```diff
@@ -91,6 +104,69 @@ void report(std::vector<cli::Diagnostic>& diagnostics, std::string file, std::si
...
+ */
+[[nodiscard]] std::optional<std::filesystem::path> confine(const std::filesystem::path& root,
+                                                            std::string_view relativePath) {
+    if (relativePath.empty() || std::filesystem::path{relativePath}.is_absolute()) {
+        return std::nullopt;
+    }
+    // A backslash is refused rather than translated. A recipe records repository paths with `/` on
+    // every platform, because `BakeReport::validate()` rejects a backslash in a recorded path - so
+    // a Windows-spelled recipe that resolved here would read its file and then fail much later, as
+    // a generic "the report is not valid", with nothing pointing at the recipe line that caused it.
```
ambroise-leclerc, line 128: **[P2] Fix the regression test so it actually reaches this backslash guard.**

The new test calls `codeFor("sub\\font.json")`. After C++ escape processing, `textRecipeText()` inserts one backslash into a TOML basic string, producing `font = "sub\font.json"`. The TOML parser recognizes `\f` and decodes it to a form-feed byte. Consequently, `Recipe::fontPackage` contains no backslash, this branch is not exercised, and the case still receives `TXT021` only because the control-character filename is unreadable/absent. Removing this guard would leave the test green.

Please encode a literal backslash through both layers—for example, pass four backslashes in the C++ string so the generated TOML contains `\\` and its parser produces one backslash—or assert the parsed `fontPackage` contains `'\\'` before running the bake. That makes the Windows/Linux claim a regression test of this rule rather than an accidental missing-file test.

tools/text/TextBake.cpp
```diff
@@ -784,6 +954,282 @@ using SlotIndex = std::vector<const atlas::GlyphSlot*>;
...
+ * whose whole meaning is conditional: it renders as nothing unless a line breaks there, and a pen
+ * walk has no line breaking, so it would bake a visible hyphen into a word. It is excluded with
+ * U+00A1's neighbour rather than by category, because there is no Unicode database here to ask.
+ */
+struct RepertoireRange {
+    char32_t first;
+    char32_t last;
+};
+
+constexpr std::array<RepertoireRange, 4> v1Repertoire{{
```
ambroise-leclerc: **[P2] Keep the enforced repertoire aligned with ADR-010's declared v1 scope.**

This makes the shaping limit mechanically enforceable, which fixes the original unsafe acceptance. However, the accepted architecture defines the complete v1 scope as “Latin/Cyrillic/Greek LTR text” (ADR-010, Context) and later defines unsupported scripts as anything *outside* Latin/Cyrillic/Greek LTR (Decision 5). This array contains only Basic/extended Latin. A font package can validly bake an identity-safe Greek or Cyrillic glyph, but the text baker now rejects it as `TXT026`.

That is a fail-closed result, but it is still a behavioral contraction and a direct code/document contradiction at a safety-relevant enforcement point. Please either:

- add the identity-safe Greek and Cyrillic repertoire promised by ADR-010, with present-glyph acceptance tests and exclusions for combining/contextual cases; or
- deliberately narrow v1 to Latin by updating ADR-010 and the affected traceability/risk-control description through the regulated-change process.

The implementation and accepted ADR should not describe different supported repertoires.

ambroise-leclerc (commented):
Follow-up review of corrections through `92a4d40`:

Resolved:
- negative pen movement now fails closed with `TXT027` and a valid-package regression test;
- mixed font/text recipe forms now fail with `TXT028`;
- symlink-based root escape is checked after canonicalization;
- unsupported shaping now has an explicit fail-closed repertoire check;
- backslash paths are rejected by the implementation before report generation.

Two issues remain in the inline comments:
1. the Latin-only repertoire conflicts with ADR-010's accepted Latin/Cyrillic/Greek v1 scope;
2. the backslash test feeds TOML's `\f` escape, so it does not exercise the backslash guard.

All current CI checks, including GCC 16, MSVC, sanitizers, security analysis, and repository lints, are green. The PR description still needs the explicit safety-impact classification and affected-artifact/verification record required by the regulated-change workflow.

tools/text/TextBake.cpp
ambroise-leclerc: Confirmed, and the test was worthless as written. Fixed in a20ba89.

I checked the parser rather than reason about it: `tools/common/Toml.cpp:173` lists `case 'f': out.push_back('\\f');` among the recognised escapes. So two backslashes in the C++ source put a lone `\\f` in the TOML, the parser decoded a form feed, and `Recipe::fontPackage` held `sub<FF>font.json` — no backslash anywhere. The case passed because a file named with a control character is absent, exactly as you say, and deleting the guard would have left it green.

Both of your remedies, because either alone still relies on my reading the layers correctly:

```cpp
const std::string backslashRecipe = textRecipeText("sub\\\\\\\\font.json", strings);
...
checks.expect(backslashParsed->fontPackage.find('\\\\') != std::string::npos,
              std::format("the parsed path keeps its backslash, got '{}'", backslashParsed->fontPackage));
```

Four backslashes so C++ yields two and the TOML parser yields one, and the parsed path asserted before the bake runs — so if a future change to either layer swallows it again, the scenario says *that* rather than continuing to report a passing rule.

Worth noting what this cost: I wrote that case specifically to make the cross-toolchain claim a regression test rather than an accident, and it was an accident anyway. Two escape layers were one more than I checked.

tools/text/TextBake.cpp
ambroise-leclerc: You are right, and I had the ADR available and did not read it. Fixed in a20ba89 by widening, not by narrowing the ADR.

`ADR-010:36` — "Latin/Cyrillic/Greek LTR text with no ligatures, no GPOS, no contextual forms and no RTL/contextual behaviour — the entire v1 scope of epic #14"; Decision 5 defines unsupported as anything outside that; and the negative consequences say it a third time. My array was Latin only, so an identity-safe Greek or Cyrillic letter would have been `TXT026`. Fail-closed, but your framing is the one that matters: code and an accepted ADR describing different repertoires is precisely the disagreement this enforcement exists to prevent.

Nine ranges now, with the marks carved out by range since there is no Unicode database here to ask by category:

- `U+0370..U+03FF` Greek and Coptic, whole — the combining Greek marks live in `U+0300..U+036F`, outside it.
- `U+0400..U+0482` and `U+048A..U+04FF` — Cyrillic with its own `U+0483..U+0489` (Mn and Me) removed.
- `U+0500..U+052F` Cyrillic Supplement, letters only.
- Latin as before, plus `U+1E00..U+1EFF` Latin Extended Additional, which is precomposed throughout.

Cyrillic Extended-A (`U+2DE0..U+2DFF`) is *entirely* combining marks and Extended-B carries several, so both stay out until somebody carves them by hand. The asymmetry is deliberate and now written down: omitting a block costs a build error, admitting one wrongly costs a device rendering nobody reviewed.

The scenario proves both directions against one fixture in which **every** code point resolves, so neither half can be an accident of the charset — a Greek alpha and a Cyrillic Zhe bake; a combining acute, a combining titlo *inside* the admitted Cyrillic block, and an Arabic alef are `TXT026`.

And the module doc now says the repertoire is not this file's to move: a baker enforcing less than the ADR ships a rendering nobody reviewed, one enforcing more refuses text the architecture promises, so a change here starts with a change to ADR-010.

AM-L (approved):

PR #240, opened 2026-08-23
Bake a text package, so a screen can carry text [#235]

### LINKED ISSUE #235: Bake a text package, so a screen can carry text

ambroise-leclerc:
Second of two for #235, on top of #239 (merged). **Closes #235.**

## What was blocked

`mdux-meduic` refuses a screen that draws text and names no approved locale — correctly, since a budget checked against no locale is a claim nobody made. With no committed text package, the endoscope screen had to be text-free, and its safety-critical node was a `NumericDisplay` chosen for being the one traceable component carrying no text key.

## What lands

`recipes/text/endoscope-monitor-en-us.toml` bakes into `generated/text/endoscope-monitor-en-us/` — 17 positioned records, one per character of "Endoscope Monitor" — and `EndoscopeMonitor.medui` gains a `Label` beside the brand mark. Its 1040x72 box is measured at build time against the widest translation every approved locale holds, which puts **#195's text-budget stage over a committed artifact for the first time**: it had only ever run over packages a test built for itself.

## ADR-011, made checkable

The compiled screen carries `"textKey": "STR-EM-TITLE"` and not the words. The pixel test now asserts that, because it is the one place the rule could break without anything else noticing — a compiler that resolved the key here would produce a screen that still validates, still renders identically today, and is wrong in every locale but one.

## The evidence chain

The screen's report gains three inputs: the font package, the text package, and its sidecar. Re-baking either invalidates the screen rather than leaving it describing text that moved. The text package's own report records the font package for the same reason.

## Caveats removed rather than reworded

`README.md`, `docs/architecture.md`, the `medui-authoring` skill, `ScreenPixelTests.cpp`, the roadmap, and both recipes. The skill now says what writing a screen with text actually takes: three files, not one, and a locale the font approves with no package listed is a build error.

## How the artifacts were produced

`mdux-bake-update`, in CI, since no local build is available to me — Homebrew's GCC 16 ships a 1-byte `bits/std.cc`, so `import std` has nothing to scan. A temporary workflow ran the target and uploaded `generated/`; it is deleted in this branch and appears nowhere in the diff.

That is the tool's own output, as ADR-007 requires, but "CI produced it" is not by itself a check — so I re-derived the 17 pen positions independently from the committed font package's advances and `unitsPerEm`/`pixelSize`, and they agree exactly:

```
expected x: [0, 10, 20, 30, 40, 49, 57, 67, 77, 87, 92, 106, 116, 126, 130, 137, 146]
actual   x: [0, 10, 20, 30, 40, 49, 57, 67, 77, 87, 92, 106, 116, 126, 130, 137, 146]
```

`evidence.text.endoscope-monitor-en-us` and `evidence.screen.endoscope-monitor` both pass on GCC 16 and MSVC, which is the byte-identity claim ADR-007 actually makes.

## One thing worth knowing for the next artifact

Introducing an artifact that another bake consumes does not build in one pass: the screen's dependency on `generated/text/.../package.json` is unsatisfiable until the text bake has staged it, and ninja fails at graph time rather than building what it can. Stage the new artifact, re-configure, then stage everything.

## Safety classification

**Potentially safety-relevant.** This changes build-time UI content and evidence for a screen that also contains the safety-critical `REQ-EM-001` node, but the new label is not safety-critical, remains deferred by the runtime, and does not change that node or the rendered pixels.

- **Requirements and risk controls:** none changed; `REQ-EM-001`, its annotation, bounds and golden reference are unchanged.
- **Affected software and evidence:** the host-side text/screen bake chain, the committed text and screen artifacts, and `ScreenPixelTests`, which verifies that the compiled label retains `STR-EM-TITLE` while the existing safety-critical trace remains discoverable.
- **Verification:** the committed text and screen artifacts are re-derived and byte-compared on GCC 16 and MSVC; the PR checks also cover the unchanged rendered panel.
- **Residual limitation:** runtime text drawing remains deferred to #17; this PR introduces no certification or production-readiness claim.

## Summary by CodeRabbit

* **New Features**
  * Added locale-aware text packages and compilation-time text measurement for the endoscope monitor screen.
  * Added a fixed-size screen title label using the `STR-EM-TITLE` translation.
  * Added baked font and text artifacts for the `en-US` locale.

* **Documentation**
  * Updated implementation status, architecture, roadmap, and screen documentation to reflect text-package support.
  * Clarified that runtime text rendering remains deferred.

* **Tests**
  * Updated compiled-screen and rendering expectations to include the new title label.


CMakeLists.txt
```diff
@@ -437,6 +437,28 @@ mdux_bake_artifact(
         atlas.bin
 )
 
+# The first committed text package (#235), and the artifact that lets a screen carry text at all.
+# Registered after the font it is measured against and before the screen that references it, which
+# is the order the evidence chain runs in: the font package is this bake's recorded input, and this
+# package is the screen's.
```
ambroise-leclerc: Registration order does not establish the evidence-chain execution order claimed here. mdux_bake_artifact() gives mdux-bake-update independent dependencies on every <artifact>-update target; there is no font → text → screen target edge. Ninja can therefore decide that the screen bake is already up to date against the old committed text package, then run the text update afterward. A wording change followed by the documented single mdux-bake-update can leave generated/text current while generated/screen/package.json and report.json still attest the preceding text digest. The analogous font → text → screen propagation has the same problem. Please make upstream staging an actual, enforced phase before dependent bakes are evaluated, or define and enforce a multi-pass workflow. Source registration order alone is not an ordering guarantee.

.agents/skills/medui-authoring/SKILL.md
```diff
@@ -10,7 +10,7 @@ governs the *authoring* of a `.medui` screen; for the baking mechanics behind it
 `evidence-pipeline`, and for the compliance framing of a safety-critical node see
 `regulatory-citations`.
 
-## Status: an authored screen reaches pixels; the text half is what remains
+## Status: an authored screen reaches pixels, text included up to the draw
```
ambroise-leclerc, line 13: The status heading is updated, but a stale sentence later in this same status section still says, ‘What it cannot yet do is carry text.’ That directly contradicts both this new heading and the artifact added by this PR: the screen now carries a text key and is budget-checked against a committed text package. The remaining limitation is that the runtime does not resolve/draw the text. Please remove or rewrite the stale sentence so the authoring skill does not send future agents back to the pre-#235 model.

ambroise-leclerc (commented):
These findings should be resolved before merge. The committed artifacts are internally consistent and the current checks are green, but the authoring update graph does not yet guarantee that an upstream font/text change is propagated into every dependent committed artifact in one documented update operation. Please also address Copilot's existing unresolved comment about declaring runs.bin as a screen dependency. Two additional documentation corrections are needed: docs/roadmap.md still reports six committed artifacts although the new text package makes seven, and the PR description needs the repository-required explicit safety classification and rationale. Safety-neutral may be appropriate if justified, but the classification still needs to be recorded.

CMakeLists.txt
ambroise-leclerc: You are right, and the comment was worse than no comment: it asserted a guarantee I had not checked. Fixed in 5a61f66.

I read the mechanism rather than argue from the diff. `mdux_bake_artifact()` ends with

```cmake
add_custom_target(${target_name}-update COMMAND ... DEPENDS ${baked_outputs} ...)
add_dependencies(mdux-bake-update ${target_name}-update)
```

so every `-update` hangs off the aggregate independently and depends only on its *own* baked outputs. There is no font → text → screen edge anywhere, and the update script writes into the source tree by a path ninja does not model, so it could not create one. The screen bake can therefore run against the previous committed text package and be staged attesting its digest. The font → text propagation you flag is the same shape.

I looked at enforcing a real edge and it does not work in this design. Making the *bake* target depend on the upstream's `-update` would order it correctly, and would also make an ordinary `cmake --build .` write into `generated/` — the rule this module enforces by construction, and CI's "Verify no source-tree writes" step. Ordering only the `-update` targets orders the copies while leaving the bakes free, which is the half that does not matter.

So I took your second option and wrote the multi-pass workflow down, in `cmake/MduXBake.cmake` beside the mechanism rather than at one call site — whoever adds the next consuming artifact reads that file, not this one. Two parts:

**Run the update to a fixpoint.** Stage, reconfigure, build, `ctest -L evidence`; a downstream artifact left behind fails its own comparison, so re-run and repeat until a pass stages nothing. A *new* consumed artifact needs this even to configure — the consumer's dependency on a file that does not exist yet is a ninja graph error, which is how this PR's own first staging attempt failed.

**Nothing escapes to a reviewer**, and I state the reason rather than the reassurance: CI builds from an empty tree, so every bake reads the committed inputs, and `report.json` records every input's digest — a screen staged against a superseded text package cannot match the bake that reads the current one, on either leg. The fixpoint rule is about not spending a CI round trip to learn it.

The CMakeLists comment now says only what is true: the placement is for a reader, and registration order is not an execution order.

.agents/skills/medui-authoring/SKILL.md
ambroise-leclerc: Fixed in 5a61f66 — and it is the more embarrassing of the two, since it sat three paragraphs below the heading I had just corrected in the same file.

It now reads: a `.medui` file carries text since #235, a `t("STR-KEY")` compiles and its box is measured against every approved locale's widest translation; what it cannot yet do is *draw* that text. Which is the limitation that actually remains, and points at #17 rather than back at the pre-#235 model.

The lesson I am taking from it: I edited the heading and the status paragraph and stopped there, without re-reading the section for anything else asserting the same fact. "Removed rather than reworded" only holds if the search is for the claim, not for the wording I remembered writing.

CMakeLists.txt
```diff
@@ -456,13 +478,20 @@ mdux_bake_artifact(
...
+# The two packages are named as SOURCES so the screen re-bakes when either moves. A forgotten entry
+# would not fail - it would make the committed screen quietly stop matching the text it was compiled
+# against, which is the drift the whole artifact exists to make impossible.
 include(cmake/MduXCompileScreen.cmake)
 mdux_compile_screen(
     ID endoscope-monitor
     RECIPE recipes/screen/endoscope-monitor.toml
+    SOURCES
+        generated/font/dejavu-ui/package.json
+        generated/text/endoscope-monitor-en-us/package.json
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and fixed in 5a61f66. `loadLocales()` reads the sidecar beside the package —

```cpp
const std::filesystem::path sidecarRelative = std::filesystem::path{relative}.parent_path() / package->sidecarPath;
...
inputs.push_back(fileRecord(sidecarRelative.generic_string(), *sidecar));
```

— and the committed screen report lists `generated/text/endoscope-monitor-en-us/runs.bin` among its inputs, so the file was read and undeclared.

The failure mode is narrower than a timestamp race and worse for being plausible: a translation edited to a different string of the same length repositions the runs, changes `runs.bin`, and leaves `package.json`'s byte length identical. `sha256` would differ, so `package.json` does change — but the dependency I want is on the bytes the compiler actually read, not on a file that happens to summarise them. It is now a `SOURCES` entry with that reasoning beside it.

No artifact bytes move: this adds a dependency edge, not an input.

README.md
```diff
@@ -38,7 +38,8 @@ A library with three zones, checked at configure time rather than by review
 `mdux_verify_trust_zones()` walks the full link graph and fails the configure step if a governed
 target ever reaches Vulkan or a windowing library.
 
-**What it is not:** a UI toolkit. There is no text, no layout, and no widgets yet — see
+**What it is not:** a UI toolkit. Text is compiled and measured but not yet drawn, and there are no
+widgets yet — see
```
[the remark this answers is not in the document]
ambroise-leclerc: Fair, and fixed in 5a61f66. "No widgets yet" was left over from before the component dictionary existed.

It now says what is actually true on each side: the dictionary is closed and compiled — a `Label`, a `Button`, a `NumericDisplay` and the rest are validated, laid out and budget-checked at build time — and the runtime draws only a `Panel`, so nothing else has an appearance yet. That is the distinction a reader needs, and it is the same one the implementation-status section makes further down.

AM-L (dismissed):

AM-L (approved):

PR #243, opened 2026-08-24
Draw a Label, so a compiled screen's text reaches pixels [#242]

### LINKED ISSUE #242: Draw a Label, so a compiled screen's text reaches pixels
Every piece needed to draw a `Label` now exists, and nothing joins them.

- `mdux::text::draw::recordRun()` records a baked run as `CoverageR8` rectangles against a font package (#162).
- A compiled screen carries `textKey` and bounds per node (#197, ADR-011 as amended by #203).
- A font package and a text package are both baked and committed (#160, #235).

`mdux::medui::render()` still draws only a `Panel`. On the committed screen that is one rectangle and five deferred nodes, one of which is the label #235 added. `Screen.cppm` says the reason is that "no text package is baked in this repository yet" — which stopped being true with #235.

This is the same shape as #235: a gap that surfaced under other issues' names and belongs to none of them.

## Work

**The join.** `render()` gains a way to receive the packages a locale-free screen must be joined to — the font package, the text package for the locale the device is running, and the sidecar bytes. Absent, `Label` stays deferred and every existing caller keeps its behaviour.

**A placement decision, made explicitly.** The shared contract's component model says nothing about where text sits inside its box, and `Screen.cppm` refuses on principle to invent per-component appearance. But something has to be decided, so the decision should be the one that is *already validated*: #195 measures the **ink** box, so the runtime should place the **ink** box at the node's origin. Then the property the compiler checked — this text fits this rectangle — is exactly the property the frame preserves. Any other rule (centring, or a baseline from an ascent the font package does not carry) would make the build-time guarantee not quite about what is drawn.

**The per-node work cap the module already promised.** `Screen.cppm` states that making per-node work a fact rather than a measurement "needs a type-level cap — TrustSC does it with `TextRuntime::<MAX_GLYPH_COMMANDS_PER_RUN>` — and that belongs with the first component whose geometry is variable". A `Label` is that component: its work scales with its run length. The cap lands here or the bounded-work claim quietly weakens.

## Acceptance

- The committed screen's label reaches pixels in `ScreenPixelTests`, compared against an expectation derived from the committed packages rather than from the renderer.
- `FrameStats::deferred` drops by one on that screen, and the count is asserted.
- Per-node work is bounded by a constant the module declares, not by the run's length.
- `ScreenNoHeapTests` still passes: the join allocates nothing.
- `Screen.cppm`'s "no text package is baked" paragraph is corrected rather than reworded.

## Out of scope

A `constexpr` emitter for font and text packages, so a device build stops parsing their JSON at startup. That is the shape #121 gave shaders and #153 asks for models, and it deserves its own issue rather than riding in on this one.

ambroise-leclerc:
Closes #242. The last link the chain was missing.

## What was missing

`recordRun()` has drawn a baked run since #162, a compiled screen has carried a `textKey` since #197, and both a font and a text package have been committed since #235. Nothing joined them. `render()` drew only a `Panel`, and `Screen.cppm` still justified that with "no text package is baked in this repository yet" — which stopped being true two PRs ago.

`render()` now takes a `TextBinding`: the font package, the text package for the locale the device is running, and the sidecar bytes. Defaulted, so every existing caller is unchanged and an unbound screen defers its labels exactly as before.

## Two decisions, made rather than stumbled into

**Where the glyphs go.** The shared component model says nothing about where text sits inside its box — I read the pinned contract rather than assume — and the font package carries no ascent, so there is no metric to derive a baseline from. Something has to be decided, so I picked the rule that is *already validated*: the run's **ink** box goes at the node's top-left corner, because that is exactly the box #195 measures against exactly that rectangle. Centring, or a baseline offset, would leave the build-time guarantee true of a rectangle nobody draws.

**The cap.** `Screen.cppm` has promised since #199 that making per-node work a fact rather than a measurement needs a type-level cap, "belonging with the first component whose geometry is variable". A `Label` is that component — its work scales with its run length. `maxGlyphsPerRun` is the cap, and a longer run is `RunTooLong` rather than truncated.

## Refusals

A key the bound package does not carry; a run whose range leaves the sidecar handed with it; a run past the cap. All three mean the caller bound the wrong thing for this screen, none can be true of packages this repository baked, and each rolls the frame back whole.

A run of blanks draws nothing and is **not** deferred: joined and found to paint nothing is a different fact from having nothing to join to.

## Evidence

The pixel scenario binds the **committed** artifacts — the same bytes `ctest -L evidence` compares — measures the ink box independently from them, then finds that box in the rendered frame and asserts its corner is the node's corner. I verified it can fail: shifting the origin by one pixel fails on `painted.left == label->bounds.x`.

The no-heap suite gains a case that renders *with* a binding, so "the join allocates nothing" is measured rather than asserted. That case caught a real allocation on its first run — my own `std::format` inside the measured loop, not the runtime. The scenario now keeps formatting out of it and says why.

## Verified locally

607/607 on `ninja-macos-clang`, the toolchain #241 made verified — `pixel`, `noheap`, `governed` and `evidence` included. The first PR in a while I could check before pushing rather than after.

## Out of scope

A `constexpr` emitter for font and text packages, so a device build stops parsing their JSON at startup — the shape #121 gave shaders and #153 asks for models. It deserves its own issue.

## Summary by CodeRabbit

* **New Features**
  * Labels can now render localized text using validated font and text packages.
  * Text is positioned at the label’s top-left corner and recorded as glyph runs.
  * Added clear errors for package mismatches, malformed runs, unknown text keys, and text that exceeds its label.

* **Bug Fixes**
  * Unbound labels are safely deferred.
  * Blank glyph runs no longer produce unnecessary drawing.

* **Documentation**
  * Updated runtime, architecture, roadmap, and README documentation to reflect current rendering capabilities.

### DIFF
--- a/.agents/skills/medui-authoring/SKILL.md
+++ b/.agents/skills/medui-authoring/SKILL.md
@@ -37,11 +37,14 @@ font approves, and the screen recipe's `[text]` table naming them — a locale t
 no package listed is a build error, because a budget checked against fewer locales than were approved
 is a claim nobody made.
 
-One limit is worth knowing before you write a screen: the runtime draws a `Panel`; every other
-component, a `Label` included, is visited, counted in `FrameStats::deferred` and left undrawn.
-Drawing text means joining a locale-free compiled screen to a text package for the locale the device
-is running, which is [#17](https://github.com/ambroise-leclerc/MduX/issues/17); live-data components
-have no geometry until the frame does.
+The runtime draws a `Panel` and, given a `TextBinding`, a `Label` (#242) — the join a locale-free
+compiled screen needs to reach glyphs: the font package, the text package for the locale the device
+is running, and its sidecar. Without one, a label is deferred rather than refused.
+
+One limit is worth knowing before you write a screen: every other component is visited, counted in
+`FrameStats::deferred` and left undrawn. A `Button` is more than its text — it has a face nothing in
+this project has decided — and live-data components have no geometry until the frame does. Both are
+[#17](https://github.com/ambroise-leclerc/MduX/issues/17).
 
 The HTML/CSS path that used to stand in for all of this - `UiFileWatcher::loadContent()`, which
 sniffed a file extension and stored the file as a string, with no parsing, layout or rendering
--- a/.agents/skills/medui-authoring/SKILL.md
+++ b/.agents/skills/medui-authoring/SKILL.md
@@ -63,8 +63,8 @@ A `.medui` file builds something in MduX today, and it reaches the screen: regis
 `mdux_compile_screen()` and it becomes a committed, byte-compared artifact plus generated C++ a
 device links, which the governed runtime draws and `ScreenPixelTests` compares pixel by pixel under
 lavapipe. It carries text too, since #235: a `t("STR-KEY")` compiles, and the box holding it is
-measured against every approved locale's widest translation. What it cannot yet do is *draw* that
-text, for the reason above.
+measured against every approved locale's widest translation. Since #242 the runtime draws it — bind
+the packages with `TextBinding::create()` and a label's glyphs reach the frame.
 
 ## Grammar shape
 
--- a/include/mdux/medui/Screen.cppm
+++ b/include/mdux/medui/Screen.cppm
@@ -54,6 +54,13 @@
  * property the compiler checked and the property the frame has the identical property. Centring, or
  * a baseline offset, would leave the build-time guarantee true of a rectangle nobody draws.
  *
+ * That argument holds only while the bound package is the one the compiler measured, and nothing in
+ * the artifacts says so (see `TextBinding`). So the same box is measured again here and refused as
+ * `TextOverflowsNode` when it does not fit, which costs one comparison over a walk the placement
+ * needs anyway. The build-time check remains the one that reports a *useful* diagnostic, at the
+ * right time, to the person who can fix it; this one exists so that a screen can never draw text
+ * over its neighbours because it was joined to the wrong package.
+ *
  * Two consequences worth stating rather than discovering. Leading placement means a right-to-left
  * locale would need a different rule - and cannot arrive without one, since `mdux-textbake` refuses
  * every RTL code point (ADR-010's v1 repertoire). And ink placement means a label whose text has no
@@ -122,7 +129,21 @@ enum class ScreenError : std::uint8_t {
     UnknownTextKey,       ///< a bound text package carries no run for a node's `textKey`
     MalformedTextRun,     ///< a run's range leaves the sidecar, or its bytes are not whole records
     RunTooLong,           ///< a run holds more than `maxGlyphsPerRun` records
+    AtlasMismatch,        ///< the text package was baked against a different font package
+    SidecarMismatch,      ///< the sidecar is not the one the text package describes
+    TextOverflowsNode,    ///< a run's ink is wider or taller than the node that names it
 };
+
+// The two token failures are kept apart because the schema keeps them apart, and for its reason: a
+// malformed name is a defect in whatever emitted the screen, while an absent one is a table that
+// does not define it. Collapsing them would tell an integrator to look in the wrong place.
+
+[[nodiscard]] std::string_view describe(ScreenError error) noexcept;
+
+// `AtlasMismatch` and `SidecarMismatch` are `create()`'s, not a frame's: both are properties of the
+// three artifacts a caller bound together, decided once. `TextOverflowsNode` is a frame's, and is
+// the check that makes the placement rule below safe against a binding this module cannot
+// authenticate - see `TextBinding` for what that means and what it does not.
 
 /**
  * @brief The largest run this runtime will draw, in records.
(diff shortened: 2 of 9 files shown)


src/medui/Screen.cpp
```diff
@@ -46,6 +49,80 @@ namespace {
...
+
+/// The run a node's `textKey` names, as a span of the bound sidecar.
+///
+/// Both failures are the caller's binding rather than the screen's: a key the package does not carry
+/// means the wrong package was bound for this screen, and a range leaving the sidecar means the
+/// package and the bytes handed with it do not belong together. Neither can be true of a package
+/// this repository baked, and both are cheap to refuse rather than to trust.
+[[nodiscard]] mdux::core::Result<std::span<const std::byte>, ScreenError> runFor(const TextBinding& binding,
+                                                                                  std::string_view   textKey) noexcept {
+    for (const mdux::text::TextRun& run : binding.text->runs) {
```
ambroise-leclerc: **[P1] Bind the screen to the text package(s) whose runs were budget-checked.**

Looking up only `textKey` does not establish that this is the text package used by the `.medui` compiler. A different, individually valid text package can use the same font and carry the same common key (for example `STR-TITLE`), so the proposed atlas-id check alone would still accept it. Its run may contain different wording or wider geometry than the run #195 measured, allowing text to escape the node bounds and overlap other UI while `render()` reports success.

Please make the compiled screen carry enough approved binding identity—such as the expected package id/digest per approved locale—and validate that identity when creating or consuming `TextBinding`. Add a test using a second valid package with the same atlas and key but different run geometry; it should fail closed and preserve the list.

src/medui/Screen.cpp
```diff
@@ -46,6 +49,80 @@ namespace {
...
+        if (run.id != textKey) {
+            continue;
+        }
+        if (run.byteEnd() > binding.runs.size()) {
```
ambroise-leclerc: **[P1] Authenticate the sidecar before treating it as this package’s run data.**

This proves only that the selected range fits the supplied span. `TextPackage` already records `sidecarByteLength`, `sidecarSha256`, and a digest per run, but none is compared with `binding.runs`. A sidecar from another package—or corrupted bytes of the same length—therefore passes whenever its decoded glyph indexes happen to be in range and can display plausible but incorrect text.

Because hashing the whole sidecar per frame would weaken the bounded-work story, consider making `TextBinding` constructible only through a one-time, allocation-free validation step that checks package structure, exact sidecar length/digest, font identity, and screen/package identity. `render()` can then consume that proof cheaply. Please include a same-length wrong-sidecar test; the current short-sidecar case does not cover this failure mode.

src/medui/Screen.cpp
```diff
@@ -46,6 +49,80 @@ namespace {
...
+            return mdux::core::err(ScreenError::MalformedTextRun);
+        }
+        if (run.byteLength % mdux::text::draw::recordSize != 0) {
+            return mdux::core::err(ScreenError::MalformedTextRun);
+        }
+        if (run.byteLength / mdux::text::draw::recordSize > maxGlyphsPerRun) {
+            return mdux::core::err(ScreenError::RunTooLong);
+        }
+        return binding.runs.subspan(static_cast<std::size_t>(run.byteOffset), static_cast<std::size_t>(run.byteLength));
```
ambroise-leclerc: **[P2] Make the range check overflow-safe before calling `subspan()`.**

`run.byteEnd()` is `byteOffset + byteLength`. For a malformed hand-built package such as `byteOffset = UINT64_MAX - 5` and `byteLength = 6`, that addition wraps to zero, so line 112 accepts the range and this `subspan()` violates its `offset <= size()` precondition. The run-length cap does not prevent this because six bytes is only one record.

Please use the subtraction form already established in `TextPackage::validate()`:

```cpp
if (run.byteOffset > binding.runs.size() ||
    run.byteLength > binding.runs.size() - run.byteOffset) {
    return err(ScreenError::MalformedTextRun);
}
```

Account for the `uint64_t`/`size_t` conversion explicitly, and add the wraparound case to the refusal test so the promised fail-closed behavior is exercised.

ambroise-leclerc:
## Safety-relevant review — blocking findings

The label path is well tested for placement, rollback, bounded run length, and allocation behavior, and the focused local scenarios passed. The remaining blockers are at the artifact join boundary:

1. The compiled screen does not identify the text package(s) whose translations were measured, so another valid package sharing a key can bypass the build-time bounds guarantee.
2. The supplied sidecar is not authenticated against the package metadata, so same-length wrong or corrupted bytes can render as plausible text.
3. The malformed-run range check can wrap before `subspan()`, violating its precondition instead of failing closed.

Existing review comments also correctly identify the font/text atlas-id mismatch, zero-byte binding behavior, and stale documentation status. These should be resolved together through a validated binding boundary rather than a collection of per-frame partial checks.

**Disposition:** changes requested. GitHub does not permit the PR author to submit a formal request-changes review on their own PR, so this is recorded as a blocking review comment instead.

**Impact classification:** safety-relevant, because this changes displayed content and refusal behavior in the governed runtime. Under the repository policy it requires maintainer/domain-expert review before merge.

**Local validation:** `git diff --check` passed; tests 456–459 and 597 passed on macOS arm64, CMake 4.3.1, Clang 21.1.8. The local offscreen test was skipped; all PR CI checks, including platform builds and pixel evidence, were green at review time.

src/medui/Screen.cpp
ambroise-leclerc: You are right, and this is the finding that reshaped the API. Fixed as far as it can be here in 310a8bf, and the remainder is now #244 rather than left implicit.

What I did not appreciate when I wrote the placement argument: it rests entirely on "#195 measured this box against this rectangle", and **nothing at run time established that the bound package was the measured one**. The whole justification was conditional on a premise no artifact carried.

Two halves, and they need different answers.

**The consequence is closed here.** `render()` now measures the ink box it is about to place and refuses `TextOverflowsNode` when it does not fit the node. That costs one comparison over a walk the placement already needs, and it means a screen can never draw text over its neighbours because it was joined to the wrong package — the specific harm you name. The test uses your shape: a second valid package, same font, same key, fifty glyphs at ten pixels of advance into a hundred-pixel node, refused with the list left as found.

**The identity needs the compiled screen to carry it**, which is a change to what a screen emits — the governed schema, the canonical package, both emitters, the committed artifact and ADR-012. That does not belong in a pull request about the runtime, so it is [#244](https://github.com/ambroise-leclerc/MduX/issues/244), specified with the acceptance test written to fail for *identity* rather than for overflow: same font, same key, and geometry that fits.

`Screen.cppm` now states the limit in the class's own documentation rather than in a commit message, under "What it still does not prove".

src/medui/Screen.cpp
ambroise-leclerc: Confirmed and fixed in 310a8bf, along the lines you sketched — a one-time validation whose result `render()` consumes cheaply.

You were right that nothing compared the bytes: `TextPackage` records `sidecarByteLength`, `sidecarSha256` and a digest per run, and I used none of them. `TextBinding` is no longer an aggregate; `create()` is the only way to obtain one, and it checks:

- `text.atlasId == font.id` — otherwise the records index a glyph table that assigns different shapes to the same numbers;
- `runs.size() == sidecarByteLength`, then `sha256(runs) == sidecarSha256`;
- per run: an overflow-safe range, whole records, within the cap, **and its own digest** — because the whole-file hash cannot see a package whose *ranges* were rewritten to point at each other's runs.

`render()` per frame is now a `find()` and a bounds comparison, so the bounded-work story is unchanged.

The same-length wrong-sidecar case you asked for is in the corpus, and it is the one that would have passed everything before: same length, valid records, in-range glyph indexes, different words.

One thing worth stating plainly: "only through `create()`" is enforced by the language, not by prose — `static_assert(!std::is_aggregate_v<TextBinding>)` sits beside the class, so an edit that makes the members public "for convenience" fails there instead of silently reopening the bypass. I added that after checking, and the check was worth running: the incremental build had been serving a **stale module BMI**, so an aggregate initialisation of `TextBinding` kept compiling against an interface that no longer existed. A clean rebuild is what surfaced it.

src/medui/Screen.cpp
ambroise-leclerc: Confirmed and fixed in 310a8bf, with the exact form you gave.

I verified the wrap rather than take it on faith: `byteOffset = UINT64_MAX - 5` with `byteLength = 6` sums to zero, so the addition form accepted the range and `subspan()` was handed an offset past its size. And you are right that the cap does not save it — six bytes is one record.

```cpp
if (run.byteOffset > runs.size() || run.byteLength > runs.size() - run.byteOffset) {
    return mdux::core::err(ScreenError::MalformedTextRun);
}
```

Now in `create()` rather than per frame, and the comment points at `TextPackage::validate()` as the place this form is already established — a package that reached here without going through it is precisely the case the function exists for.

The wraparound case is in the refusal corpus, so the fail-closed behaviour is exercised rather than promised. The same helper in the pixel test got the same treatment: it checked bounds *after* forming the span, which Copilot flagged separately.

include/mdux/medui/Screen.cppm
```diff
@@ -87,6 +119,57 @@ enum class ScreenError : std::uint8_t {
...
+    const mdux::text::TextPackage* text{nullptr};
+
+    /// The sidecar bytes `text` addresses ranges of - the committed `runs.bin`, or a mapping of it.
+    std::span<const std::byte> runs{};
+
+    /// Whether all three are present. A partial binding is treated as no binding rather than as an
+    /// error: a caller assembling one incrementally should see deferred nodes, not a refused frame.
+    [[nodiscard]] constexpr bool complete() const noexcept {
+        return font != nullptr && text != nullptr && !runs.empty();
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and fixed in 310a8bf. You are right that a zero-byte sidecar is a legitimate package — `TextPackage::validate()` accepts one with no runs — and `complete()` classified it as "no binding", so its labels were deferred for a reason that was not true.

The predicate is now `bound()`, and it asks only whether the packages are present. A package with no runs is bound; a label whose key it does not carry is then `UnknownTextKey`, which is the accurate answer rather than silence.

The empty-translation case you mention is also distinguished now, one layer up: a run that decodes to nothing but blanks is drawn — zero rectangles — and **not** counted as deferred, because "joined and found to paint nothing" is a different fact from "nothing to join to". There is a scenario pinning exactly that.

tests/render/ScreenPixelTests.cpp
```diff
@@ -123,6 +129,143 @@ void recordFrame(VkCommandBuffer commandBuffer, void* context) {
...
+    InkBox box;
+    for (std::size_t offset = 0; offset < records.size(); offset += mdux::text::draw::recordSize) {
+        const auto placement = mdux::text::draw::decodeRecord(records.subspan(offset, mdux::text::draw::recordSize));
+        if (!placement.has_value()) {
+            return InkBox{};
+        }
+        const mdux::font::GlyphRecord& glyph = bound.font.glyphs[placement->packageIndex];
+        if (glyph.isBlank()) {
+            continue;
+        }
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and fixed in 310a8bf. The checks were after the span existed rather than before, which is the wrong order for a helper that reads committed bytes.

```cpp
if (run->byteOffset > bound.runs.size() || run->byteLength > bound.runs.size() - run->byteOffset
    || run->byteLength % mdux::text::draw::recordSize != 0) {
    return InkBox{};
}
```

Subtraction rather than addition, for the wraparound the maintainer flagged on the runtime side of the same arithmetic. `packageIndex` is bounds-checked against `font.glyphs` before indexing too.

Returning "no ink" rather than throwing is deliberate: the scenario's `REQUIRE(derived.found)` then fails naming the run, which is a diagnosis, where undefined behaviour is not.

Worth noting the exposure was smaller than it looks and the fix still worth making — these bytes come from `generated/`, so a corrupt artifact would already have failed `ctest -L evidence`. But a test helper that is only safe because *another* test passed is not a property I want to rely on.

.agents/skills/medui-authoring/SKILL.md
```diff
@@ -37,11 +37,14 @@ font approves, and the screen recipe's `[text]` table naming them — a locale t
...
-is running, which is [#17](https://github.com/ambroise-leclerc/MduX/issues/17); live-data components
-have no geometry until the frame does.
+The runtime draws a `Panel` and, given a `TextBinding`, a `Label` (#242) — the join a locale-free
+compiled screen needs to reach glyphs: the font package, the text package for the locale the device
+is running, and its sidecar. Without one, a label is deferred rather than refused.
+
+One limit is worth knowing before you write a screen: every other component is visited, counted in
+`FrameStats::deferred` and left undrawn. A `Button` is more than its text — it has a face nothing in
+this project has decided — and live-data components have no geometry until the frame does. Both are
+[#17](https://github.com/ambroise-leclerc/MduX/issues/17).
```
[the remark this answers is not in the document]
ambroise-leclerc: Both fixed in 310a8bf, and you were right to consolidate them — I had corrected the headline claims in each file and not re-read the sections for other statements of the same fact. That is the second time this session, so the lesson is landing: search for the *claim*, not for the wording I remember writing.

- `SKILL.md`: the sentence three paragraphs below the heading now says the runtime draws text since #242 and points at `TextBinding::create()`, instead of saying it cannot.
- `README.md`: "Text and glyph rendering" was not merely mislabelled, it was in the **Not started** block — while `mdux-textbake`, the committed atlas, the coverage draw path and now the runtime join all exist. Moved into the implemented rows with what is and is not drawn stated: a `Label` is, a `Button`'s text is not, because a button is more than its text (#17).

src/medui/Screen.cpp
```diff
@@ -96,6 +180,67 @@ mdux::core::Result<FrameStats, ScreenError> render(const ScreenPackage& screen,
...
+
+            const auto records = runFor(text, label->textKey);
+            if (!records.has_value()) {
+                return refuse(records.error());
+            }
+
+            const auto ink = measureInk(*text.font, *records);
+            if (!ink.has_value()) {
+                return refuse(ink.error());
+            }
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed, and fixed more broadly than proposed, in 310a8bf.

The atlas check you name is real and is now in `TextBinding::create()` as `AtlasMismatch`, with your reasoning in the code: a package baked against another font makes every glyph index still in range, so nothing downstream can detect it and the frame draws a plausible sentence made of the wrong letters.

But the maintainer's [P1 on the same lines](https://github.com/ambroise-leclerc/MduX/pull/243#discussion_r3847763932) shows the atlas check alone is not sufficient: a second package with the *same* font and the same key passes it and can still carry wider or different text. So `create()` also authenticates the sidecar by length and `sha256`, and each run by its own digest; `render()` re-measures the ink and refuses `TextOverflowsNode`; and the part none of that can reach — that this is the package the compiler measured — is [#244](https://github.com/ambroise-leclerc/MduX/issues/244), because it needs the compiled screen to carry the identity, which changes what a screen emits (ADR-012).

The mismatch test is in the refusal corpus, and the rollback is asserted on the frame-level refusals where a list exists to roll back.

AM-L (approved):

PR #245, opened 2026-08-25
Bind a screen to the text packages it was compiled against [#244]

### LINKED ISSUE #244: Bind a screen to the text packages it was compiled against
`TextBinding::create()` (#242) proves that a font package, a text package and a sidecar describe each other. It cannot prove that the text package is the one `mdux-meduic` measured the screen against, because nothing in the artifacts links them.

## The gap

A second, individually valid text package — same font, same `atlasId`, carrying the same key, with its own consistent sidecar and digests — passes every check `create()` makes. Its run may hold different wording than #195 measured. `render()` then draws it.

The *safety* consequence is already closed: #242 re-measures the ink and refuses `TextOverflowsNode` rather than drawing over a node's neighbours. What remains is a correctness one — a device can display words no reviewer approved for that screen, and every artifact still verifies.

## What would close it

The compiled screen carries the identity of the packages it was compiled against — per approved locale, the package id and its `package.json` digest — and `TextBinding::create()` takes the screen and checks the bound package against that list.

The compiler already knows both: `report.json` records every text package as an input with its digest. What changes is that the *package* — the thing a device holds — carries it too.

## Why it is not part of #242

It changes what a compiled screen emits, which is ADR-012's subject. That means the governed schema, the canonical package JSON, both emitters, the committed screen artifact and the ADR itself, none of which belongs in a pull request about the runtime.

## Work

- A record type in `mdux.medui.schema` and a span on `ScreenPackage`; `validate()` covers it.
- `mdux-meduic` emits it; both emitters carry it into generated C++.
- `TextBinding::create()` gains a `ScreenPackage` parameter and a `PackageNotApproved` refusal.
- Re-bake `generated/screen/endoscope-monitor/`, and update ADR-011/ADR-012.

## Acceptance

- A screen bound to a valid package it was not compiled against fails to render, with its own code.
- The test uses a second package with the same font, the same key and *fitting* geometry, so it fails for identity rather than for overflow.

ambroise-leclerc:
## Summary

Bind each compiled screen to the exact text packages reviewed during compilation. The screen artifact now carries an approved package ID and `package.json` SHA-256 per locale, and `TextBinding::create()` fails closed with `PackageNotApproved` when a different valid package is supplied.

The compiler, canonical JSON representation, both C++ emitters, runtime binding, committed endoscope screen artifact, tests, and ADR-011/ADR-012 are updated together.

Closes #244

- [x] Maintainer-authored change; an invitation is not applicable.
- [x] Maintainer-authored change; contributor CLA acceptance is not applicable.

Invitation / CLA issue: none — maintainer-authored

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [ ] `FILE_SET CXX_MODULES` lists — a module interface added, removed or moved
- [ ] `tests/CMakeLists.txt` — test source lists and suite membership
- [x] Committed artifacts under `generated/`
- [ ] None of the above

## Rebase state

- **Rebased onto:** `b1750b606403e2c94fac0038482a9d45bf1c6579`

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run locally; macOS verified tuple used
- [ ] `ctest --test-dir build-gcc` — not run locally; macOS verified tuple used
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — passed, together with docs-lint unit tests, file-header lint, AI-reference check, and schema drift check
- [x] Other: `cmake --build build-macos-clang -j 4` passed
- [x] Other: `ctest --test-dir build-macos-clang --output-on-failure -j 4` — 611 passed, 0 failed, 1 skipped
- [x] Other: `ctest --test-dir build-macos-clang -L evidence --output-on-failure -j 4` — 554 passed
- [x] Other: `cmake --build build-macos-clang --target mdux-bake-update -j 4` — committed artifact regenerated and byte-identity checks passed

The skipped test is `offscreen_tests`: the local host has no usable Vulkan loader/ICD. The target and generated-screen approval assertions compile; CI must execute the pixel leg.

## Post-merge gate

- [ ] That run is green. *(No successor; nothing is gated.)*

## Regulatory impact

This is a safety-relevant, fail-closed approval-integrity change: a device can bind only a text package whose locale, package ID, and canonical package digest were recorded when the screen was compiled. It changes governed schema/runtime behavior and committed traceability evidence. ADR-011, ADR-012, runtime refusal tests, no-allocation tests, and the generated screen package/report were updated. Maintainer/domain review is required before merge. No certification or standards-compliance claim is introduced.

## Summary by CodeRabbit

* **New Features**
  * Compiled screen packages now record approved locales, text package IDs, and SHA-256 digests.
  * Runtime rendering verifies that text packages match the screen’s approved content.
  * Package metadata supports reliable round-trip storage and validation of approved translations.

* **Bug Fixes**
  * Prevents unapproved, modified, or noncanonical text packages from being used.
  * Rejects text approvals that are missing required metadata or applied to text-free screens.

* **Documentation**
  * Updated architecture guidance for locale approvals and authenticated text binding.

### DIFF
--- a/include/mdux/medui/Schema.cppm
+++ b/include/mdux/medui/Schema.cppm
@@ -31,16 +31,18 @@
  * That is the same split `mdux.ml.schema` makes with `mdux-mlbake`, and it is the price of a type a
  * device can hold in `.rodata`.
  *
- * ## The package is locale-free, and that is load-bearing
+ * ## The layout is locale-free; the approved package identities are not
  *
  * ADR-011, as amended by #203, carries `textKey` and `colorToken` as **validated names** rather than
  * as glyph runs and RGBA8. The device resolves each against a governed table for the locale it is
  * running - a bounded lookup, not a parse - and the per-locale glyph runs stay in the text package
  * where ADR-010 put them.
  *
- * The consequence that decides the file layout: **adding an approved locale rewrites no screen
- * artifact.** Translations change far more often than layouts, and a package carrying runs would
- * make the frequent change rewrite the stable artifact - and its digest, and its review.
+ * The screen still carries no locale-selected glyph run, so one layout serves every approved
+ * locale. It does carry the identity of every text package the compiler measured: locale, package
+ * id and canonical `package.json` digest. A changed translation therefore rewrites this small
+ * approval manifest, deliberately - otherwise a different, individually valid package could be
+ * substituted after review while every local consistency check still passed.
  *
  * The cost, stated as a cost: one rectangle serves every locale, so it must be the one that
  * survives the widest approved translation. That is what #195 measures, and why the budget below
@@ -83,6 +85,7 @@ export module mdux.medui.schema;
 import std;
 import mdux.core.result;
 import mdux.draw;
+import mdux.evidence.digest;
 import mdux.evidence.report;
 
 export namespace mdux::medui {
@@ -111,6 +114,9 @@ enum class SchemaError : std::uint8_t {
     NonPositiveMaxLength,      ///< a text input that can hold no character
     EmptyBudget,               ///< a screen with nodes whose budget can hold no primitive
     BudgetExceedsIndexWidth,   ///< more vertices than a 16-bit index can address
+    EmptyApprovedLocale,       ///< a text-package approval does not name its locale
+    EmptyApprovedPackageId,    ///< a text-package approval does not name its package
+    DuplicateApprovedLocale,   ///< two text-package approvals claim the same locale
 };
 
 [[nodiscard]] constexpr std::string_view describe(SchemaError error) noexcept {
@@ -147,6 +153,12 @@ enum class SchemaError : std::uint8_t {
             return "the screen has nodes and a budget that can hold no primitive";
         case SchemaError::BudgetExceedsIndexWidth:
             return "the vertex budget exceeds what a 16-bit index can address";
+        case SchemaError::EmptyApprovedLocale:
+            return "an approved text package does not name its locale";
+        case SchemaError::EmptyApprovedPackageId:
+            return "an approved text package does not name its package id";
+        case SchemaError::DuplicateApprovedLocale:
+            return "two approved text packages claim the same locale";
     }
     return "unknown schema error";
 }
@@ -547,18 +559,34 @@ static_assert(std::variant_size_v<NodePayload> == 11, "an alternative was added 
 }
 
 /**
+ * @brief One text package this screen was compiled and reviewed against.
+ *
+ * The package remains outside the screen: it owns the locale's glyph runs and sidecar. This record
+ * is only its identity, sufficient for `TextBinding::create()` to refuse a different package even
+ * when that package is internally valid, uses the same font and carries the same keys.
+ */
+struct TextPackageApproval {
+    std::string_view locale;           ///< the package's BCP 47 locale
+    std::string_view packageId;        ///< the package header id
+    evidence::Digest packageSha256{};  ///< SHA-256 of its canonical `package.json`
+
+    [[nodiscard]] constexpr bool operator==(const TextPackageApproval&) const noexcept = default;
+};
+
+/**
  * @brief A whole compiled screen as generated code exposes it and the runtime consumes it.
  *
  * Non-owning and `constexpr`-constructible throughout, so a generated translation unit can place one
  * in read-only memory and `static_assert` that it validates.
  */
 struct ScreenPackage {
-    std::string_view              id;
-    std::uint64_t                 schemaVersion{evidence::kSchemaVersion};
-    std::int32_t                  surfaceWidth{0};
-    std::int32_t                  surfaceHeight{0};
-    std::span<const CompiledNode> nodes;
-    mdux::draw::DrawBudget        budget{};
+    std::string_view                     id;
+    std::uint64_t                        schemaVersion{evidence::kSchemaVersion};
+    std::int32_t                         surfaceWidth{0};
+    std::int32_t                         surfaceHeight{0};
+    std::span<const TextPackageApproval> approvedTextPackages;
+    std::span<const CompiledNode>        nodes;
+    mdux::draw::DrawBudget               budget{};
 
     /// Checks every invariant a consumer is entitled to assume. See the module comment for the one
     /// invariant it deliberately leaves to the compiler: whether the budget is *large enough*.
@@ -764,6 +792,21 @@ constexpr mdux::core::ResultVoid<SchemaError> ScreenPackage::validate() const no
     }
     if (surfaceWidth <= 0 || surfaceHeight <= 0) {
         return err(SchemaError::NonPositiveSurface);
+    }
+
+    for (std::size_t index = 0; index < approvedTextPackages.size(); ++index) {
+        const TextPackageApproval& approval = approvedTextPackages[index];
+        if (approval.locale.empty()) {
+            return err(SchemaError::EmptyApprovedLocale);
+        }
+        if (approval.packageId.empty()) {
+            return err(SchemaError::EmptyApprovedPackageId);
+        }
+        for (std::size_t earlier = 0; earlier < index; ++earlier) {
+            if (approvedTextPackages[earlier].locale == approval.locale) {
+                return err(SchemaError::DuplicateApprovedLocale);
+            }
+        }
     }
 
     for (std::size_t index = 0; index < nodes.size(); ++index) {
--- a/include/mdux/medui/Schema.cppm
+++ b/include/mdux/medui/Schema.cppm
@@ -98,25 +98,27 @@ inline constexpr std::string_view packageKind = "screen";
 inline constexpr std::string_view colorTokenPrefix = "Theme.Colors.";
 
 enum class SchemaError : std::uint8_t {
-    UnsupportedSchemaVersion,  ///< the package declares a version this module does not read
-    EmptyId,                   ///< the screen has no id, so no directory and no evidence entry
-    NonPositiveSurface,        ///< a surface with no extent cannot contain a rectangle
-    EmptyNodeId,               ///< a node with no id cannot be named by a golden or a requirement
-    DuplicateNodeId,           ///< two nodes share an id, so a golden could name either
-    DegenerateBounds,          ///< a rectangle with no extent, which `DrawList` refuses to record
-    BoundsOutsideSurface,      ///< a rectangle the declared surface does not contain
-    MalformedColorToken,       ///< a colour that is not a `Theme.Colors.<Token>` name
-    UnknownColorToken,         ///< a well-formed name the governed table does not define
-    UnknownPayload,            ///< a payload this module cannot name, or one left valueless
-    EmptyRequiredName,         ///< a spec field the component dictionary requires is empty
-    NoStates,                  ///< a status indicator that can show nothing
-    StateColorCountMismatch,   ///< per-state tints that do not pair one-to-one with the states
-    NonPositiveMaxLength,      ///< a text input that can hold no character
-    EmptyBudget,               ///< a screen with nodes whose budget can hold no primitive
-    BudgetExceedsIndexWidth,   ///< more vertices than a 16-bit index can address
-    EmptyApprovedLocale,       ///< a text-package approval does not name its locale
-    EmptyApprovedPackageId,    ///< a text-package approval does not name its package
-    DuplicateApprovedLocale,   ///< two text-package approvals claim the same locale
+    UnsupportedSchemaVersion,    ///< the package declares a version this module does not read
+    EmptyId,                     ///< the screen has no id, so no directory and no evidence entry
+    NonPositiveSurface,          ///< a surface with no extent cannot contain a rectangle
+    EmptyNodeId,                 ///< a node with no id cannot be named by a golden or a requirement
+    DuplicateNodeId,             ///< two nodes share an id, so a golden could name either
+    DegenerateBounds,            ///< a rectangle with no extent, which `DrawList` refuses to record
+    BoundsOutsideSurface,        ///< a rectangle the declared surface does not contain
+    MalformedColorToken,         ///< a colour that is not a `Theme.Colors.<Token>` name
+    UnknownColorToken,           ///< a well-formed name the governed table does not define
+    UnknownPayload,              ///< a payload this module cannot name, or one left valueless
+    EmptyRequiredName,           ///< a spec field the component dictionary requires is empty
+    NoStates,                    ///< a status indicator that can show nothing
+    StateColorCountMismatch,     ///< per-state tints that do not pair one-to-one with the states
+    NonPositiveMaxLength,        ///< a text input that can hold no character
+    EmptyBudget,                 ///< a screen with nodes whose budget can hold no primitive
+    BudgetExceedsIndexWidth,     ///< more vertices than a 16-bit index can address
+    EmptyApprovedLocale,         ///< a text-package approval does not name its locale
+    EmptyApprovedPackageId,      ///< a text-package approval does not name its package
+    EmptyApprovedPackageDigest,  ///< a text-package approval carries no package identity
+    DuplicateApprovedLocale,     ///< two text-package approvals claim the same locale
+    MissingTextPackageApproval,  ///< a text-bearing screen approves no text package
 };
 
 [[nodiscard]] constexpr std::string_view describe(SchemaError error) noexcept {
@@ -157,8 +159,12 @@ enum class SchemaError : std::uint8_t {
             return "an approved text package does not name its locale";
         case SchemaError::EmptyApprovedPackageId:
             return "an approved text package does not name its package id";
+        case SchemaError::EmptyApprovedPackageDigest:
+            return "an approved text package does not carry its package digest";
         case SchemaError::DuplicateApprovedLocale:
             return "two approved text packages claim the same locale";
+        case SchemaError::MissingTextPackageApproval:
+            return "a text-bearing screen approves no text package";
     }
     return "unknown schema error";
 }
@@ -781,6 +787,16 @@ struct ScreenPackage {
     return requireColor(spec->colorToken);
 }
 
+/// Whether this payload needs the font/locale inputs the text-budget stage approves.
+///
+/// Kept beside `validatePayload()` so a component added to the closed payload set cannot acquire
+/// text-bearing fields without the schema's approval-manifest rule being visible in the same edit.
+[[nodiscard]] constexpr bool needsTextPackageApproval(NodePayload payload) noexcept {
+    return std::holds_alternative<LabelSpec>(payload) || std::holds_alternative<ClockSpec>(payload) || std::holds_alternative<ButtonSpec>(payload)
+           || std::holds_alternative<CriticalButtonSpec>(payload) || std::holds_alternative<NumericDisplaySpec>(payload)
+           || std::holds_alternative<StatusIndicatorSpec>(payload) || std::holds_alternative<TextInputSpec>(payload);
+}
+
 constexpr mdux::core::ResultVoid<SchemaError> ScreenPackage::validate() const noexcept {
     using mdux::core::err;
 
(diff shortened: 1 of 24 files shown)


src/medui/Screen.cpp
```diff
@@ -201,8 +222,7 @@ std::string_view describe(ScreenError error) noexcept {
     return "unknown screen error";
 }
 
-mdux::core::Result<FrameStats, ScreenError> render(const ScreenPackage& screen, mdux::draw::DrawList& list,
-                                                   const TextBinding& text) noexcept {
+mdux::core::Result<FrameStats, ScreenError> render(const ScreenPackage& screen, mdux::draw::DrawList& list, const TextBinding& text) noexcept {
```
ambroise-leclerc, line 227: **1. `TextBinding` is not tied to the screen that approved it — the new gate is bypassable with two screens.** *(verified)*

`TextBinding` stores only `font_`, `text_`, `runs_` (`include/mdux/medui/Screen.cppm:232`) — nothing identifying the `ScreenPackage` that approved it. `render()` then uses whatever binding it is handed, unconditionally.

Concrete scenario: a device holds screens A and B; A approves text package P1, B approves P2. The app calls `TextBinding::create(screenA, font, P1, runs)`, which succeeds, then `render(screenB, list, binding)`. B draws P1's wording — which B's review never approved. That is exactly the substitution ADR-011's amendment claims to close.

The module comment ("decided once… properties of the four artifacts a caller bound together") assumes a one-screen-per-binding discipline the API does not enforce.

**Fix:** store the approving `const ScreenPackage*` (or its `id`) in `TextBinding`, and have `render()` refuse a binding that belongs to another screen.

src/medui/Screen.cpp
```diff
@@ -172,6 +175,22 @@ mdux::core::Result<TextBinding, ScreenError> TextBinding::create(const mdux::fon
         }
     }
 
+    // Package identity is the canonical package.json, not only the fields the runtime happens to
+    // read. Re-serializing the validated owning package gives the same bytes the baker committed and
+    // prevents a caller from supplying an approved digest beside a different in-memory package.
+    const auto canonical = text.write();
```
ambroise-leclerc: **4. Heap allocation inside a `noexcept` device-runtime function — an allocation failure calls `std::terminate` instead of failing closed.** *(verified)*

`text.write()` → `toJson()` builds a whole `json::Value` DOM (nested `std::vector`/`std::string`) plus the serialized `std::string`. Both `TextPackage::write()` and `TextBinding::create()` are `noexcept`, so a `std::bad_alloc` from any of those allocations terminates the process rather than returning `PackageNotApproved`. On a constrained device this turns a recoverable start-up refusal into a hard crash.

It also cuts against ADR-012's stated positive ("a device build … parses nothing at startup") by pulling JSON serialization into the device path.

**Alternatives:** hash the fields directly — streaming `evidence::Digest` updates over locale, id, sidecar digest and run digests — or have the caller supply the package bytes it already loaded.

tools/medui/Compile.cpp
```diff
@@ -454,7 +456,7 @@ loadLocales(const Recipe& recipe, const std::filesystem::path& root, std::vector
         }
         inputs.push_back(fileRecord(sidecarRelative.generic_string(), *sidecar));
 
-        locales.push_back(LoadedLocale{.package = std::move(*package), .sidecar = *sidecar});
+        locales.push_back(LoadedLocale{.package = std::move(*package), .sidecar = *sidecar, .packageSha256 = packageRecord.sha256});
```
ambroise-leclerc, line 474: **2. The compiler records the digest of the text package's *file bytes*; the runtime compares a digest of `write()`'s re-serialization. Nothing enforces that the two coincide.** *(verified)*

Here `loadLocales()` stores `packageRecord.sha256`, i.e. `sha256(raw file bytes)`. At `src/medui/Screen.cpp:185` the runtime computes `sha256(TextPackage::parse(file)->write())`. These agree only if the committed `package.json` is byte-for-byte canonical **and** `parse`→`write` is lossless. `TextPackage::parse()` does not reject unknown members, and `json::write()` re-sorts and reformats.

So a text package that is valid but not byte-canonical — hand-edited whitespace, a different key order, an extra member, or output from an older baker — compiles cleanly, passes `checkLocaleWiring`, `checkTextBudgets` and `readPackage`, and is then **permanently refused on device** with `PackageNotApproved`, with no host-side signal at all. Fail-closed, but silent.

**Fix:** at compile time, assert that `locale.package.write()` reproduces the input bytes exactly, and diagnose the mismatch there where an author can act on it.

tools/medui/Compile.cpp
```diff
@@ -618,6 +620,13 @@ std::optional<CompileOutputs> run(const Recipe&                 recipe,
         textPackages.push_back(locale.package);
     }
 
+    std::vector<ms::TextPackageApproval> approvals;
```
ambroise-leclerc, line 647: **5. A recipe that names text packages for a screen with no text keys silently produces an empty approval manifest.** *(verified — the cause is the `if (measurable)` gate at line 597)*

`loadFont()`/`loadLocales()` run only under `if (measurable)`, so when `needsTextBudget(screen)` is false the recipe's `[text]` packages are never loaded and `locales` is empty — which makes `approvals` here empty too. Meanwhile `report.json`'s `options.textPackages` still lists them, and `inputs` does not.

Any caller that creates bindings uniformly at start-up then gets `PackageNotApproved` for a screen whose own recipe declares the very package it is offering.

**Fix:** either diagnose the unused `[text]` table, or record its approvals regardless of `measurable`.

tools/medui/Package.cpp
```diff
@@ -306,24 +325,34 @@ std::span<const std::string_view> ScreenDocument::internList(std::span<const std
...
+void ScreenDocument::setHeader(std::string_view                         id,
+                               std::int32_t                             surfaceWidth,
+                               std::int32_t                             surfaceHeight,
+                               mdux::draw::DrawBudget                   budget,
+                               std::span<const ms::TextPackageApproval> approvedTextPackages) {
     id_            = intern(id);
     surfaceWidth_  = surfaceWidth;
     surfaceHeight_ = surfaceHeight;
     budget_        = budget;
+    approvedTextPackages_.reserve(approvedTextPackages.size());
```
ambroise-leclerc: **3. `setHeader()` appends approvals instead of replacing them, unlike every other field it sets.** *(verified)*

`id_`, `surfaceWidth_`, `surfaceHeight_` and `budget_` are all assigned; `approvedTextPackages_` is only `reserve()`d and then pushed into. The name and the doc comment ("Records the header") both imply idempotence.

A second `setHeader()` call on the same document therefore leaves the manifest carrying both sets of entries, and `validate()` reports `DuplicateApprovedLocale` for a document whose header was merely re-set.

**Fix:** `approvedTextPackages_.clear();` before the loop.

tools/medui/Package.cpp
```diff
@@ -489,6 +523,25 @@ class Sink {
...
 }
 
+[[nodiscard]] std::optional<mdux::evidence::Digest> readDigest(const json::Value& object, std::string_view key, std::string_view what, const Sink& sink) {
+    const json::Value* member = object.find(key);
+    if (member == nullptr) {
+        sink.fail(memberWrong, std::format("{} is missing the member '{}'", what, key));
+        return std::nullopt;
+    }
+    const auto text = member->asString();
+    if (!text.has_value()) {
```
ambroise-leclerc, line 537: **8. `readDigest()` reports the wrong reason for a non-string member.** *(plausible — cosmetic)*

When `member->asString()` fails — e.g. `"packageSha256": 42` — the message is "is not a lowercase SHA-256 string", which reads as a complaint about the string's *format*. The type error and the 64-hex-digit error below should be distinguishable, as they are for every other reader in this file (`readName` just above says "is not a string").

Minor, but it points a bake failure at the wrong edit.

include/mdux/medui/Schema.cppm
```diff
@@ -766,6 +794,21 @@ constexpr mdux::core::ResultVoid<SchemaError> ScreenPackage::validate() const no
         return err(SchemaError::NonPositiveSurface);
     }
 
+    for (std::size_t index = 0; index < approvedTextPackages.size(); ++index) {
```
ambroise-leclerc, line 813: **6. `validate()` checks the approval's locale and package id but never its digest, so an unset (all-zero) `packageSha256` passes every host gate.** *(verified)*

`SchemaTests.cpp` itself constructs `.packageSha256 = {}` approvals that validate. If a future emitter or reader path drops the digest, the screen validates, `readPackage()` accepts it, the generated `static_assert` passes — and the failure surfaces only on the device, as a permanent `PackageNotApproved` with no diagnostic pointing at the cause.

Given the manifest is a fail-closed safety record, an `EmptyApprovedDigest` check alongside the two new name checks would catch it at bake time.

ambroise-leclerc (commented):
## Review: binding a screen to its approved text packages

Eight findings. Six were verified against the source at `6343449`; the two marked *plausible* are reasoned from the code but not executed.

The headline one is that **the approval gate this PR builds is bypassable across two screens** — `TextBinding` keeps no record of which `ScreenPackage` approved it, and `render()` re-checks nothing. Two others are silent deployment traps: the compiler stores the digest of the package *file bytes* while the runtime compares a digest of `parse()`→`write()`, and the one test that would catch the divergence sits behind the Vulkan skip.

| # | Where | Finding |
|---|-------|---------|
| 1 | `src/medui/Screen.cpp:225` | `TextBinding` is not tied to the screen that approved it |
| 2 | `tools/medui/Compile.cpp:459` | Compiler hashes file bytes; runtime hashes re-serialization |
| 3 | `tools/medui/Package.cpp:337` | `setHeader()` appends approvals instead of replacing them |
| 4 | `src/medui/Screen.cpp:181` | Heap allocation inside a `noexcept` device-runtime `create()` |
| 5 | `tools/medui/Compile.cpp:597` | Recipe `[text]` silently ignored when the screen has no text keys |
| 6 | `include/mdux/medui/Schema.cppm:797` | `validate()` never checks the approval digest is set |
| 7 | `tests/render/ScreenPixelTests.cpp:167` | Digest agreement tested only behind the Vulkan gate |
| 8 | `tools/medui/Package.cpp:533` | `readDigest()` reports a format error for a type error |

Details inline.

tests/render/ScreenPixelTests.cpp
```diff
@@ -159,8 +163,8 @@ struct BoundText {
 
     /// Through `create()`, which is the only way to obtain one - and which proves the three
     /// committed artifacts describe each other before a frame reads any of them.
-    [[nodiscard]] medui::TextBinding binding() const {
-        auto made = medui::TextBinding::create(font, text, runs);
+    [[nodiscard]] medui::TextBinding binding(const medui::ScreenPackage& screen) const {
+        auto made = medui::TextBinding::create(screen, font, text, runs);
```
ambroise-leclerc: **7. The only test that exercises the compile-side and runtime-side digests against each other sits behind the Vulkan gate.** *(plausible)*

`CompileTests.cpp:206` asserts only the compiler side (`approval.packageSha256 == sha256(file)`). `ScreenRuntimeTests.cpp` builds every expectation with `approvalFor()`, i.e. `sha256(write())`, over synthetic packages. Only `BoundText::binding(package)` here binds the *committed* text package to the *committed* screen — and this test is skipped wherever no Vulkan loader/ICD exists (the PR notes it was skipped locally).

The invariant behind finding 2 is therefore unverified on any non-Vulkan host.

**Fix:** a cheap non-Vulkan test asserting

```
sha256(contentsOf("generated/text/endoscope-monitor-en-us/package.json"))
  == sha256(TextPackage::parse(...)->write())
```

tools/medui/Package.cpp
ambroise-leclerc: Resolved in d80a7f5. setHeader now constructs a temporary replacement vector and assigns it, which also handles an input span aliasing existing document storage. Both replacement and alias cases are tested.

src/medui/Screen.cpp
ambroise-leclerc: Resolved in d80a7f5 using the suggested caller-bytes alternative. The runtime hashes the already-loaded canonical package span and no longer serializes JSON. screen.noheap.symbolScan now passes, along with the allocation tests.

include/mdux/medui/Screen.cppm
```diff
@@ -205,28 +198,35 @@ public:
     constexpr TextBinding() noexcept = default;
 
     /**
-     * @brief Proves `font`, `text` and `runs` describe each other, or says which one does not.
+     * @brief Proves the screen approved `text`, and that `font`, `text` and `runs` agree.
      *
      * Hashes the sidecar, so it is linear in its size - once, at start-up, never in a frame.
      *
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in d80a7f5. The API comment now states that create hashes canonical package bytes plus the sidecar, once at startup, and that both hashes are allocation-free.

tools/medui/Compile.cpp
ambroise-leclerc: Resolved in d80a7f5. loadLocales requires package.write to reproduce the input bytes exactly and diagnoses noncanonical input with MEDUI-E002. create hashes those same input bytes. Tests cover noncanonical rejection and committed compiler/runtime digest agreement without Vulkan.

include/mdux/medui/Schema.cppm
```diff
@@ -766,6 +794,21 @@ constexpr mdux::core::ResultVoid<SchemaError> ScreenPackage::validate() const no
...
+        }
+        if (approval.packageId.empty()) {
+            return err(SchemaError::EmptyApprovedPackageId);
+        }
+        for (std::size_t earlier = 0; earlier < index; ++earlier) {
+            if (approvedTextPackages[earlier].locale == approval.locale) {
+                return err(SchemaError::DuplicateApprovedLocale);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in d80a7f5. ScreenPackage::validate now returns MissingTextPackageApproval for text-bearing payloads with an empty manifest, while panel-only and empty screens remain valid. The schema suite covers both sides.

src/medui/Screen.cpp
```diff
@@ -172,6 +175,22 @@ mdux::core::Result<TextBinding, ScreenError> TextBinding::create(const mdux::fon
...
+    if (!canonical.has_value()) {
+        return mdux::core::err(ScreenError::PackageNotApproved);
+    }
+    const mdux::evidence::Digest packageSha256 = mdux::evidence::sha256(std::as_bytes(std::span{canonical->data(), canonical->size()}));
+
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in d80a7f5. create now accepts and hashes already-loaded canonical package bytes, so governed code no longer calls text.write or allocates. The screen no-heap symbol scan and allocation tests pass.

tools/medui/Package.cpp
ambroise-leclerc: Resolved in d80a7f5. readDigest now reports a non-string value as not a string, while malformed string content retains the 64-lowercase-hex diagnostic. The package test covers the JSON type error.

src/medui/Screen.cpp
```diff
@@ -172,6 +175,22 @@ mdux::core::Result<TextBinding, ScreenError> TextBinding::create(const mdux::fon
...
     }
 
+    // Package identity is the canonical package.json, not only the fields the runtime happens to
+    // read. Re-serializing the validated owning package gives the same bytes the baker committed and
+    // prevents a caller from supplying an approved digest beside a different in-memory package.
+    const auto canonical = text.write();
+    if (!canonical.has_value()) {
+        return mdux::core::err(ScreenError::PackageNotApproved);
+    }
+    const mdux::evidence::Digest packageSha256 = mdux::evidence::sha256(std::as_bytes(std::span{canonical->data(), canonical->size()}));
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in d80a7f5 with a stricter host boundary. The compiler now rejects parseable but noncanonical text package JSON with MEDUI-E002, and the runtime hashes the same caller-supplied canonical bytes without serialization. A noncanonical compile regression and a non-Vulkan committed-package binding test cover the invariant.

src/medui/Screen.cpp
```diff
@@ -172,6 +175,22 @@ mdux::core::Result<TextBinding, ScreenError> TextBinding::create(const mdux::fon
...
+
+    const auto approved = std::ranges::find_if(screen.approvedTextPackages, [&](const TextPackageApproval& candidate) {
+        return candidate.locale == text.locale && candidate.packageId == text.header.id && candidate.packageSha256 == packageSha256;
+    });
+    if (approved == screen.approvedTextPackages.end()) {
+        return mdux::core::err(ScreenError::PackageNotApproved);
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in d80a7f5. TextBinding retains locale, package ID, and digest; render checks that identity against its target screen before recording anything. The new two-screen regression proves a fitting binding approved for screen A is PackageNotApproved on screen B.

include/mdux/medui/Schema.cppm
ambroise-leclerc: Resolved in d80a7f5. EmptyApprovedPackageDigest rejects the all-zero digest during constexpr schema validation, and the schema suite now covers it explicitly.

src/medui/Screen.cpp
ambroise-leclerc: Resolved in d80a7f5. The binding retains the approved locale, package ID, and digest, and render rechecks that identity against the supplied screen before recording. The medui-screen-binding-target regression covers two different fitting packages and confirms rollback.

tools/medui/Compile.cpp
ambroise-leclerc: Resolved in d80a7f5 by diagnosing the unused text table. A text-free screen with font or locale inputs now fails with MEDUI-E002 instead of emitting an empty manifest and inconsistent report inputs.

tools/medui/Package.cpp
```diff
@@ -306,24 +325,34 @@ std::span<const std::string_view> ScreenDocument::internList(std::span<const std
...
+                               std::span<const ms::TextPackageApproval> approvedTextPackages) {
     id_            = intern(id);
     surfaceWidth_  = surfaceWidth;
     surfaceHeight_ = surfaceHeight;
     budget_        = budget;
+    approvedTextPackages_.reserve(approvedTextPackages.size());
+    for (const ms::TextPackageApproval& approval : approvedTextPackages) {
+        approvedTextPackages_.push_back(
+            ms::TextPackageApproval{.locale = intern(approval.locale), .packageId = intern(approval.packageId), .packageSha256 = approval.packageSha256});
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in d80a7f5. setHeader builds replacement approvals first and then replaces document storage, preserving alias safety. Package tests cover both a second header and an input span aliased from package().

tests/render/ScreenPixelTests.cpp
ambroise-leclerc: Resolved in d80a7f5. CompileTests now loads the committed font, text package, and runs, then creates a TextBinding against the freshly compiled screen. This exercises compiler/runtime digest agreement on every host without Vulkan.

include/mdux/medui/Screen.cppm
```diff
@@ -205,33 +201,64 @@ public:
...
      * @param text the text package for the locale being run
+     * @param packageJson the canonical package bytes from which `text` was loaded
      * @param runs the sidecar bytes `text` addresses ranges of
      */
-    [[nodiscard]] static mdux::core::Result<TextBinding, ScreenError> create(const mdux::font::FontPackage& font,
+    [[nodiscard]] static mdux::core::Result<TextBinding, ScreenError> create(const ScreenPackage&           screen,
+                                                                             const mdux::font::FontPackage& font,
                                                                              const mdux::text::TextPackage& text,
+                                                                             std::span<const std::byte>     packageJson,
                                                                              std::span<const std::byte>     runs) noexcept;
```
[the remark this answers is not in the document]
ambroise-leclerc: Resolved in b1597ca. TextPackage::canonicalSha256() now streams the exact canonical package representation into the governed SHA-256 implementation without allocating. TextBinding::create() requires that digest to equal the supplied approved JSON digest before retaining the package, so approved bytes for A cannot authenticate parsed text/runs from B. The new split-input regression expects PackageNotApproved; canonical writer parity (ordinary, escaped, and empty-run packages), allocation counters, and the no-heap symbol scan all pass.

AM-L (approved):

PR #247, opened 2026-08-26
Ask whether the Linux Clang leg can run the evidence tests [#246]

### LINKED ISSUE #246: Verify whether the Linux Clang leg can run the evidence tests, and correct what claims it already does
Follow-up to #48, which added the GCC 16 leg and left the Clang half of its title open, and to
#222, which changed what is reachable here.

## What

Establish whether `clang-build.yml` — Linux, Clang, currently `workflow_dispatch` only — can
configure, build and run `ctest --preset ninja-clang -L evidence` green. Then correct the two
places that describe the leg, and re-enable it if the answer is yes.

## Why now

**#222 removed the reason the leg was disabled.** The workflow's own header comment says Clang 20's
`std` module support "is not ready for this codebase", and the `ninja-clang` preset pins
`clang-20`/`clang++-20`. Since #222 the project builds and tests green on **Clang 21.1.8 with
libc++** on every push, on the macOS lane. The blocker recorded in the comment is a *Clang 20*
blocker; nothing in the tree has tested Clang 21 on Linux.

`cmake/toolchains/macos-arm64-llvm.cmake` shows what made it work, and the interesting part is not
macOS-specific:

```cmake
set(CMAKE_CXX_STDLIB_MODULES_JSON
    "${_mdux_llvm_root}/lib/c++/libc++.modules.json" CACHE FILEPATH "" FORCE)
set(CMAKE_CXX_FLAGS_INIT "-stdlib=libc++")
```

Pointing CMake at libc++'s `modules.json` is how `import std` resolves. The Linux job already
installs `libc++-20-dev`/`libc++abi-20-dev` but sets neither variable, and the preset carries no
toolchain file — so it is not obvious the leg ever failed for the reason the comment gives.

## Two documentation defects this uncovered

Both are present-tense claims about mechanisms that do not run. This is the defect class #116 found
in ADR-005 and fixed by rewriting it to describe only what CI executes.

1. **`docs/adr/ADR-007-evidence-pipeline-doctrine.md:127`** — decision 6, the doctrine's own
   determinism claim, reads:

   > The evidence tests run on Windows/MSVC **and** Linux/GCC (and Clang, now that issue #48
   > re-enabled that leg)

   **Issue #48 did not re-enable that leg.** `clang-build.yml` has no `push` or `pull_request`
   trigger, and its header comment says so explicitly. The parenthesis is false, and it sits in the
   ADR paragraph written to stop CI being "simplified" to one leg without knowingly discarding the
   claim.

2. **`.github/workflows/clang-build.yml:15-21`** — the header still lists adding a Clang evidence
   step as future work:

   > When it is turned back on for real, add an evidence step matching the MSVC and GCC legs […]
   > Clang's floating-point code generation differs from both MSVC's and GCC's, so a third
   > independent toolchain strengthens the byte-identity claim in ADR-007 further (decision 6).

   Since #222 the macOS lane runs `ctest --preset ninja-macos-clang -L evidence` on every push. The
   third-toolchain claim that paragraph reserves for the future **is already earned**, by Clang 21
   and libc++ on a different platform. The comment is not wrong about wanting a Linux Clang leg; it
   is wrong that the claim is still outstanding.

## Acceptance

- [ ] A recorded attempt to build `ninja-clang` on Linux with Clang 21 (apt.llvm.org
      `llvm-toolchain-21`), including whether `import std` resolves with an explicit
      `CMAKE_CXX_STDLIB_MODULES_JSON` and `-stdlib=libc++`.
- [ ] `ctest --preset ninja-clang -L evidence` result recorded — green, or the failure named
      precisely enough that the next attempt starts from it.
- [ ] ADR-007 decision 6 states only what CI runs. Whatever the outcome, the false `#48`
      parenthesis goes, and the earned macOS/Clang leg is described accurately.
- [ ] `clang-build.yml`'s header comment matches reality: the Clang 20 rationale updated or
      replaced, and the "when it is turned back on, add an evidence step" paragraph corrected to
      say the third-toolchain claim is already held by the macOS lane.
- [ ] If the leg is green: `push`/`pull_request` triggers added, the preset moved off `clang-20`,
      and an evidence step matching the MSVC and GCC legs. If it is not: the real blocker written
      into the comment in place of the stale one, and the leg left on manual dispatch.

**A red result must not be forced green.** The honest outcome of this issue may be "still blocked,
for this specific reason instead" — which is worth more than the leg, because it is what the two
documents above should have said all along.

ambroise-leclerc:
Closes #246.

## Result: the leg is green, and it caught two things three green legs did not

[Run 33023734904](https://github.com/ambroise-leclerc/MduX/actions/runs/33023734904): **616/616**,
evidence step included. It now runs on every push, and has already verified itself on this PR.

The premise it was disabled on was wrong. `import std` was never the remaining blocker: libc++-21-dev
ships its manifest at `/usr/lib/llvm-21/lib/libc++.modules.json`, the new toolchain file finds it,
and configure succeeds. Two real defects were the blocker, and **neither was reachable on any other
leg**:

### 1. `ShaderPackage::toJson()` exceeded the frame-size guard

```
error: stack frame size (4120) exceeds limit (4096) [-Werror,-Wframe-larger-than]
```

24 bytes. The limit is the determinism guard from #63, and `cmake/MduXDeterminism.cmake` states why
beside it: *"no heap" must not quietly become "enormous stack", which on a device is the same bug
wearing a different hat.* **The limit was left alone and the function shrank instead** — raising it
would retire a device-safety guard to accommodate a CI lane.

The cause was structural, not one large object: all four sections were built inline, so their
`json::Value` temporaries lived in sibling scopes of one frame, and Clang reuses those slots less
aggressively than GCC. Each section is now its own function; `toJson()` holds one at a time and reads
as one line per member of the document.

Behaviour is unchanged *by construction*, and the byte-identity tests are what say so rather than
inspection: `ctest -L evidence` 7/7, 103/103 shader tests, identical emitted `package.json`.

This is the only leg that compiles that function for x86-64 with Clang, which is why arm64 macOS and
GCC never saw it.

### 2. `InstallTreeConsumer` built against the wrong standard library

Failed on `'__config' file not found`. The nested consumer configure already forwarded `AR`,
`RANLIB`, the std module manifest and the macOS sysroot — but not the compiler flags, so it never
received `-stdlib=libc++`. Invisible on macOS, where libc++ *is* Clang's default.

The symptom is a missing header; the defect is worse than that. The consumer could compile against
one standard library while linking a MduX built against another — precisely what this test exists to
catch rather than to reproduce.

## Documentation corrected

- **ADR-007 decision 6** claimed the evidence tests run on Clang "now that issue #48 re-enabled that
  leg". #48 did not — `clang-build.yml` never carried a `push` trigger. That false present-tense
  claim sat in the paragraph written to stop the byte-identity claim being weakened by accident,
  which is the defect class #116 found in ADR-005. The correction is kept visible in the ADR rather
  than silently overwritten.
- It now says **four legs across three operating systems**, and says what the fourth buys, which is
  *not* a fourth toolchain: the same compiler and standard library as the macOS lane, on the same OS
  as the GCC lane. That separates "a different toolchain produced identical bytes" from "a different
  *platform* produced identical bytes" — two claims the doctrine had been making as one.
- The workflow header keeps the obsolete Clang 20 rationale as a visible correction, so the next
  reader sees why it was wrong.

## One consequence worth a reviewer's judgement

**Every PR now carries a fourth lane.** The argument for it is above — it caught two defects on its
first real run. If the CI cost is judged badly spent, removing the triggers is a one-line revert and
the leg reverts to manual dispatch with everything else intact.

## Summary by CodeRabbit

- **New Features**
  - Added automated Linux Clang 21/libc++ validation for pushes and pull requests.
  - Added build configuration for discovering and validating the required LLVM and libc++ tooling.

- **Bug Fixes**
  - Fixed install-tree consumer builds to use the same standard-library configuration as the main build.
  - Addressed issues detected by the expanded Clang validation.

- **Documentation**
  - Updated build preset and evidence-pipeline documentation to reflect automatic Linux Clang coverage and its findings.

- **Refactor**
  - Reorganized shader package JSON generation without changing its output or validation behavior.

### DIFF
--- a/.github/workflows/clang-build.yml
+++ b/.github/workflows/clang-build.yml
@@ -1,25 +1,49 @@
-# Clang leg — manual-dispatch only, not part of the automatic CI set.
+# Clang leg on Linux — manual-dispatch while issue #246 runs.
 #
-# This job is NOT wired to push or pull_request on purpose: Clang 20's `std` module support is
-# not ready for this codebase (issue #48 added the GCC 16 leg but did not re-enable this one, so
-# the Clang half of that story remains open). It is kept as a live workflow rather than a
-# commented-out `.disabled` file so that:
+# This job is NOT wired to push or pull_request yet, and #246 has now established why — the reason
+# is not the one this comment used to give.
+#
+# The old rationale, kept so nobody restores it: "Clang 20's `std` module support is not ready for
+# this codebase", with the preset pinned to clang-20 to match. That is obsolete twice over. #222 made
+# Clang 21.1.8 with libc++ continuously verified on macOS, and run 32945833117 on this workflow
+# showed `import std` resolving fine on Linux too: libc++-21-dev ships its module manifest at
+# /usr/lib/llvm-21/lib/libc++.modules.json, the toolchain file finds it, and CMake configures
+# cleanly. The module story is no longer the blocker.
+#
+# **The actual blocker, measured rather than assumed.** The build fails on exactly one diagnostic:
+#
+#     src/shader/Schema.cpp:322: error: stack frame size (4120) exceeds limit (4096) in
+#     'mdux::shader::ShaderPackage@mdux.shader.schema::toJson() const' [-Werror,-Wframe-larger-than]
+#
+# 24 bytes over, in one function, and nothing else in the tree. That limit is not incidental: it is
+# the determinism guard from issue #63 (cmake/MduXDeterminism.cmake), and its comment states the
+# reason — "no heap" must not quietly become "enormous stack", which on a device is the same bug
+# wearing a different hat. So the two available fixes are not equivalent. Reducing `toJson()`'s frame
+# is real work; raising the number retires a device-safety guard to accommodate a CI leg, and this
+# comment exists to say that is not a trade this project makes silently.
+#
+# Note the target, because it explains why macOS is green and this is not: the macOS lane is arm64
+# and this runner is x86-64. Same compiler version, different register allocation and ABI. The
+# function is a long linear chain of `json::Value` temporaries in sibling scopes, and Clang reuses
+# those stack slots less aggressively than GCC does.
+#
+# Until that is settled, the leg stays manual-dispatch. Run it from the Actions tab; because it is
+# `workflow_dispatch` only, a red result never blocks a PR, which is what makes it usable as an
+# experiment. The evidence step below is live rather than deferred, so the moment the build passes,
+# this workflow answers the remaining half of #246 without further edits.
+#
+# It is kept as a live workflow rather than a commented-out `.disabled` file so that:
 #
 #   - the YAML stays syntax-checked by GitHub instead of rotting silently,
 #   - Dependabot can see and bump the action versions pinned below,
-#   - re-enabling it is adding a trigger, not a rename plus an un-comment pass.
+#   - enabling it is adding a trigger, not a rename plus an un-comment pass.
 #
-# Run it from the Actions tab when checking whether a newer Clang has caught up. Because it is
-# `workflow_dispatch` only, a red result here never blocks a PR.
-#
-# When it is turned back on for real, add an evidence step matching the MSVC and GCC legs in
-# windows-build.yml / linux-gcc16-build.yml:
-#
-#     - name: Verify evidence artifacts (Clang)
-#       run: ctest --preset ninja-clang -L evidence
-#
-# Clang's floating-point code generation differs from both MSVC's and GCC's, so a third
-# independent toolchain strengthens the byte-identity claim in ADR-007 further (decision 6).
+# On the third-toolchain claim in ADR-007 decision 6: an earlier version of this comment reserved it
+# for whenever this leg came back. It is no longer outstanding. Since #222 the macOS lane runs
+# `ctest --preset ninja-macos-clang -L evidence` on every push, so byte-identity already holds across
+# MSVC, GCC and Clang 21/libc++. What a Linux Clang leg would add is that third toolchain on the
+# same OS as the GCC leg — isolating the standard library and code generator from the platform —
+# which is a sharper claim, not the first one.
 
 name: Clang Build (manual)
 
@@ -74,7 +98,7 @@ jobs:
       with:
         cmake-version: '4.3.1'
 
-    - name: Install LLVM/Clang 20
+    - name: Install LLVM/Clang 21 and libc++
       run: |
         set -euo pipefail
         sudo apt-get update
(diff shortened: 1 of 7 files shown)


ambroise-leclerc (commented):
Three issues remain on `f59ca28`:

- **[P1] Keep the libc++ manifest coupled to `MDUX_LLVM_ROOT`.** The existing unresolved CodeRabbit thread is valid: the compiler/tool paths are force-updated when the environment root changes, but the already-cached `CMAKE_CXX_STDLIB_MODULES_JSON` is retained. Reconfiguring an existing build tree can therefore combine LLVM tools from the new root with module metadata from the old root. Please rediscover/invalidate it when the selected root changes, or fail with an explicit fresh-build-tree requirement. https://github.com/ambroise-leclerc/MduX/pull/247#discussion_r3867520212

- **[P2] Make “Setup Clang 21” verify the compiler this job actually uses.** In current green run 33024073242, lines 123–126 print `Ubuntu clang version 18.1.3` twice; only the later CMake configure reports Clang 21.1.8 because the toolchain bypasses `PATH`. Thus the named setup/verification step succeeds while demonstrating the wrong compiler. Invoke the versioned binaries (or set and verify the alternatives explicitly) so the CI evidence is internally consistent.

- **[P2] Finish updating ADR-007’s two-leg wording.** Decision 6 now correctly says four legs, but the unchanged consequences still say “on two toolchains at once” (line 198), “the two-leg CI check” (line 213), and “why both legs exist” (line 217). That leaves the governed evidence doctrine contradicting itself about the control being adopted.

The `ShaderPackage::toJson()` refactor otherwise preserves field order/content, and the current Clang run passed 616/616 plus all 558 evidence/evidence-unit checks. All current PR checks are green.

ambroise-leclerc:
## Review comments addressed

**Fixed — the stale-manifest defect is real, and worse than the symptom.** `CMAKE_CXX_STDLIB_MODULES_JSON` is set `CACHE ... FORCE`, so it is always `DEFINED` on a re-configure and the `if(NOT DEFINED ...)` guard never fired again once a cache existed. Re-configuring a build tree with a different `MDUX_LLVM_ROOT` therefore took the new root's `clang++` and kept the old root's libc++ metadata — two LLVM installations mixed silently, under a pipeline whose whole claim is byte-identical output. The discovered manifest now records the root it came from and is re-discovered when that moves; an operator's explicit `-D` pin has no recorded root and is still left alone.

**Fixed earlier — `file(GLOB_RECURSE)` ordering.** Sorted before selection, and a root holding more than one candidate now warns naming them all rather than picking silently.

**Not applying — the `@brief` suggestion on `src/shader/Schema.cpp:408,425`.** It contradicts this repository's documented convention. `CONTRIBUTING.md:97`:

> **`///` line comments carry no `@brief`.** The tag belongs to `/** */` blocks.

The two new comments are `///` line comments, so they are already in house style; adding `@brief` would make them the only ones in the tree that are not. `mdux-docs-lint` agrees — it checks file-level blocks, not `///` lines.

**Moot — the two `docs/roadmap.md` comments.** That file was split out of this PR into #248 after those were posted, and both points were fixed there: the `clang-build.yml` description is current, and the Evidence row now says four legs. The `#218` MD018 note I left alone deliberately — lines 85 and 99 of that document already open paragraphs with a bare issue number, so it is house style, and CommonMark does not render `#218` as a heading anyway (a heading needs a space after the `#`).

ambroise-leclerc:
All three addressed. Verified on [run 33067337005](https://github.com/ambroise-leclerc/MduX/actions/runs/33067337005): 616/616 and 558 evidence checks, all PR checks green.

**[P1] — already fixed, one commit after the revision you read.** Your review landed on `f59ca28`; `32f929d` preceded your message. The diagnosis was right, and the precise cause is worth recording: `CMAKE_CXX_STDLIB_MODULES_JSON` is set `CACHE ... FORCE`, so it is always `DEFINED` on a re-configure and `if(NOT DEFINED ...)` never fired again once a cache existed. The manifest now records the root it was discovered under and is re-discovered when that moves; an operator's explicit `-D` has no recorded root and is still left alone.

**[P2] Setup step — fixed, and `update-alternatives` removed rather than repaired.** Confirmed exactly as you described: `Ubuntu clang version 18.1.3`, twice, under a passing step. Since nothing in this build consults `PATH` — the toolchain addresses `/usr/lib/llvm-21/bin` directly — making the alternative win would have verified nothing while preserving the illusion. The step now checks the binary the build actually uses and **asserts** the major version instead of printing it, matching the macOS lane's gate. It now reports `Ubuntu clang version 21.1.8`.

**[P2] ADR-007 — fixed, with one deliberate deviation from the literal wording.** All three leftovers are updated (lines 198, 213, 217).

Line 198 now reads **"three toolchains and three operating systems"**, not four. Four legs, three toolchains: Clang 21/libc++ runs on two platforms. That is the distinction decision 6 introduces, and writing "four toolchains" there would have re-created the conflation this branch exists to correct. Line 217's risk is also now "reduced to fewer legs" rather than "one leg", since the failure it mitigates is no longer binary.

Happy to take "four legs" literally in all three if you would rather the consequences mirror the decision's heading word-for-word.

AM-L (approved):

Issue #222, opened 2026-08-21
Support macOS as a verified target platform


ambroise-leclerc:
## Goal

Make macOS an explicitly supported and continuously verified MduX target, initially scoped to **Apple Silicon macOS using upstream LLVM/Clang, libc++, Ninja, and MoltenVK**.

This is a platform-support programme, not a certification or production-readiness claim.

## Proposed initial support contract

- Apple Silicon (`arm64`) only for the first supported configuration.
- Pin an exact verified upstream LLVM/Clang and CMake combination.
- Use libc++ standard-library modules and Ninja.
- Use the LunarG Vulkan SDK / MoltenVK for Vulkan execution over Metal.
- Keep AppleClang, Homebrew GCC, and Intel macOS unsupported until each has its own passing validation lane.

## Preliminary evidence

An exploratory local build on Apple Silicon macOS found:

- AppleClang 21 could not provide the module dependency-scanning/import-`std` configuration required by the project.
- Homebrew GCC 16.1 passed the repository's version gate with CMake 4.1.1, but its installed `std.cc` and `std.compat.cc` were effectively empty, so the synthesized `std` module could not be built.
- Homebrew LLVM 21.1.8 provided `std.cppm` and `libc++.modules.json`. With explicit libc++ discovery and LLVM archiver settings, `MduXCore` and `mdux-screenemit` built successfully.
- The exploratory full build was not completed. It stopped on three Clang warning-as-error findings listed below, and no complete macOS CTest run has yet passed.
- Vulkan execution could not be validated in that environment because no Metal device was visible to MoltenVK.

These observations guide the implementation; they are not acceptance evidence for supported status.

### 1. Define and integrate the macOS toolchain

- [ ] Select and pin the initial supported LLVM/Clang, libc++, CMake, Ninja, Vulkan SDK/MoltenVK, and macOS versions.
- [ ] Add a macOS toolchain file or equivalent stable configuration that discovers the libc++ module manifest without embedding a Homebrew Cellar version path in normal usage.
- [ ] Configure `llvm-ar` and `llvm-ranlib`, or stop forcing response files for the Apple system archiver.
- [ ] Move `CMAKE_CXX_EXTENSIONS OFF` before `project()` so CMake's synthesized standard-library module and MduX consumers use a consistent language dialect.
- [ ] Replace or version-gate the hard-coded experimental `import std` UUID so supported CMake minor versions fail clearly rather than becoming incompatible implicitly.
[...]

Issue #219, opened 2026-08-20
Close ClockFormat and SystemEvent across the compiler and the schema


ambroise-leclerc:
Part of #15. Follow-up from #218, where closing these sets in the schema alone was tried and
withdrawn.

## What

TrustSC closes two of the three named values a compiled node carries, in its governed crate:

- `ClockFormat { TimeSeconds, DateTimeSeconds }` — `HH:MM:SS` and `YYYY-MM-DD HH:MM:SS`
- `SystemEvent { NoOp, TriggerHalt }`

and leaves the third open: a text input's glyph set is `glyph_set_id: &'static str`. MduX carries all
three as names, so two diverge and one is already aligned. Under the parity criterion the two should
close; `charset` should stay a name.

**They also earn it independently.** Because the format is closed, TrustSC's budget stage knows a
`TimeSeconds` clock renders `HH:MM:SS` and measures those eight glyphs against the node's bounds
(`trustsc-ui-dsl-authoring`, the clock branch of the text-budget pass). An open name cannot be
measured, only looked up — which is why #195 needs a product-supplied table for it. And a screen that
can name any system event can name one nothing implements, which is worst discovered on the press of
a critical button.

## Why it could not land in #218

The schema is imported by the compiler, so closing the set there alone leaves the canonical type
unable to represent a screen the implemented pipeline accepts:

- `mdux.tools.medui.semantic` gives `format` and `on_press` the `Identifier` domain — any identifier
  is accepted;
- `mdux.tools.medui.textbudget` (#195) resolves a clock format through the supplied
  `DynamicTextRule` table and reports `MEDUI-E053` when it is absent;
- fixtures across three suites write `HH_MM`, `HH_MM_TZ`, `DIGITS_TO_A`, `TimeFormat.HoursMinutes`
  and `SystemEvent.Stop` — none representable by a two-value enumeration.

## The blocking question this issue has to answer first

**Which diagnostic reports a named value outside a closed set?** Nothing registered fits:
`MEDUI-E071` is specific to `cv_check`, `MEDUI-E053` means dynamic text escaping the baked charset,
and `MEDUI-E033` means a value of the wrong *kind* rather than an unknown member of a known set. The
shared contract lists "unknown … formats, system events" among the things that must fail compilation
but assigns no code. TrustSC uses its own `code::UNREGISTERED`.

So this needs a code from `Compliatory/MedUI`, or an explicit decision to reuse one — and that is the
shared contract's call, not an implementation detail.

[...]

PR #248, opened 2026-08-26
Re-anchor the roadmap to what the tree and the issue tracker actually say

ambroise-leclerc:
Re-anchor the roadmap to what the tree and the issue tracker actually say

The document was anchored to `acbe102` at the v0.6.0 baseline; `develop` is six commits further on.
Every one of the thirteen epics was re-queried rather than carried forward: ten closed, three open
(#16, #17, #19), so the 13/10/3 counts were right and are kept.

What the document did not say:

- **#219 is open against a closed epic.** It is labelled "Part of #15", and #15 closed on 23 August.
  It is not a thirteenth child - #218 tried closing `ClockFormat` and `SystemEvent` in the schema
  alone and withdrew - so it now sits under #15 with the reason it outlives it.
- **#153 is open.** It was listed under #18 with no state marker, under a line reading "All nine
  closed". The nine children are closed; #153 is a follow-up that is not one of them.
- **#222 is closed and nowhere in the document**, though it added a third continuously verified
  platform. It now appears in the Tests and Packaging rows.
- **#244 closed after the document was last written**, and it is what makes the text binding
  authenticated rather than conventional.
- **#246 is new**, from re-reading the CI claims this refresh touched.

A "Since the tag" section records the six post-v0.6.0 commits with the conclusion that matters: none
of them closed an epic, so 10/13 stands and the next tag still waits on #16 or #17.

The test count moves from "438 at v0.5.0, and more since" to 616/616 at `6124bcb`, **sourced from
CI rather than a local run**, and the note under the table says why. A clean local build of the same
commit with the same preset and the same Clang 21.1.8 fails three evidence-unit scenarios - two
SEGFAULT, one failed assertion - on a host running macOS 26.5.2 where the verified lane runs
macOS 15. The generated module's own `static_assert(screen.validate().has_value())` compiles while
the same call fails at run time, so it is a constexpr/runtime divergence rather than a bad screen.
macOS 26 is not the supported configuration, but that divergence is worth knowing about before it
becomes one.

ADR-007 decision 6's correction is described in "Two claims C++ makes better": the byte-identity
claim now rests on three toolchains, and it arrived via macOS rather than the Linux Clang leg the
ADR credited.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,9 +1,14 @@
 # MduX → TrustSC parity roadmap
 
-> Backlog · ambroise-leclerc/MduX · updated 23 August 2026
-> Epic status verified at the v0.6.0 release baseline, `develop` @ `acbe102` · 23 August 2026 ·
-> `#15` closed, all twelve children.
-> The divergence table below was last re-verified on 17 August 2026 and is not re-checked here.
+> Backlog · ambroise-leclerc/MduX · updated 26 August 2026
+> Epic status re-verified against `develop` @ `6124bcb` · 26 August 2026. All thirteen epics were
+> queried on GitHub: ten closed, three open (`#16`, `#17`, `#19`). **No epic changed state since the
+> v0.6.0 baseline** — what moved is post-tag content work and one new platform, recorded under
+> [Since the tag](#since-the-tag).
+> Three open non-epic issues are tracked in their owning sections rather than only on GitHub:
+> `#219` under #15, `#153` under #18, and `#246` under the CI claim it corrects.
+> The divergence table below has its *UI authoring*, *Tests* and *Packaging* rows re-verified on
+> 26 August 2026; the other five date from 17 August 2026 and are not re-checked here.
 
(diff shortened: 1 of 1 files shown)


ambroise-leclerc:
## Review comment addressed

**Fixed — the "Since the tag" accounting was internally inconsistent.** It said six commits and tabulated four PRs, silently dropping #238 and the back-merge, which defeats the section's own purpose of making the tag-to-tree gap readable. It now counts seven from `acbe102` — the baseline this document actually carried, which is the honest boundary since `v0.6.0` is not a clean ancestor of `develop` — and names the two commits that have no row and why.

While fixing it I found a second inconsistency of my own: the Evidence row still said artifacts are byte-compared on "both CI legs" while the Tests and Packaging rows two lines below had already been updated. Both are now current, and since #246 turned out to close rather than stay open, they say **four legs across three operating systems**.

That also returns the open non-epic count to two (#153, #219).

AM-L (approved):

Issue #261, opened 2026-08-28
S6 — Button and CriticalButton with requirement binding


ambroise-leclerc:
Part of #17. **Unconstrained:** #219 closed on 31 August, so `SystemEvent` is already a closed set
of `NoOp` and `TriggerHalt`, and a screen naming anything else fails to compile with `MEDUI-E034`.
The "Why #219 matters here" section below is kept as the reasoning, not as a dependency.

## Work

`Button` and `CriticalButton`, with `requirement:` binding and events.

`CriticalButton` carries a `requirement:` that ties the control to a governance requirement id, which
is what makes a safety-critical action traceable from the screen to the SDF rather than by
convention. A `@safety_critical` node without an explicit `requirement:` is already an error (#196);
this extends that rule to the control that triggers the action.

## Why #219 matters here

#219 closes `SystemEvent` to `{ NoOp, TriggerHalt }`. A screen that can name any system event can
name one nothing implements — worst discovered on the press of a critical button, which is exactly
this component.

## Acceptance

- [ ] A `CriticalButton` without a `requirement:` is a compile error.
- [ ] The bound event resolves against the closed `SystemEvent` set; an unimplemented name is a
      compile error, not a silent no-op at press time.
- [ ] The traceability matrix export from #35 shows the requirement reached from the screen.

Issue #258, opened 2026-08-28
S3 — NumericDisplay and Clock live binding


ambroise-leclerc:
Part of #17. **Unblocked:** #219 closed on 31 August, so `ClockFormat` is a closed set and the
budget stage measures a clock instead of looking it up. The section below is kept because it records
why this issue waited.

## Why #219 comes first

#219 closes `ClockFormat` to `{ TimeSeconds, DateTimeSeconds }`. Until it lands, the format is an
open name, and an open name **cannot be measured** — only looked up. TrustSC's budget stage knows a
`TimeSeconds` clock renders `HH:MM:SS` and measures those eight glyphs against the node's bounds;
MduX cannot, which is why #195 needs a product-supplied table for it.

Building `Clock` first therefore means building on that table and removing it afterwards. Land #219
first and the budget stage measures the clock the way it measures every other text node.

## Work

- `NumericDisplay`: live binding of a numeric value, with the format fixed at compile time so its
  widest rendering is measurable against the node's bounds.
- `Clock`: live binding against the closed `ClockFormat` from #219.

## Acceptance

- [ ] The widest rendering of each format is measured at compile time against the node's bounds and
      refused if it does not fit — the same rule every other text node is held to.
- [ ] Neither component allocates in a frame.
- [ ] A golden reference pins bounds and colour but never the varying value, per #16's rule for
      dynamic kinds.

Issue #256, opened 2026-08-28
S1 — mdux-imagebake (QOI) and the Image component


ambroise-leclerc:
Part of #17. No predecessor — independent of #219, so this is the child that can start immediately.

## Work

- `mdux-imagebake`: a host-only QOI decoder and baker, following ADR-007's pattern.
- `mdux.image.schema`: the committed package, with the sidecar carrying pixels.
- A `sampledRgba` draw path added to the existing fixed-budget vertex recording from #13.
- The `Image` component and its `img("ID")` authoring form in the compiler.

## The scope cut, restated so it is not re-litigated by accident

**QOI, not PNG.** QOI is ~200 lines, lossless and public domain — trivially qualifiable as a
from-scratch decoder. PNG needs a hand-written DEFLATE inflater: ~400 lines of much subtler code,
for a format designers can export out of. Document the PNG→QOI export step instead of writing the
inflater.

## Acceptance

- [ ] The decoder is host-only and never linked into `MduXCore` or `MduX` — the trust-zone link
      check enforces this, so a mistake here is a configure-time failure rather than a review miss.
- [ ] A committed image package under `generated/image/`, byte-compared on all four CI legs.
- [ ] The screen draws it, verified by a pixel test, the way #242 did for `Label`.

Issue #255, opened 2026-08-28
S5 — CI verification across all locales, with a diff image on failure


ambroise-leclerc:
Part of #16. Blocked by #254. Closes the epic.

## Work

Invoke the dedicated host tool directly in CI:

```sh
mdux-verify-ui --screen=generated/screen/<id> --locales=all
```

Run it across every render scope and upload a diff image when a check fails — a failure a reader can
see beats one they have to reconstruct. Do not route verification through `MedicalUiExample` or any
other example application.

The CI invocation exercises the same `MduXVerifyUiLib` path that #254 uses to produce
`verification.json`. The committed JSON remains outcome evidence only; the failure diff is a
driver-dependent measurement for a human and therefore stays outside the screen bundle and outside
the source tree.

## Current-screen integration gate

ADR-014 records that the committed `endoscope-monitor` screen's golden `NumericDisplay` and
`SignalTrace` nodes are currently deferred. #17 is the planned route to render them. The final CI
gate must exercise those committed goldens after the relevant rendering work lands; it must not make
CI green by deleting or replacing them, weakening their checks, accepting empty regions, or choosing
a different screen solely because the current one exposes the missing implementation.

## Acceptance

- [ ] `mdux-verify-ui --screen=... --locales=all` runs on lavapipe on the Linux render legs and
      through MoltenVK on macOS on every automatic CI event selected by those workflows.
- [ ] Every CI leg that re-derives `verification.json` actually executes the verification producer;
      a missing loader/device is an infrastructure failure, never a skipped or copied evidence run.
- [ ] Every golden obligation and mandatory text obligation is discharged; a skip is a failure.
- [ ] The committed `endoscope-monitor` golden nodes are actually drawn and checked; #17's missing
      rendering is resolved rather than hidden by changing the obligation set or test subject.
- [ ] A valid textless screen with goldens receives one locale-free render scope rather than zero
      work.
- [ ] `--no-tests=error`-equivalent guarding: a run over a screen that enumerates zero combined
      obligations fails.
- [ ] The diff image is attached to the run as a CI artifact, not written into
      `generated/screen/<id>/` and not only printed.
- [ ] No example target is used as the verification entry point or gains a verification-only Vulkan
      dependency.
- [ ] ADR-007 decision 6's leg count is updated if the verified matrix changes, and the ADR's
      consequences are updated with it — that paragraph has already drifted from its own decision
      once (#246).

PR #262, opened 2026-08-28
Open Wave 6: promote #16 and #17's children, and stop calling them blocked

ambroise-leclerc:
Open Wave 6: promote #16 and #17's children, and stop calling them blocked

#15 closed on 23 August at 12/12, which unblocked both Wave 6 epics. Neither noticed: #16 still
carried a "Current status (2026-08-15) — Blocked on #15" block and #17 a bare "Blocked by: #15",
five days after the fact, and this document repeated both.

The eleven children are now individual issues (#251-#255, #256-#261), which is this document's own
stated convention for an unblocked epic rather than a new one.

Two orderings are worth having written down, because neither is visible from the epic titles:

- **#219 is a prerequisite for #17's Clock and Button work**, not a nicety. It closes `ClockFormat`
  and `SystemEvent`. An open format name cannot be measured, only looked up, so #258 would be built
  against the product-supplied table #195 needs today and then have it removed. And a screen that
  can name any system event can name one nothing implements — worst discovered on the press of the
  critical button #261 builds.
- **#16's bounds and colour-hash checks wait on nothing.** The epic's own sequencing note calls them
  fully exercisable at the solid-rect slice, which shipped in v0.6.0, and #242's Label has since
  given ink containment real ink to check. The epic reads as gated on content it no longer needs.

#256 is the one to start in #17: it depends on nothing else in either epic.

## Summary by CodeRabbit

* **Documentation**
  * Refreshed the product roadmap through 29 August 2026.
  * Updated epic statuses, sequencing, prerequisites, and promoted roadmap items.
  * Added current tracking details for open work and standalone items.
  * Updated test verification information, including macOS CI references and successful Linux/Clang status.
  * Revised roadmap footer metadata to reflect the latest project status.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -160,8 +160,9 @@ lane that strengthens the byte-identity claim rather than moving any epic. Nothi
 
 ### Thirteen epics
 
-Child issues exist for the six actionable epics. The rest keep their breakdown as a
-checklist in the epic body, promoted when unblocked.
+Child issues exist for every actionable epic. #16's and #17's were promoted on 28 August 2026, when
+#15's closure unblocked them; the convention is that a blocked epic keeps its breakdown as a
+checklist in its own body until then.
 
 ---
 
@@ -370,30 +371,46 @@ table for a format that could have been measured.
 
 _Blocks #16, #17_
 
-#### #16 — Rendered-truth verification · **Blocked #13, #15**
+#### #16 — Rendered-truth verification · **Open, unblocked**
 
 Render offscreen, then check that critical content appears where the compiled screen says
 it will, in the declared tint, in every approved locale — and emit that as evidence. Bounds
 and colour checks are exercisable before a single glyph exists.
 
-- S1 ADR: automated UI verification
-- S2 Bounds, ink containment, colour hash
-- S3 The verify driver
-- S4 Evidence report emission
-- S5 CI across all locales
+- #251 S1 ADR: automated UI verification
+- #252 S2 Bounds, ink containment, colour hash
+- #253 S3 The verify driver
+- #254 S4 Evidence report emission
+- #255 S5 CI across all locales
 
-#### #17 — Content components · **Blocked #15**
+Sequential: each child is blocked by its predecessor. Two things landed after the epic was written
+that make it cheaper than it reads — #242 draws a `Label`, so ink containment has real ink to check
+rather than waiting on #17, and #244 makes "every approved locale" a property of the screen's own
+manifest rather than a caller-supplied list. **#252's bounds and colour-hash checks wait on
+nothing**: the epic's own note calls them fully exercisable at the solid-rect slice, which shipped
+in v0.6.0.
+
+#### #17 — Content components · **Open, unblocked**
 
 The rest of the component dictionary. Two deliberate scope cuts: QOI rather than PNG in v1,
 and no IME — input-method editing is a platform concern that does not belong inside a
 governed renderer.
 
-- S1 Image baker and the `Image` component
-- S2 `SignalTrace` — shares the demonstrator's sample ring
-- S3 `NumericDisplay` and `Clock`
-- S4 `StatusIndicator` — has a waiting consumer in the ECG demonstrator
-- S5 `TextInput` (display and caret only)
-- S6 Buttons with requirement binding
+- #256 S1 Image baker and the `Image` component
+- #257 S2 `SignalTrace` — shares the demonstrator's sample ring
+- #258 S3 `NumericDisplay` and `Clock` — **blocked by #219**
+- #259 S4 `StatusIndicator` — has a waiting consumer in the ECG demonstrator
+- #260 S5 `TextInput` (display and caret only)
+- #261 S6 Buttons with requirement binding — touches #219
+
+Largely independent of one another, unlike #16's. **#256 is the one to start**, depending on nothing
+else here.
+
+**#219 orders two of them, and is a prerequisite rather than a nicety.** It closes `ClockFormat` and
+`SystemEvent`. An open format name cannot be *measured*, only looked up, so #258's `Clock` would be
+built against the product-supplied table #195 needs today and then have it removed; and a screen
+that can name any system event can name one nothing implements, which is worst discovered on the
+press of the critical button #261 builds.
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/roadmap.md
```diff
@@ -160,8 +160,9 @@ lane that strengthens the byte-identity claim rather than moving any epic. Nothi
 
 ### Thirteen epics
 
-Child issues exist for the six actionable epics. The rest keep their breakdown as a
-checklist in the epic body, promoted when unblocked.
+Child issues exist for every actionable epic. #16's and #17's were promoted on 28 August 2026, when
```
ambroise-leclerc: The universal claim is not true for the current backlog. Epic #19 is open and still carries S4-S6 as unnumbered checklist entries, while the dependencies those items follow (#15 and #18) are closed. That makes #19 actionable under the convention stated here, but it has no promoted child issues. Please either promote those remaining items as well or narrow this sentence to say that child issues now exist for both Wave 6 epics. Otherwise the roadmap establishes a convention and contradicts it later in the same document.

docs/roadmap.md
```diff
@@ -370,30 +371,46 @@ table for a format that could have been measured.
...
+- #252 S2 Bounds, ink containment, colour hash
+- #253 S3 The verify driver
+- #254 S4 Evidence report emission
+- #255 S5 CI across all locales
 
-#### #17 — Content components · **Blocked #15**
+Sequential: each child is blocked by its predecessor. Two things landed after the epic was written
+that make it cheaper than it reads — #242 draws a `Label`, so ink containment has real ink to check
+rather than waiting on #17, and #244 makes "every approved locale" a property of the screen's own
+manifest rather than a caller-supplied list. **#252's bounds and colour-hash checks wait on
```
ambroise-leclerc: This conflicts with both the preceding sequencing sentence and issue #252. The issue is explicitly blocked by #251 because the ADR defines the derive-dont-trust rule that `mdux.verify` must implement. The geometric and colour checks need no additional renderer/content work, but they still must wait for that predecessor. Please qualify this as "wait on no additional content" or "can begin immediately after #251" so the roadmap does not invite implementation before its governing decision.

ambroise-leclerc:
One additional documentation-consistency finding cannot be attached inline because the affected header is outside this diff:

**Stale document update date — `docs/roadmap.md:3`.** The header still says the roadmap was updated on 26 August 2026, while this PR adds status and promotion records explicitly dated 28 August 2026. That makes the document-level date false as soon as this change merges. Please advance the general `updated` date to 28 August. The separately qualified statements that epic state and selected tables were re-verified on 26 August can remain dated 26 August; they describe narrower baselines and are not contradicted by this PR.

ambroise-leclerc:
Both addressed.

**`docs/roadmap.md:163` — promoted rather than narrowed.** You were right that the sentence was
false, and I took the first of your two options: #19's S4–S6 are now #263, #264 and #265, and the
epic body is updated. The convention this document states is the right one; it was the backlog that
had drifted from it, so weakening the sentence would have recorded the drift instead of fixing it.

They had been actionable for a while — S4–S6 followed #15 and #18, closed on 23 August and 3 August
— which is the same staleness this PR set out to fix in #16 and #17, one epic further along than I
looked. The roadmap's #19 section now carries the numbers too, and notes that the three are
independent of one another, with #264 the cheapest since it needs no compiler work.

**`docs/roadmap.md:389` — my overstatement, corrected.** "Wait on nothing" contradicted #252, which
I had myself blocked by #251. The distinction I was reaching for is real but I collapsed it: the
bounds and colour-hash checks need no *further content* — they are exercisable against the
solid-rect slice that shipped in v0.6.0 — but they still follow the ADR that fixes the
derive-don't-trust rule they implement. It now reads:

> Content is not the constraint; the governing decision is.

which keeps the useful half of the claim without inviting implementation ahead of #251.

---

Unrelated to this PR, but it came out of acting on it: **#219 turned out to be blocked upstream**,
which changes #17's critical path. Its own blocking question — which diagnostic reports a named value
outside a closed set — has no answer in the pinned contract (`d5136a8`). `MEDUI-E033` is the near
miss and is wrong: it means a value of the wrong *kind*, while `format: HH_MM` is a well-formed
identifier that is simply not a member. Allocating a code needs a PR to `Compliatory/MedUI` and a
re-pin here, so I stopped rather than inventing one. Full analysis on #219.

That blocks #258 and orders #261.

ambroise-leclerc:
Both CodeRabbit comments addressed — one fixed and widened, one declined with evidence.

## Fixed, and it was broader than reported

**`docs/roadmap.md:167` — the header date was stale.** Correct, and the same block also carried a
stale baseline commit: `6124bcb`, when `develop` has been at `bdf539c` since #250 merged.

I did not simply advance the date. That header is a *verification claim* — "epic status re-verified
against `develop` @ X on date Y" — so restamping it without redoing the query would have
manufactured exactly the kind of unbacked present-tense assertion this document has been corrected
for twice already (#246, and ADR-005 before it). So I re-queried all thirteen epics: ten closed,
three open (#16, #17, #19), unchanged.

Two consequences fell out of that:

- **The "two open non-epic issues" line had become wrong**, not merely stale. Sixteen non-epic
  issues are open now: fourteen are the children promoted on 28 August, two are standalone (#153,
  #219). The metrics table splits them, since collapsing fourteen tracked children into "outside an
  epic" reads as backlog nobody owns.
- **The test count's anchor moved with it**, so I re-checked rather than copying it forward:
  616/616 at `bdf539c` too, run 33149254738.

## Declined, with the renderer's own answer

**`docs/roadmap.md:164` — "`#16's` starts a line, Markdown treats this as an ATX heading."** That
premise is false, so I have not applied the fix.

CommonMark and GFM require a space after `#` for an ATX heading. Rather than rely on my reading of
the spec, I asked GitHub's own renderer:

```
POST /markdown  →
<p>#16's and #17's children were promoted.</p>
<p>#218 tried closing the two sets.</p>
<div class="markdown-heading"><h1 ...>A real heading</h1>...
```

Both issue references render as paragraph text; only `# A real heading` becomes a heading. What
fires here is markdownlint's MD018, a heuristic for a *suspected typo* — and markdownlint does not
run in this repository's Documentation Lint job.

There is also a consistency argument. Lines 85 and 99 of this document already open sentences with a
bare issue number, and both predate this PR. Rewriting only my line would leave the file
inconsistent in order to satisfy a warning about a rendering that does not happen.

AM-L (approved):

PR #269, opened 2026-08-29
Bump codeql-action to 4.37.9 in one move, and group it so it stays that way

ambroise-leclerc:
Bump codeql-action to 4.37.9 in one move, and group it so it stays that way

Dependabot opened #266 and #268 for the two codeql-action sub-actions in codeql.yml. Both are red on
their own, and neither is wrong: the analyze step rejects a configuration written by a different
version.

    ##[error]Loaded a configuration file for version '4.37.7', but running version '4.37.9'

Each PR bumps one sub-action and leaves the other behind, so each produces exactly that mismatch.
They are only green together, which no individual Dependabot PR can be.

This bumps both refs in one commit, and adds a `groups:` entry so future codeql-action updates
arrive as one PR. Without it this recurs on every CodeQL release, and the failure looks like a
CodeQL problem rather than a split update.

The target SHA was verified upstream rather than trusted from the PR title: cdf488f is tagged
v4.37.9 in github/codeql-action, and its commit is that release's merge. Pinning by SHA only helps
if the SHA is the right one.

#267 is untouched by this - it bumps upload-sarif in scorecard.yml, is green on its own, and merges
independently.

## Summary by CodeRabbit

* **Chores**
  * Updated security scanning tools to a newer version.
  * Grouped related security tool updates so they can be delivered together.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,11 +8,18 @@ updates:
     schedule:
       interval: "weekly"
     groups:
-      # `github/codeql-action/init` and `.../analyze` must run at the same version: the analyze step
-      # rejects a configuration written by a different one with "Loaded a configuration file for
-      # version X, but running version Y". Ungrouped, Dependabot opens one PR per sub-action and
-      # each is red on its own, because each leaves the other behind. Grouping them makes the unit
-      # of update the unit that has to move together.
+      # One action, one version, everywhere it is used.
+      #
+      # `init` and `.../analyze` must match or the build breaks outright: analyze rejects a
+      # configuration written by a different version with "Loaded a configuration file for version
+      # X, but running version Y". Ungrouped, Dependabot opens one PR per sub-action, and each is
+      # red on its own because each leaves the other behind.
+      #
+      # The pattern deliberately covers `upload-sarif` in scorecard.yml too, which is a different
+      # workflow and cannot break that way. Letting it drift is still not worth it: #267 merged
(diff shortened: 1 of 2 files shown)


ambroise-leclerc:
Good catch, and the underlying observation is exactly right: **the pattern covered more than the
comment justified.** `github/codeql-action*` does match `upload-sarif` in `scorecard.yml`, while the
comment beside it only explained the `init`/`analyze` coupling.

I widened the comment rather than narrowing the pattern, because checking the tree turned up an
argument for the broader scope that neither of us had:

```
codeql.yml:58    init@ff2f1c6         # v4.37.7
codeql.yml:65    analyze@ff2f1c6      # v4.37.7
scorecard.yml:46 upload-sarif@cdf488f # v4.37.9
```

**That skew is live on `develop` right now.** #267 merged ahead of this PR, so the repository is
currently running one action at two versions — three parts of a single release moved separately,
which is the situation grouping exists to prevent.

You are right that `upload-sarif` cannot break the way `analyze` does; nothing reads a configuration
it wrote. But the reason to group it is not breakage, it is that one action carrying two versions
across a repository has no upside, and this PR is the proof that they drift apart when nothing holds
them together. So the invariant is now stated as "one action, one version, everywhere it is used",
with the `init`/`analyze` hard requirement given as the strongest case rather than the only one.

If you would still rather keep the two workflows independent, narrowing to an explicit two-entry
list is a one-line change and I will make it — the trade is fewer coupled PRs against a version skew
nobody is watching for.

AM-L (approved):

Issue #253, opened 2026-08-28
S3 — mdux-verify-ui across every approved locale


ambroise-leclerc:
Part of #16. Follows #252, which is closed; this issue is unblocked.

## Placement and command

Create a dedicated host-only verification tool:

```sh
mdux-verify-ui --screen=generated/screen/<id> --locales=all
```

The implementation lives under `tools/verify/` and follows the repository's existing host-tool
shape:

- `MduXVerifyUiLib` owns artifact loading, obligation enumeration, offscreen rendering and outcome
  production, so tests call it directly;
- `mdux-verify-ui` is a thin executable that parses the command line, calls the library and formats
  diagnostics through the shared tool envelope.

The library and executable live outside the governed zone, may perform file I/O and call the Vulkan
adapter, and are not installed or exported. They call the governed `mdux.verify` checks; they are
never linked into `MduXCore` or another governed target.

Do not put this mode in `MedicalUiExample` or another example. `MedicalUiExample` remains the
no-device/no-window frame-construction demonstration and receives neither a `--verify-ui` option nor
a Vulkan-device dependency.

`--screen` names the committed `generated/screen/<id>/` bundle. It is a location, not an expectation
override: the tool reads `package.json` and `goldens.json` there and resolves the governed theme,
font and approved per-locale text artifacts from their recorded identities and digests.

## Work

- Derive every production expectation view from the committed screen package, goldens sidecar,
  governed theme table and approved font/text artifacts rather than caller-supplied values.
- Enumerate the complete obligation set before rendering:
  - every golden × each of its `cvChecks` × every render scope;
  - every compiled `textKey` node × `InkContainment`/`LocalizedTextPresence` × every approved locale.
- Render once per approved locale and run all applicable checks against that frame in memory.
- For a valid textless screen, create one explicit locale-free render scope so `Bounds` and
  `ColorHash` obligations still run.
- `--locales=all` means the screen's exact approved set from its own manifest. The caller cannot
  narrow the set, and a locale the screen was not compiled against is an error rather than a skip.
- Preserve ADR-014's distinction between a completed check that failed and a run that could not be
  made. No unsupported check, unreadable artifact, digest mismatch, missing locale, loader or device
  can become a pass.

## Acceptance

- [ ] `MduXVerifyUiLib` and `mdux-verify-ui` are registered under `tools/verify/`; no example gains
      ownership of the driver or a `--verify-ui` mode.
[...]

Issue #251, opened 2026-08-28
S1 — ADR: automated UI verification


ambroise-leclerc:
Part of #16. No predecessor — this is the epic's first child.

## Work

Fix what rendered-truth verification is, and what it is not, before any of it is written. The
decisions this ADR has to make:

- **The checks are pure functions over a CPU framebuffer.** Golden checks receive an
  artifact-derived golden expectation; mandatory text checks receive an artifact-derived text
  expectation carrying the approved locale and bound run/atlas data. No GPU is needed to unit-test
  either on synthetic inputs.
- **The expected golden set is derived once, in the baker.** `collectGoldens()` is ADR-011's single
  predicate implementation. The verifier consumes `goldens.json` and must not re-apply a predicate
  whose inputs are absent from `package.json`; CI re-runs the baker and byte-compares the committed
  sidecar on all four legs.
- **Every obligation is discharged.** Golden obligations and mandatory text obligations are
  separate families; text nodes do not need goldens, and textless screens get a locale-free render
  scope rather than an empty Cartesian product.
- **What the evidence artifact claims.** Each `verification.json` outcome makes only the claim for
  the check that actually ran. A `Bounds`-only node does not become a tint claim.
- **One screen artifact bundle.** `mdux_compile_screen()` remains the sole screen bake registration;
  `verification.json` and its digest extend the existing bundle and `report.json`.
- **The relationship to IEC 62304:2006 §5.7 Software system testing.** This is project intent for a
  future mechanism, not an implemented risk reduction, certified control or complete system test.

## Acceptance

- [ ] A numbered ADR in `docs/adr/`, contiguous with the existing series.
- [ ] It describes only mechanisms that will exist when the epic closes, in language that does not
      claim they run yet — see #249 and the `mdux-named-mechanisms` check.
- [ ] The derive-don't-trust rule has one named implementation and a named owner for every mismatch
      direction.
- [ ] The golden and mandatory-text obligation families, including the textless render scope, are
      unambiguous.
- [ ] The artifact registration/report ownership and outcome-specific claim wording are explicit.

PR #273, opened 2026-09-01
Check a rendered frame against the screen that produced it [#252]

### LINKED ISSUE #252: S2 — mdux.verify: GoldenBounds, InkContainment, ColorHash, localized-text presence
Part of #16. Blocked by #251.

## Work

`mdux.verify`, a governed module of pure functions over a CPU framebuffer:

- **`GoldenBounds`** — the golden node's rendered content occupies the declared rectangle.
- **`ColorHash`** — the golden node's rendered content carries the tint its token resolves to.
- **`InkContainment`** — a compiled text node's run stays inside its node bounds.
- **`LocalizedTextPresence`** — the approved locale's bound run is the one present on screen.

There are two typed, read-only expectation views:

- a golden expectation: golden entry + named compiled node + resolved tint when needed;
- a text expectation: compiled `textKey` node + approved locale + bound text run/atlas data +
  resolved tint.

The host driver derives production expectations from committed artifacts. Unit tests construct
synthetic views. Neither view carries Vulkan types, allocates, throws or opens files.

`Bounds` and `ColorHash` remain the closed shared `CvCheck` set. `InkContainment` and
`LocalizedTextPresence` are mandatory for every compiled node carrying a `textKey`; do not add them
to `CvCheck`, and do not require such a node to have a golden.

The verifier consumes `goldens.json` and never re-applies ADR-011's predicate. `collectGoldens()` is
the single predicate implementation; its fixtures detect a wrong predicate, and the screen evidence
test re-derives a missing committed golden. This module fails closed when a golden id does not
resolve or when its duplicated bounds/text/color names disagree with the package node.

## Acceptance

- [ ] Each check is a pure function over a framebuffer and the appropriate expectation view,
      unit-tested on synthetic images with no GPU.
- [ ] Tests cover a dangling golden id and every duplicated golden field disagreeing with its named
      package node.
- [ ] Tests prove an unpositioned text node with no golden still receives both mandatory text
      checks.
- [ ] Tests prove a textless golden can be checked in a locale-free render scope.
- [ ] Governed-zone clean: no platform headers, no throwing — `mdux-governed-lint` and
      `governed.noThrow.symbolScan` both pass.

ambroise-leclerc:
## Summary

ADR-014 (#251) fixed what rendered-truth verification checks and what it cannot, and recorded that
nothing it described ran yet. This is the first of its four implementation children: `mdux.verify`,
a governed module of four pure functions over a CPU framebuffer and an artifact-derived expectation.

- **`goldenBounds()`** — the content inside a golden's declared rectangle *is* that rectangle. An
  equality rather than a containment, because that is ADR-014's own worked example: a rectangle
  moved by one pixel still overlaps its declared box, so only the equality fails on it.
- **`colorHash()`** — every painted pixel is a blend of the ground and the resolved tint, and at
  least one is the tint exactly. No tolerance is needed: a convex combination of two integer
  channels cannot round outside the interval they span, and a fully covered pixel is exactly the
  tint because coverage modulates alpha and never rgb.
- **`inkContainment()`** — the ink the committed run and font metrics predict, placed by the rule
  `mdux.medui.screen` fixes, is inside the node, and is what the frame actually shows there.
- **`localizedTextPresence()`** — per glyph, in both directions: every non-blank record of the
  approved locale's run painted something inside its own rectangle, and every pixel of the node
  outside all of those rectangles is still the ground. The second half is what stops a different
  translation from satisfying the first.

Two expectation views, obtainable only through a `create()` that resolves them against the compiled
screen — which is where ADR-014 decision 2 lives. The golden view fails closed on a dangling node id
and on **every** field the sidecar duplicates disagreeing with the node it names, and on a
`ColorHash` with no tint to compare against. The text view needs no golden entry and no
`position:`: it is raised from the `textKey` alone, in one approved locale.

### Three decisions worth a reviewer's attention

**Both views carry two resolved colours, where ADR-014's prose names one.** The tint is what a check
compares *against*; deciding whether a pixel was painted **at all** needs the value it would have
had otherwise, and no amount of knowing the tint supplies it. Both are resolved through the governed
table and the driver's own fixed clear colour — never read back out of the frame under test, which
would make a check compare a measurement against itself. Documented at length in the module header.

**`GoldenBounds` is an equality, so a component that fills only part of its node does not discharge
it.** Stated in the header rather than discovered later: this matters for #17, and the rule to apply
is that such a component needs a golden that says so, not that the check is loosened.

**Two rules that had one implementation keep having one.** `quantise()` moves out of
`Screen.cpp`'s anonymous namespace into `mdux.medui.schema`, because the runtime turns a token
into the bytes a frame carries and this module compares those bytes back — two copies would agree
until one rounded differently, and the failure would read as a rendering defect. `CvCheck` is
defined in `mdux.verify` and aliased by `mdux.tools.medui.goldens`, in that direction because
ADR-004 does not let a device-side reader depend on a host tool.

**A run that paints no ink is refused, not passed.** A single space is legitimate for the runtime,
which draws nothing; it is not something a rendered-truth obligation can discharge, and a pass would
be indistinguishable from a screen that lost its text.

ADR-014 itself is not amended. Its "nothing described below runs today" is a statement of what was
true when it was accepted, and its Review Date clause already points at #253.

Closes #252

Invitation / CLA issue: #252

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked (#251 merged as #272)
- **Merge order:** mergeable now. #253 is blocked by this.

## Shared registries touched

- [x] `CMakeLists.txt` (root) — targets, trust-zone declarations, install/export set
- [ ] None of the above

`tools/medui/Goldens.cppm` and `.cpp` are edited but no host-tool target's source list changes.

## Rebase state

- **Rebased onto:** `bba93abbe946a128742ce869e230b87609fc75c0` (`origin/develop`)

## Verification

Local leg is GCC 16.1.0 / CMake 4.2.3 / Ninja 1.13.2, Release, Vulkan 1.4.341 with lavapipe.

- [x] `cmake -S . -B build-252 -G Ninja -DCMAKE_BUILD_TYPE=Release && cmake --build build-252` — clean, no warnings
- [x] `ctest --test-dir build-252` — **647/647 passed**, of which 18 are the new `verify_spec`
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — OK, 83 files, 0 findings
- [x] Other: `python3 tools/governed-lint/mdux_governed_lint.py` — OK, 34 files, 0 findings;
      `ctest -L governed` — `governed.noThrow.symbolScan` and its negative both pass;
      `offscreen_tests` (the `pixel` label, under lavapipe) — 36/36, including the authored-screen
      panel and label scenarios that exercise the moved `quantise()`;
      `clang-format` clean on every added and edited file.
- Not run locally: MSVC, macOS/Clang 21 and Linux/Clang 21 legs — CI covers them.

## Post-merge gate

      dependent PR may merge.
- **Post-merge `develop` run:** pending — #253 is the successor and must wait for it.
- [ ] That run is green.

## Regulatory impact

This is the mechanism an IEC 62304:2006 §5.7 argument would eventually point at, and it changes no
claim today. Nothing here supplies a software system, system requirements or a representative use
environment, and ADR-014's scope limit is unchanged and restated in the module header: verification
compares a frame against the compiled screen, both come from one source, so what is established is
internal consistency and not that the screen is the right screen.

Artifacts touched: `docs/architecture.md` and `AGENTS.md` module and test-target tables (the
latter's executable counts were already stale by six and are corrected to the tree). No ADR is
amended, no committed artifact under `generated/` changes, and no risk control or traceability
record is claimed. `mdux.medui.screen`'s behaviour is unchanged: `quantise()` moved between
namespaces with its arithmetic intact, and the lavapipe pixel scenarios that compare its output byte
for byte still pass.

## Summary by CodeRabbit

- **New Features**
  - Added rendered-output verification for framebuffer bounds, colors, text placement, and localized text presence.
  - Added validation for malformed or inconsistent rendering expectations and framebuffer data.
  - Centralized color conversion for consistent RGBA rendering.

- **Bug Fixes**
  - Improved detection of misplaced, missing, incorrectly colored, or substituted text and graphics during verification.
  - Prevented invalid rendering metadata from being accepted as valid test expectations.

### DIFF
--- a/include/mdux/medui/Schema.cppm
+++ b/include/mdux/medui/Schema.cppm
@@ -318,9 +318,20 @@ inline constexpr std::array<ThemeColor, 8> themeColors{
  * against the same token. Two copies of this arithmetic would agree until the day one of them
  * rounded differently, and the failure would read as a rendering defect rather than as the drift it
  * was.
+ *
+ * **NaN is mapped to zero rather than clamped.** A NaN compares false against every bound, so the
+ * two clamps below would both let it through and the conversion to `std::uint8_t` would be undefined
+ * - which is a poor way for a governed function to answer a question. `!(channel >= 0.0F)` is the
+ * one test that catches it and negatives together. Zero because a channel this function cannot
+ * interpret contributes nothing, and because the alternative is an error path on a function whose
+ * only production input is a `constexpr` table of literals: the value is unreachable through
+ * `resolveColorToken()`, and being total costs one comparison.
  */
 [[nodiscard]] constexpr std::uint8_t quantise(float channel) noexcept {
-    const float clamped = channel < 0.0F ? 0.0F : (channel > 1.0F ? 1.0F : channel);
+    if (!(channel >= 0.0F)) {
+        return 0;
+    }
+    const float clamped = channel > 1.0F ? 1.0F : channel;
     const float scaled  = clamped * 255.0F;
     const float rounded = scaled + 0.5F;
     return static_cast<std::uint8_t>(rounded);
--- a/include/mdux/medui/Schema.cppm
+++ b/include/mdux/medui/Schema.cppm
@@ -84,6 +84,7 @@ export module mdux.medui.schema;
 
 import std;
 import mdux.core.result;
+import mdux.core.units;
 import mdux.draw;
 import mdux.evidence.digest;
 import mdux.evidence.report;
@@ -297,6 +298,37 @@ inline constexpr std::array<ThemeColor, 8> themeColors{
         }
     }
     return mdux::core::err(ThemeError::UnknownToken);
+}
+
+/**
+ * @brief A linear channel as the byte an `R8G8B8A8_UNORM` target carries.
+ *
+ * Quantisation and nothing else: the governed table stores linear RGBA, a vertex colour is read by
+ * the shader as a plain 0..1 UNORM value, so the byte is the linear value scaled. No transfer
+ * function is applied here, because whether the *swapchain* is sRGB is the renderer's decision and
+ * applying one in two places is how a colour ends up encoded twice.
+ *
+ * The multiply and the add are separate statements so that neither the rounding nor the result
+ * depends on whether the compiler fuses them. `mdux_enforce_fp_determinism(MduXCore)` already turns
+ * contraction off for this target, and this is the belt to that pair of braces: a frame that is
+ * byte-compared across toolchains cannot afford a last-bit difference in a colour.
+ *
+ * Exported rather than kept private to the runtime, and that move is #252's rather than tidiness.
+ * The runtime turns a token into the bytes a frame carries; the verifier compares those bytes back
+ * against the same token. Two copies of this arithmetic would agree until the day one of them
+ * rounded differently, and the failure would read as a rendering defect rather than as the drift it
+ * was.
+ */
+[[nodiscard]] constexpr std::uint8_t quantise(float channel) noexcept {
+    const float clamped = channel < 0.0F ? 0.0F : (channel > 1.0F ? 1.0F : channel);
+    const float scaled  = clamped * 255.0F;
+    const float rounded = scaled + 0.5F;
+    return static_cast<std::uint8_t>(rounded);
+}
+
+/// The governed table's linear RGBA as the quantised colour a frame actually shows.
+[[nodiscard]] constexpr mdux::core::ColorRgba8 quantise(const std::array<float, 4>& linear) noexcept {
+    return mdux::core::ColorRgba8{.r = quantise(linear[0]), .g = quantise(linear[1]), .b = quantise(linear[2]), .a = quantise(linear[3])};
 }
 
 /**
--- a/include/mdux/verify/Verify.cppm
+++ b/include/mdux/verify/Verify.cppm
@@ -1,0 +1,766 @@
+/**
+ * @file Verify.cppm
+ * @brief Rendered-truth checks: four pure functions over a CPU framebuffer and an artifact-derived
+ *        expectation, answering whether a frame shows what the compiled screen said it would.
+ *
+ * @compliance ADR-004 Trust zones in C++ (governed zone: std only, no Vulkan, no windowing)
+ * @compliance ADR-005 Error handling and exceptions policy (Result-returning, noexcept)
+ * @compliance ADR-011 The deterministic `.medui` compile boundary
+ * @compliance ADR-012 What a compiled screen emits, and which parts are committed
+ * @compliance ADR-014 What rendered-truth verification checks, and what it cannot
+ *
+ * Part of MduXCore, and that placement is ADR-014 decision 1 rather than an inheritance: the checks
+ * have the same shape as the screen runtime - bounded arithmetic over caller-owned storage, no
+ * allocation, no throw, no file - and putting them in the governed zone is what turns "the verifier
+ * does no work a device could not do" from a convention into a lint result. `mdux-governed-lint`
+ * and `governed.noThrow.symbolScan` cover this module with no second registration.
+ *
+ * **The driver is not here.** Rendering a frame, reading `goldens.json`, enumerating obligations and
+ * writing `verification.json` are what ADR-004 and ADR-005 keep out of this zone; they are #253's
+ * and #254's, and they call in. What this module owns is the four questions themselves.
+ *
+ * ## The four checks, and which are opt-in
+ *
+ * - **`goldenBounds()`** - the golden node's rendered content occupies the declared rectangle;
+ * - **`colorHash()`** - that content carries the tint its colour token resolves to;
+ * - **`inkContainment()`** - a compiled text node's run stays inside its node bounds;
+ * - **`localizedTextPresence()`** - the approved locale's bound run is the one on screen.
+ *
+ * The first two are `Bounds` and `ColorHash`, the closed `CvCheck` set an author opts into per node.
+ * That set belongs to the shared language's `safety` capability, pinned in `medui-conformance.toml`,
+ * and `MEDUI-E071` rejects a name outside it - so widening it is an upstream change and a re-pin,
+ * never an edit here. `CvCheck` is *defined* in this module and aliased by
+ * `mdux.tools.medui.goldens` rather than declared twice, for the reason ADR-012 decision 4 gives
+ * about the golden predicate: two implementations of one closed set agree until the day they matter.
+ *
+ * The last two are **not** opt-in and are deliberately not `CvCheck` enumerators. They apply to
+ * every compiled node whose spec carries a `textKey`, in every approved locale, whether or not that
+ * node has a golden entry. A glyph run leaving the box that was budgeted for it is a defect however
+ * the author annotated the node, and #195 already measured that box at compile time - these are the
+ * same claim, verified against pixels.
+ *
+ * ## What a check is given, and what it may never be given
+ *
+ * A check takes a `FramebufferView` - bytes, extent, row stride, format - and one of two expectation
+ * views. Both views are read-only, caller-owned, and obtainable only through a `create()` that
+ * resolves them against a compiled screen; neither carries a Vulkan type, allocates, throws or opens
+ * a file. The GPU's involvement ended at the readback that already exists in
+ * `mdux.render.offscreen`.
+ *
+ * ADR-014 decision 2 is the rule the `create()` functions enforce: **every expectation is derived
+ * from a committed artifact, and none is supplied by the caller.** A golden expectation is built
+ * from a `goldens.json` entry *and* the compiled node it names, and it fails closed when the id does
+ * not resolve or when the entry's duplicated bounds, text key or colour token disagree with that
+ * node. A text expectation is built from the compiled node, the approved locale, the run the
+ * approved text package binds and the font package it was baked against. A production caller has
+ * exactly those sources. A unit test may construct a synthetic view to exercise a pure check, which
+ * is what makes the suite runnable with no GPU.
+ *
+ * The verifier **never re-applies ADR-011's golden predicate**. `collectGoldens()` is its single
+ * implementation, it runs in the baker while the AST still carries the predicate's inputs, and
+ * `evidence.screen.<id>` re-derives `goldens.json` and byte-compares it on all four automatic CI
+ * legs. A golden missing for a node the predicate selects fails four byte comparisons before it can
+ * reach this module. What this module owns is the other direction: a golden naming a node that does
+ * not exist, or naming one whose rectangle it does not agree with.
+ *
+ * ## Two resolved colours, and why one is not enough
+ *
+ * Both expectation views carry a **tint** and a **ground**, and both are resolved through the same
+ * governed table (`mdux::medui::resolveColorToken()`) rather than measured from any frame.
+ *
+ * The tint is what a check compares *against*: it is the colour the node's token names. The ground
+ * is what the frame would show at that node if the node painted nothing - the surface the driver
+ * cleared to, or the tint of the compiled panel underneath. Deciding whether a pixel was painted at
+ * all needs that second value, and no amount of knowing the tint supplies it: a node drawn in
+ * `Theme.Colors.Nominal` over a cleared surface and the same node not drawn at all differ only in
+ * what the untouched pixels are. The ground is a property of the screen and of the driver's own
+ * fixed clear colour, so it is still derived rather than argued - what it must never be is a value
+ * read back out of the frame under test, which would make every check compare a measurement against
+ * itself.
+ *
+ * With both in hand the two colour claims are exact, and stay exact without a tolerance:
+ *
+ * - a pixel the run or the node painted is an alpha blend of the tint over the ground, so each of
+ *   its channels lies **between** the ground's and the tint's. A painted pixel outside that closed
+ *   interval was painted by something else, and `Finding::ForeignColour` says so.
+ * - a **fully covered** pixel is exactly the tint, because coverage modulates alpha and never rgb.
+ *   So "carries its tint" is an equality rather than a proximity, which is the same discipline
+ *   `tests/render/PixelTests.cpp` applies to a whole frame: a one-channel difference is the smallest
+ *   wrong answer a renderer can give, and the one a tolerance would wave through.
+ *
+ * The cost of that second rule is stated rather than discovered: content that never reaches full
+ * coverage anywhere - a hairline trace, or a glyph run rasterised too small for any pixel to be
+ * solid - cannot be said to carry its tint, and `colorHash()` reports `TintAbsent` rather than
+ * passing on a blend. That is fail-closed in the direction this epic wants.
+ *
+ * ## What each check can see, and what it cannot
+ *
+ * Worth stating here rather than leaving a reader to infer it from the implementation, because the
+ * limits are structural rather than temporary.
+ *
+ * `goldenBounds()` asks whether the content inside the declared rectangle *is* that rectangle: the
+ * bounding box of everything painted there has to be the box the compiler resolved, edge for edge.
+ * ADR-014's worked example is what fixes that reading rather than the weaker "there is something in
+ * there" - "a unit test paints a rectangle into one and asserts that `GoldenBounds` passes, then
+ * moves it by a pixel and asserts that it fails", and only an equality fails on a rectangle that
+ * moved by one pixel while still overlapping its own box.
+ *
+ * Two consequences, both stated because a reader would otherwise meet them as surprises. Content
+ * that deliberately fills part of its node - digits inset in a `NumericDisplay`'s box - does not
+ * discharge `Bounds` against that node's rectangle, because the rectangle the golden pins is the
+ * node's; a component whose painted extent is smaller than its box needs a golden that says so, and
+ * that is a question for whoever gives such a component its geometry. And overflow *into a
+ * neighbour* is not attributable to this node at all: the committed screen packs
+ * `insufflation-pressure` and `ecg-lead-ii` edge to edge, so a band scanned outside the rectangle
+ * would report the neighbour's pixels as this node's spill. Attribution from a framebuffer alone is
+ * not available, and a check that guessed at it would fail on a correct screen - which trains a team
+ * to ignore it.
+ *
+ * `inkContainment()` compares the ink the committed run and font metrics *predict*, placed by the
+ * rule `mdux.medui.screen` fixes, against the ink the frame actually shows inside the node. A run
+ * whose predicted box leaves the node fails on the prediction; a frame that clipped, moved or lost
+ * the run fails on the comparison.
+ *
+ * `localizedTextPresence()` is the one that distinguishes one locale from another, and it does so
+ * per glyph: every non-blank record of the approved locale's run must have painted something inside
+ * its own placed rectangle, and every pixel of the node *outside* every such rectangle must still be
+ * the ground. The second half is what makes it a presence check rather than a plausibility check.
+ * The baked atlas slot is exactly the glyph's bitmap, so the renderer paints nothing outside it, and
+ * a different translation's run - different glyphs, different advances, different spaces - cannot
+ * satisfy both halves at this locale's rectangles. Its limit is the mirror of `goldenBounds()`': a
+ * second node overlapping this one would show up as ink where this locale's run paints none.
+ *
+ * ## No allocation, no throw, and a `create()` that cannot be brace-elided
+ *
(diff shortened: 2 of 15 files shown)


src/verify/Verify.cpp
```diff
@@ -0,0 +1,567 @@
...
+    }
+
+    const auto rowBytes = static_cast<std::uint64_t>(width) * stride;
+    if (static_cast<std::uint64_t>(rowStride) < rowBytes) {
+        return err(VerifyError::RowStrideTooSmall);
+    }
+    // The last row need not be padded, so the requirement is every full row but the last, plus one
+    // row of pixels. 64-bit throughout: a large extent multiplied in `size_t` on a 32-bit host would
+    // wrap into admitting exactly the buffer this refuses.
+    const auto required = (static_cast<std::uint64_t>(height) - 1) * static_cast<std::uint64_t>(rowStride) + rowBytes;
```
ambroise-leclerc: Blocking: guard this size calculation against overflow before constructing the view. For example, with `width = 1`, `height = 2`, `rowStride = SIZE_MAX`, and a four-byte span, `required` wraps from `SIZE_MAX + 4` to `3`, so `create()` admits the view. `pixelAt(0, 1)` then computes `offset = SIZE_MAX`; its `offset + 4` bounds check wraps to `3`, after which `bytes_[SIZE_MAX]` is read out of bounds. The comment says 64-bit arithmetic prevents hostile wrapping, but unsigned 64-bit arithmetic still wraps here. Please use checked multiplication/addition (or a division/subtraction-based bound), harden `pixelAt()` so its defensive check cannot wrap, and add a `SIZE_MAX`-stride regression test.

include/mdux/verify/Verify.cppm
```diff
@@ -0,0 +1,766 @@
...
+     * colour token does not resolve, and a run that paints no ink at all.
+     *
+     * The last refusal is a policy rather than an accident. A run whose glyphs are all blank - a
+     * single space - is legitimate for the runtime, which draws nothing and does not count the node
+     * as deferred. It is not something a rendered-truth obligation can discharge: there is no ink to
+     * find, so a pass would be indistinguishable from a screen that lost its text. ADR-014 decision
+     * 3 says a check this build cannot perform fails, so this refuses rather than passing vacuously,
+     * and the screen is what changes.
+     */
+    [[nodiscard]] static mdux::core::Result<TextExpectation, VerifyError> create(const mdux::medui::CompiledNode& node,
```
ambroise-leclerc: Blocking: this factory does not establish the artifact provenance promised by the module contract and ADR-014 decision 2. It accepts any non-empty locale string, arbitrary run bytes, and any `FontPackage`; it never sees the `ScreenPackage`, `TextPackage`, canonical package digest, or the existing `TextBinding` that authenticates all of them. A caller can therefore create a production-looking expectation for an unapproved locale/run/font, and a future driver mistake would become a passing verification artifact. Please construct production expectations from a validated `TextBinding` plus the target screen (checking `approvedBy(screen)` and selecting the node run), while keeping synthetic construction behind a clearly test-only/internal path if the unit suite needs it.

src/verify/Verify.cpp
```diff
@@ -0,0 +1,567 @@
...
+            // nothing to find and nothing missing.
+            continue;
+        }
+        outcome.glyphIndex = index;
+        if (!frame.contains(*glyph)) {
+            outcome.found      = *glyph;
+            outcome.foundValid = true;
+            return failed(outcome, Finding::RegionOutsideFrame);
+        }
+        if (!paintedBox(frame, *glyph, ground).inked) {
```
ambroise-leclerc: Blocking: this establishes occupancy of each glyph bounding rectangle, not that the approved glyph was rendered. One arbitrary non-ground pixel anywhere inside every non-blank glyph rectangle satisfies this half; the second half only requires all painted pixels to remain inside the union of those rectangles. Consequently a sparse set of pixels—or different glyph shapes with the same placements and bounds—can satisfy `LocalizedTextPresence` (and a few edge pixels can also preserve the overall `inkContainment` extent). Because `TextExpectation` carries metrics but no authenticated atlas coverage, the implementation cannot substantiate its public claim that the approved locale’s bound run is the one on screen. Please compare against the expected atlas glyph masks/coverage, or narrow the check and its evidence claim, and add a sparse-ink adversarial test.

src/verify/Verify.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+/**
+ * @brief Whether `pixel` could be the tint blended over the ground at some coverage.
+ *
+ * Channel-wise containment, and it needs no tolerance. Alpha blending produces a convex combination
+ * of two integer channel values, and rounding a value that lies between two integers cannot leave
+ * the closed interval they span - so a correct blend is always inside, and a pixel outside was
+ * painted by something the expectation does not describe.
+ */
+[[nodiscard]] bool blendOf(ColorRgba8 pixel, ColorRgba8 ground, ColorRgba8 tint) noexcept {
```
ambroise-leclerc: Blocking: channel-wise interval membership is necessary for a blend, but it is not sufficient. All colour channels must result from one shared coverage value. With black ground and the current `Theme.Colors.ScoreDigits` tint `(33, 184, 107, 255)`, the pixel `(33, 0, 107, 255)` passes this function even though no coverage can produce it: red/blue require full coverage while green requires zero. If the region contains one exact tint pixel, `colorHash()` then accepts every such impossible pixel and returns `Held`. Please validate a common blend factor (using integer cross-products and an explicit allowance for UNORM rounding) and add a regression containing one exact-tint pixel plus an impossible channel combination.

src/verify/Verify.cpp
ambroise-leclerc: Fixed in `8c93f821d4defeb56dbd0cc6d6841a1f07cc8e1e`, and the comment was wrong as well as the code — widening to 64 bits does nothing about wrapping, which is exactly what your `rowStride = SIZE_MAX` case shows.

`create()` now establishes the bound by division and subtraction, before anything is multiplied:

```cpp
if (room < rowBytes) { return err(VerifyError::FramebufferTooSmall); }
if (rows != 0 && static_cast<std::uint64_t>(rowStride) > (room - rowBytes) / rows) {
    return err(VerifyError::FramebufferTooSmall);
}
```

`pixelAt()` is hardened the same way — `offset > bytes_.size() - pixelBytes` after a `size() < pixelBytes` guard, rather than `offset + pixelBytes > size()`, so its own defensive check cannot wrap past the end of the caller's span. It is now the second gate rather than the only one.

Regression added to `verify-framebuffer-fails-closed`: `create(bytes, 1, 2, SIZE_MAX, Rgba8Unorm)` must be refused as `FramebufferTooSmall`. It fails on the previous implementation.

src/verify/Verify.cpp
ambroise-leclerc: Fixed in `8c93f821d4defeb56dbd0cc6d6841a1f07cc8e1e`. You are right, and the counter-example is now in the test rather than only in the review: channel-wise membership is necessary and not sufficient, because a blend applies **one** coverage to every channel.

`blendOf()` is replaced by `couldBeBlend()`, which intersects the coverage interval each channel implies. Integer cross-products, no division and no float: the coverage a channel implies is the rational `(pixel - ground) / (tint - ground)`, so the intervals are compared by multiplying out, and `allowance` is one UNORM step for the device's float-to-eight-bit quantisation. A channel whose tint and ground are equal constrains nothing about the coverage and everything about the pixel, so it is handled separately.

On your example, with black ground and `(33, 184, 107)`: red gives `[32/33, 34/33]`, green gives `[-1/184, 1/184]`, the intersection is empty, `ForeignColour`.

`verify-golden-color-hash-common-factor` paints exactly that: a rectangle filled with the exact tint plus `(tint.r, ground.g, tint.b)` at one pixel, and asserts `ForeignColour` with that pixel named. The same scenario then paints `blend(ground, tint, 128)` in the same place and asserts `Held` — otherwise the first half could be satisfied by a rule that rejected everything but the tint itself.

src/verify/Verify.cpp
ambroise-leclerc: Fixed in `8c93f821d4defeb56dbd0cc6d6841a1f07cc8e1e` by the first of your two options: the view now carries the coverage.

You are right that metrics alone cannot substantiate the claim. `TextExpectation` gains the atlas sheet — validated against `font.atlas.byteLength == width * height`, with every glyph slot required to be inside it — and `glyph(index)` returns a `PlacedGlyph` carrying both the destination rectangle and the slot it paints from. `localizedTextPresence()` then compares **every pixel of every placed glyph** against `blend(ground, tint, coverage)` within one UNORM step: a texel the atlas leaves at zero must be exactly the ground, a fully covered one exactly the tint, and everything between is pinned to the coverage the baker recorded. The blend equation is the one `tests/render/TextPixelTests.cpp` already derives its expectations from and compares to real pixels with no tolerance, so it is this repository's established statement rather than a second opinion.

Two adversarial scenarios, both of which pass on the previous implementation:

- `verify-text-presence-rejects-sparse-ink` — one pixel painted inside each glyph rectangle. Every metrics-only check accepts it.
- `verify-text-presence-rejects-substituted-glyph` — the solid glyph's coverage painted at the sparse glyph's placement and bounds. Same rectangle, same tint, different letter.

One consequence worth flagging: the finding is now decided **per glyph** rather than at the first disagreeing pixel. A run drawn in the wrong tint disagrees at its very first pixel and has painted plenty, while an absent glyph disagrees at the same pixel and has painted nothing — deciding from scan position called the first one `GlyphMissing`, which would send a reader looking for a missing translation instead of a wrong colour.

`InkContainment` is left as the extent claim it actually makes, and its comment now says so explicitly and points at presence for shape, rather than implying it establishes both.

include/mdux/verify/Verify.cppm
ambroise-leclerc: Fixed in `8c93f821d4defeb56dbd0cc6d6841a1f07cc8e1e`, along the line you drew.

`create()` now takes the screen, the node and a `mdux::medui::TextBinding`, and adds exactly the provenance the binding cannot know about on its own:

```cpp
if (!binding.bound())                    return err(VerifyError::TextBindingUnbound);
if (!binding.approvedBy(screen))         return err(VerifyError::BindingNotApproved);
if (screen.find(node.id) != &node)       return err(VerifyError::NodeNotInScreen);
if (scope.tag() != binding.text()->locale) return err(VerifyError::ScopeIsNotTheBoundLocale);
```

The run is then looked up by the node's own `textKey` in the bound package. The caller supplies no locale, no bytes and no font — `TextBinding::create()` has already proved the manifest approves this locale, id and canonical digest, that the package was baked against this font, and that every range hashes to what it recorded.

The node identity check is `find(id) != &node` rather than `!= nullptr` deliberately: a node from another screen with the same id would otherwise be verified against a manifest that never mentioned it. And the scope check is what stops an English frame being reported as a passing German outcome.

The synthetic path is kept, behind `createSynthetic()`, because ADR-014 decision 1 admits it by name and bounds it in the same sentence. Its comment says it establishes none of the above and that a driver reaching for it is a review finding on #253 rather than a runtime condition this module can detect.

`verify-text-expectation-provenance` builds a real binding — font package, text package, canonical bytes from `TextPackage::write()`, sidecar, and a screen approval computed from the package's own digest, so nothing is typed out by hand — then shows each refusal: an unbound binding, a screen approving the same locale and id under a different digest, a node from another screen, and a `de-DE` scope over an `en-US` binding.

include/mdux/verify/Verify.cppm
```diff
@@ -0,0 +1,766 @@
...
+/**
+ * @brief One obligation's result: what was expected, what was found, and where.
+ *
+ * Structured rather than a formatted string, because this module allocates nothing and a governed
+ * type that carried a message would have to. #253's driver renders these into the sentence ADR-014
+ * asks for; everything that sentence needs is here, including the render scope, so a reader can tell
+ * a failure in `de-DE` from the same node's pass in `en-US`.
+ *
+ * `found` is only meaningful for the findings that measured something. `expected` is always the
+ * rectangle the check was asked about.
```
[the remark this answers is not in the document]
ambroise-leclerc: Good catch — the comment was written before `inkContainment()` started carrying the predicted ink box, and `localizedTextPresence()` now does the same thing per glyph.

Rather than making the field mean one thing everywhere, the comment now says what each check puts there, since a driver formatting "expected X, found Y" has to know which X it is holding:

> `expected` is **the rectangle this outcome's claim is about**, which is not always the node's box. `goldenBounds()` and `colorHash()` name the golden's declared rectangle; `inkContainment()` names the node's rectangle while the failure is about containment and the predicted ink box once the comparison has moved on to the frame; `localizedTextPresence()` names one glyph's rectangle for a glyph-level finding and the node's otherwise.

ambroise-leclerc:
## Review addressed in `8c93f82`

All four blocking findings and the documentation one, each taken as stated. Per-thread detail is in the replies; the summary and what changed about the PR's claims:

| Finding | Fix | Regression |
|---|---|---|
| `FramebufferView` overflow | bound established by division and subtraction; `pixelAt()` guards by subtraction too | `create(bytes, 1, 2, SIZE_MAX, …)` refused |
| `blendOf()` per-channel only | `couldBeBlend()` intersects the coverage interval each channel implies, integer cross-products, one-step UNORM allowance | exact-tint fill + `(tint.r, ground.g, tint.b)` → `ForeignColour`; genuine half-coverage blend still `Held` |
| `LocalizedTextPresence` proved occupancy, not shape | the view carries the baked coverage sheet; every pixel of every placed glyph compared against `blend(ground, tint, coverage)` | sparse ink, and a substituted glyph at the same placement and bounds |
| `TextExpectation` had no provenance | `create(screen, node, binding, atlas, scope, ground)`; run looked up by the node's own `textKey`; synthetic path split off as `createSynthetic()` | a real `TextBinding` built from self-consistent artifacts, plus four refusals |
| `CheckOutcome::expected` doc | says what each check puts there | — |

### Three things worth calling out beyond the fixes

**The blend equation is not a new one.** `tests/render/TextPixelTests.cpp` already derives its expectations from `result = tint * a + ground * (1 - a)`, with `a = (tint.a / 255) * (coverage / 255)`, and compares them to real pixels with **no** tolerance under lavapipe and MoltenVK. `blend()` is that equation in integers, so the coverage check is comparing against the repository's established statement of what a covered texel produces rather than against a second opinion. The one-step allowance is for the device's float-to-eight-bit quantisation and nothing else — a wrong shape misses by far more, and a wrong tint by more still.

**A finding that changed meaning.** `localizedTextPresence()` now decides `GlyphMissing` vs `CoverageDiffers` **per glyph** rather than at the first disagreeing pixel. A run drawn in the wrong tint disagrees at its very first pixel and has painted plenty; an absent glyph disagrees at the same pixel and has painted nothing. Deciding from scan position called the first one "painted nothing", which sends a reader to look for a missing translation instead of a wrong colour. Caught by the wrong-tint scenario when the coverage comparison landed.

**`InkContainment` was narrowed rather than strengthened.** Your note that a few edge pixels preserve its extent is right, and the answer is not to duplicate the coverage comparison in both checks: it is an extent claim, its comment now says so explicitly, and it points at presence for shape. The two checks now establish different things, which is what having two of them is for.

### Verification

Local leg GCC 16.1.0 / CMake 4.2.3 / Ninja, Release, lavapipe.

- `ctest --test-dir build-252` — **651/651** (was 647; four new scenarios)
- `verify_spec` — 22 scenarios, 0 failed
- `mdux-governed-lint` — 0 findings; `ctest -L governed` — both scans pass
- `offscreen_tests` (`pixel`, lavapipe) — 36/36
- `mdux-docs-lint` — 0 findings; `clang-format` clean

Every new scenario was checked against the previous implementation and fails on it.

src/verify/Verify.cpp
```diff
@@ -0,0 +1,815 @@
...
+                const ColorRgba8 wanted = blend(ground, tint, expectation.coverage(*placed, dx, dy));
+                if (!withinOneStep(*pixel, wanted) && !disagreed) {
+                    disagreed   = true;
+                    firstX      = x;
+                    firstY      = y;
+                    firstFound  = *pixel;
+                    firstWanted = wanted;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `071e511c448bd994cccd65070d5fad7cbb2779f1`, by the first of your two options: the coverages are composed.

You are right that nothing constrains the advance. Rejecting overlapping placements was the other option and I did not take it: a font whose glyphs legitimately overlap would then be unverifiable, and refusing the screen because the checker cannot express what the renderer does is the failure mode ADR-014's risk section is about. Composing is also what the draw path *actually* does, so it is exact rather than tolerant — one quad per glyph blended in turn expands to the tint weighted by `1 - (1 - a1)(1 - a2)` and the ground by its complement, and the composition is exact only because every glyph of one run carries one tint over one ground. The allowance became one UNORM step **per composite** rather than a fixed one, since the device quantises to eight bits at each step.

Reading every glyph for every pixel of the node would have been quadratic in the glyph count for nothing: outside the run's ink box no glyph can cover a pixel, so the walk is skipped there and the answer is the ground.

`verify-text-presence-composites-overlap` is the scenario you asked for, with one deliberate choice. It places the **sparse** glyph twice rather than the solid one: where a solid glyph overlaps anything the composition is 255 and so is either coverage alone, so that scenario would have claimed a composited check while testing nothing. Two half-covered texels compose to 192 — a value neither glyph produces by itself — and the negative half paints that column from one glyph's coverage alone and must fail, missing by four against a two-layer allowance. The fixture painter also had to change: it blends each quad over what the previous one left, rather than computing a composited expectation in one pass, or the frame would have shared the assumption it is meant to test.

Your reading is confirmed on the tree as it stands: the committed `endoscope-monitor-en-us` run has no overlapping pair, so this was a fault waiting on a font rather than one in a current artifact — which is why the scenario has to construct one.

include/mdux/medui/Schema.cppm
```diff
@@ -299,6 +300,37 @@ inline constexpr std::array<ThemeColor, 8> themeColors{
...
+ * The runtime turns a token into the bytes a frame carries; the verifier compares those bytes back
+ * against the same token. Two copies of this arithmetic would agree until the day one of them
+ * rounded differently, and the failure would read as a rendering defect rather than as the drift it
+ * was.
+ */
+[[nodiscard]] constexpr std::uint8_t quantise(float channel) noexcept {
+    const float clamped = channel < 0.0F ? 0.0F : (channel > 1.0F ? 1.0F : channel);
+    const float scaled  = clamped * 255.0F;
+    const float rounded = scaled + 0.5F;
+    return static_cast<std::uint8_t>(rounded);
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `071e511c448bd994cccd65070d5fad7cbb2779f1`. Correct as stated: NaN compares false against both clamps, so it reached the conversion, and that is undefined for a value outside the destination's range.

```cpp
if (!(channel >= 0.0F)) {
    return 0;
}
const float clamped = channel > 1.0F ? 1.0F : channel;
```

`!(channel >= 0.0F)` is the one test that catches NaN and negatives together, so the clamp below is left with only the upper bound. Zero because a channel this function cannot interpret contributes nothing, and because an error path would be a poor trade on a function whose only production input is `themeColors`, a `constexpr` table of literals — the value is unreachable through `resolveColorToken()`.

Worth saying why it is worth fixing anyway rather than closing as unreachable: this branch is what *exported* `quantise()` out of `Screen.cpp`'s anonymous namespace, so it is public API now and a total function is part of what that costs. One comparison.

tests/verify/GoldenCheckTests.cpp
```diff
@@ -0,0 +1,531 @@
...
+                      checks.expect(wrong.expectedColor == tint, "beside the tint it expected");
+
+                      // A legitimate blend of the tint over the ground at half coverage: inside the
+                      // interval everywhere, and equal to the tint nowhere.
+                      Canvas partial{16, 20, ground};
+                      partial.fill({4, 4, 8, 6},
+                                   ColorRgba8{.r = static_cast<std::uint8_t>(tint.r / 2),
+                                              .g = static_cast<std::uint8_t>(tint.g / 2),
+                                              .b = static_cast<std::uint8_t>(tint.b / 2),
+                                              .a = 255});
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in `071e511c448bd994cccd65070d5fad7cbb2779f1` — `mv::blend(ground, tint, 128)`.

The scenario passed as written, since `tint / 2` still satisfies the common-factor test with the one-step allowance, but you are right that it passed for a reason the scenario is not about: it is meant to be *a legitimate blend that never reaches the tint*, and `tint / 2` is only coincidentally one. Building it from the function under test's own blend equation makes it that by construction, and it stops being sensitive to whether a channel is odd.

AM-L (approved):

PR #278, opened 2026-09-02
S4 — Emit verification.json as an evidence artifact [#254]

### LINKED ISSUE #254: S4 — Emit verification.json as an evidence artifact
Part of #16. Blocked by #253.

## Work

Emit `generated/screen/<id>/verification.json` as the fourth committed output of the existing screen
artifact bundle, following ADR-007's pattern: canonical JSON, floats as `u32` bit patterns, the
existing `report.json` extended with the fully resolved verification options and output digest, and
byte-identity re-derived in CI.

Consume the obligation outcomes produced by `MduXVerifyUiLib` from #253. The artifact writer is
host-only and shares that library path with the `mdux-verify-ui` executable; it does not re-enumerate
obligations, re-apply ADR-011's predicate or invent a parallel verification implementation.

`mdux_compile_screen()` remains the sole owner of
`mdux_bake_artifact(KIND screen ID <id> ...)`. Extend its single screen-bundle production sequence
and fixed `OUTPUTS` so compilation, rendered verification and serialization produce one coherent
bundle. Do not make a second registration for the same kind/id and do not create a second report at
the same path.

Normal builds write the enlarged bundle into `${CMAKE_BINARY_DIR}/mdux_bake/screen/<id>/` and
compare it with the committed bundle. `mdux-bake-update` remains the only path that stages the
result into `generated/screen/<id>/`; neither `mdux-verify-ui` nor an ordinary build writes directly
into the source tree.

The artifact records one outcome per enumerated obligation. Each outcome makes only its own claim:
`Bounds` says where the node appeared, `ColorHash` says which tint was observed, and the two text
outcomes say which approved run appeared and whether its ink remained contained. A node carrying
only `Bounds` must not be described as tint-checked. The artifact names the screen, goldens, font and
per-locale text package inputs by digest and does not claim the screen is clinically correct.

## Four-leg execution constraint

ADR-007 and ADR-014 require the enlarged screen bundle to be re-derived and byte-compared on all
four automatic evidence legs. A leg re-derives `verification.json` only if the #253 producer really
renders and evaluates every obligation there. Copying the committed file, serializing synthetic
passing outcomes, or treating a missing loader/device as a skip would satisfy byte comparison while
providing no rendered evidence and is forbidden.

If all four legs cannot execute the producer, stop and revise the recorded ADR decision explicitly;
do not silently reduce the leg count or weaken the meaning of `evidence.screen.<id>` in this issue.

## Acceptance

- [ ] `verification.json` is added to the existing screen bundle's `OUTPUTS`, so the one
      `evidence.screen.<id>` test re-runs the real verification producer and byte-compares all four
      files on every ADR-007 evidence leg.
- [ ] The writer consumes the same enumerated outcomes as `mdux-verify-ui`; there is no second
      obligation predicate, synthetic production path or example-owned path.
- [ ] The existing screen `report.json` records the new output digest and fully resolved logical
      options; no second report or duplicate bake target/test is created.
- [ ] Every outcome is scoped to one obligation and cannot imply an unperformed check.
- [ ] A missing loader/device or other inability to run fails artifact production distinctly from a
      failed rendered check; neither can become a committed passing outcome.
- [ ] No commit SHA, timestamp, absolute path, duration or measured pixel sample appears in the
      committed artifact.
- [ ] The committed artifact is regenerated only by `mdux-bake-update` and its diff is reviewable.

ambroise-leclerc:
Closes #254.

`mdux-verify-ui` rendered and evaluated every obligation and then threw the result away at the
process boundary. This serializes it as the screen bundle's fourth committed file.

## The shape

`mdux_compile_screen()` remains the one `screen/<id>` registration. What it gains is a second
stage: `mdux_bake_artifact()` learns `THEN_TOOLS`, so a single custom command runs `mdux-meduic`
and then the new `mdux-verify-bake` over the same output directory, under one `OUTPUTS` set and one
`evidence.screen.<id>` comparison. The rendered half therefore cannot end up with weaker evidence
than the compiled half.

Two tools rather than one because their dependencies differ. Reading a `.medui` file needs no GPU,
and fusing them would make a Vulkan device a prerequisite of every screen compile, of
`mdux-medui-check`, and of every test that drives a compile as a call.

The writer consumes `RunResult` and enumerates nothing itself — no second obligation predicate, no
synthetic production path. Findings serialize through a new `spell(Finding)` rather than
`describe()`, so improving a diagnostic's wording cannot fail an evidence comparison on four legs.

## What the artifact says

One outcome per enumerated obligation, each making only that obligation's claim. A node carrying
only `Bounds` appears once, as a `Bounds` outcome, and nothing in the file reads as saying its tint
was checked. The file names the screen package, goldens sidecar, shader package and per-locale font
and text packages by digest.

No measurement reaches it — no found rectangle, no sampled colour, no path, no duration. A
measurement would make a byte-compared file a property of the driver tuple that produced the frame
rather than of its declared inputs, which is ADR-007 decision 5's argument about commit SHAs applied
to pixels.

An inability to run fails production and writes nothing. A check that ran and did not hold is
recorded as the finding it produced. The committed `endoscope-monitor` artifact therefore carries
three `NothingPainted` findings today — exactly what ADR-014 predicted for a screen whose
`NumericDisplay` and `SignalTrace` the runtime still defers to #17. An artifact that hid them would
be worse than none; #255 owns the gate.

## The four-leg constraint

Two of the four evidence legs could not have re-derived this file, and that was invisible while
nothing there rendered:

| Leg | Before | Now |
|---|---|---|
| Linux/GCC 16 | lavapipe | unchanged |
| macOS/Clang 21 | MoltenVK | unchanged |
| Linux/Clang 21 libc++ | loader, no `mesa-vulkan-drivers` — **zero devices** | lavapipe from the distribution |
| Windows/MSVC | `Vulkan-Loader` with no ICD — **zero devices** | Mesa's Windows build, pinned by tag and SHA-256 |

Since the bake now renders, a leg without a device fails to build rather than producing a smaller
artifact, so neither can satisfy the byte comparison without having actually verified the screen.
Both legs also gain the `-L pixel` skip guard the other two already carry. ADR-007 decision 6 records
that the device is now an obligation of each leg rather than an accident of two.

## Verified

- Two independent bakes produce byte-identical `package.json`, `goldens.json`, `report.json` and
  `verification.json`.
- `VK_DRIVER_FILES=/nonexistent.json` exits 1, writes nothing and leaves `report.json` untouched —
  distinct from a failed check, which exits 0 and records the finding.
- `generated/` is unchanged in digest across a build; only `mdux-bake-update` stages it.
- 666/666 tests including the `evidence` label; evidence-lint, docs-lint and the governed no-throw
  scans are clean.

Not verified locally: the device here is radv, not lavapipe or MoltenVK. The artifact holds outcomes
rather than samples, so byte-identity should survive four drivers — but this PR's own CI is what
proves it.

## Summary by CodeRabbit

- **New Features**
  - Screen bundles now include a `verification.json` artifact summarizing verification outcomes, resolved locales, and input integrity data.
  - Screen generation now performs verification automatically after rendering and updates the bundle report with verification stages.
  - Bundle updates are published atomically, preventing partial output when generation fails.

- **Bug Fixes**
  - Linux Clang and Windows CI now provide software Vulkan rendering and fail when rendered tests are skipped, improving test reliability.

- **Documentation**
  - Updated architecture, roadmap, and decision records to document the rendered verification workflow and CI coverage.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -119,14 +119,24 @@ jobs:
         if (-not $icd) {
           throw "mesa3d $version carries no 64-bit lavapipe ICD manifest"
         }
-        # Both spellings, and the older one is the one that actually works here. `VK_DRIVER_FILES`
-        # arrived in loader 1.3.207; the SDK pinned above is 1.3.204.0, whose loader ignores it
-        # entirely and reports "no usable ICD" with the variable set correctly in the environment.
-        # `VK_ICD_FILENAMES` is what that loader reads. Setting both keeps this working when the SDK
-        # pin moves forward, since the newer loader prefers `VK_DRIVER_FILES`.
-        "VK_DRIVER_FILES=$($icd.FullName)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-        "VK_ICD_FILENAMES=$($icd.FullName)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-        Write-Host "lavapipe ICD: $($icd.FullName)"
+        # Registered in the registry rather than pointed at with VK_DRIVER_FILES, and the loader
+        # is emphatic about why. This runner is elevated, and an elevated loader ignores every
+        # driver-path environment variable as a hardening measure - it will not let a planted ICD
+        # into an administrator process. With the variable set correctly it says, in as many words:
+        #
+        #   INFO: Loader is running with elevated permissions.
+        #         Environment variable VK_DRIVER_FILES will be ignored
+        #
+        # and then fails vkCreateInstance with "no usable ICD". The loader it says that with is
+        # C:\Windows\SYSTEM32\vulkan-1.dll version 1.3.301 - the system loader, not the SDK's - so
+        # the variable is new enough to be understood and is discarded anyway. HKLM\...\Khronos\
+        # Vulkan\Drivers is the documented way to install an ICD system-wide, it is what an elevated
+        # loader still reads, and writing it is exactly what being elevated makes possible.
+        $key = 'HKLM:\SOFTWARE\Khronos\Vulkan\Drivers'
+        New-Item -Path $key -Force | Out-Null
+        # A driver entry is a REG_DWORD whose *name* is the manifest path; 0 means enabled.
+        New-ItemProperty -Path $key -Name $icd.FullName -PropertyType DWord -Value 0 -Force | Out-Null
+        Write-Host "registered lavapipe ICD: $($icd.FullName)"
 
     - name: Install Ninja
       run: |
@@ -156,13 +166,6 @@ jobs:
       shell: cmd
 
     - name: Build
-      env:
-        # This leg's Vulkan device is a software rasterizer installed by hand two steps above, and
-        # the build renders through it (#254). When a loader declines an ICD it does so silently and
-        # reports only "no usable ICD" at vkCreateInstance, which says nothing about why. Loader
-        # diagnostics turn that into the loader's own sentence, in the log, next to the bake that
-        # failed.
-        VK_LOADER_DEBUG: all
       run: cmake --build --preset ninja-msvc
       shell: cmd
 
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -84,6 +84,44 @@ jobs:
           & "$env:VULKAN_SDK\bin\vulkaninfoSDK.exe" --summary
         }
 
+    # Lavapipe, so this leg can execute the rendered-truth producer (#254).
+    #
+    # The component list above installs a loader without a driver, and `windows-latest` has no GPU,
+    # so this runner enumerated zero physical devices until now. That was invisible while nothing
+    # here rendered. It stopped being invisible when `verification.json` joined the screen bundle:
+    # the file is re-derived by `mdux-verify-bake` during the ordinary build, so a leg with no ICD
+    # does not produce a smaller artifact - it fails to build. ADR-007 decision 6 asks all four legs
+    # to re-derive every committed artifact, and #254 forbids reading a missing device as a skip, so
+    # the software rasterizer is what keeps this leg in the claim rather than an exception to it.
+    #
+    # Mesa's own Windows build, pinned by release tag and verified against the SHA-256 GitHub
+    # records for that exact asset. An unpinned "latest" would let a third-party binary change what
+    # this leg renders with between two runs of the same commit, under a byte-identity claim.
+    - name: Install lavapipe (Mesa software Vulkan)
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = 'Stop'
+        $version  = '26.2.0'
+        $expected = 'dcb2719ef346dab5b609fcb193a5f13cfc4b0502e3f4de1ad43d349477402f47'
+        $archive  = Join-Path $env:RUNNER_TEMP 'mesa.7z'
+        $target   = Join-Path $env:RUNNER_TEMP 'mesa'
+        $url = "https://github.com/pal1000/mesa-dist-win/releases/download/$version/mesa3d-$version-release-msvc.7z"
+        Invoke-WebRequest -Uri $url -OutFile $archive
+        $actual = (Get-FileHash -Algorithm SHA256 -Path $archive).Hash.ToLower()
+        if ($actual -ne $expected) {
+          throw "mesa3d $version is $actual, not the pinned $expected"
+        }
+        & 7z x $archive "-o$target" -y | Out-Null
+        # Located rather than hardcoded: the archive's internal layout is Mesa's business, and a
+        # moved file should fail here by name instead of as a device that silently never appears.
+        # The manifest's library_path is relative, so the DLL has to stay beside it.
+        $icd = Get-ChildItem -Path $target -Recurse -Filter 'lvp_icd*64*.json' | Select-Object -First 1
+        if (-not $icd) {
+          throw "mesa3d $version carries no 64-bit lavapipe ICD manifest"
+        }
+        "VK_DRIVER_FILES=$($icd.FullName)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        Write-Host "VK_DRIVER_FILES=$($icd.FullName)"
+
     - name: Install Ninja
       run: |
         $NINJA_URL = "https://github.com/ninja-build/ninja/releases/download/v1.11.1/ninja-win.zip"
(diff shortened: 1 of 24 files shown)


generated/screen/endoscope-monitor/report.json
```diff
@@ -41,6 +46,10 @@
     {
       "path": "package.json",
       "sha256": "79bb460bbf10e51b69eeddde9782e3ef99c464a2c8d9e71cc0601623703fde07"
+    },
+    {
+      "path": "verification.json",
```
ambroise-leclerc, line 51: **[P1] Enregistrer la provenance du second outil dans l’unique rapport.**

Cette nouvelle sortie est produite par `mdux-verify-bake`, alors que le même rapport conserve plus bas `tool: "mdux-meduic"` et ne mentionne nulle part le second exécutable. Un lecteur attribue donc `verification.json` au compilateur de screen, ce qui est factuellement faux. Cela rompt aussi le but déclaré d’ADR-007 : pouvoir déterminer quel outil, à quelle version, a produit un artefact.

Le second rapport étant explicitement interdit, le format du rapport unique doit représenter la chaîne multi-outils — par exemple par une provenance structurée de l’étape de vérification, avec son outil et sa version — et l’ADR/le schéma doivent préciser cette représentation. Ajouter seulement le digest de sortie et les locales résolues ne suffit pas à identifier son producteur.

tools/verify/VerifyBakeMain.cpp
```diff
@@ -0,0 +1,183 @@
...
+        if (!writeText(temporary, text)) {
+            discard();
+            return false;
+        }
+        staged.push_back(std::move(temporary));
+    }
+
+    for (std::size_t index = 0; index < files.size(); ++index) {
+        std::error_code failure;
+        std::filesystem::rename(staged[index], files[index].first, failure);
```
ambroise-leclerc: **[P2] Restaurer les sorties si une promotion ultérieure échoue.**

Le staging évite les fichiers tronqués, mais ne garantit toujours pas « les deux fichiers ou aucun ». Si ce premier `rename()` réussit puis que celui de `report.json` échoue — verrou Windows/antivirus, erreur d’I/O ou interruption — `staged[0]` n’existe plus. `discard()` ne peut donc ni supprimer la nouvelle `verification.json`, ni restaurer sa version précédente. Le bundle reste composé d’une vérification nouvelle et d’un rapport ancien.

C’est précisément le cas de promotion partielle que le commentaire CodeRabbit initial demandait de restaurer, bien que le fil soit maintenant marqué résolu. Il faut sauvegarder les cibles précédentes et effectuer un rollback vérifié, ou publier par remplacement atomique d’un répertoire/génération. Ajoutez un test avec injection d’échec sur la seconde promotion afin de vérifier que les deux fichiers restent byte-identiques à leur état initial.

cmake/MduXCompileScreen.cmake
```diff
@@ -133,14 +134,26 @@ function(mdux_compile_screen)
...
+        THEN_TOOLS mdux-verify-bake
         RECIPE ${ARG_RECIPE}
         SOURCES
             ${screen_source}
             ${ARG_SOURCES}
-        # Fixed here rather than per call site: all three are unconditional (ADR-012, decision 1).
+            # The committed artifacts the render reads. They are already dependencies of the compile
+            # for the text packages; the shader package is new here, and it is a real edge: re-baking
+            # mdux-ui changes the frame this screen is verified against.
+            generated/shader/mdux-ui/package.json
```
ambroise-leclerc, line 149: **[P2] Déclarer également l’atlas de police lu par la nouvelle étape.**

La dépendance shader est correctement ajoutée ici, mais `mdux-verify-bake` lit aussi le fichier désigné par `font->atlas.path` (`generated/font/dejavu-ui/atlas.bin`) dans `loadLocale()`. L’appel actuel à `mdux_compile_screen()` ne passe que le `package.json` de police, le package de texte et `runs.bin` dans `SOURCES`; l’atlas manque donc au graphe Ninja.

Dans un build incrémental dont les quatre sorties existent déjà, une modification isolée de l’atlas ne relance pas cette custom command. `evidence.screen.endoscope-monitor` peut alors comparer un ancien bundle de build et passer sans réexécuter le producteur rendu, contrairement à la contrainte de #254. Ajoutez explicitement `generated/font/dejavu-ui/atlas.bin` aux sources du screen, ou résolvez systématiquement tous les sidecars référencés pour que chaque fichier effectivement lu soit une dépendance.

.github/workflows/windows-build.yml
```diff
@@ -84,6 +84,60 @@ jobs:
...
+    # this leg renders with between two runs of the same commit, under a byte-identity claim.
+    - name: Install lavapipe (Mesa software Vulkan)
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = 'Stop'
+        $version  = '26.2.0'
+        $expected = 'dcb2719ef346dab5b609fcb193a5f13cfc4b0502e3f4de1ad43d349477402f47'
+        $archive  = Join-Path $env:RUNNER_TEMP 'mesa.7z'
+        $target   = Join-Path $env:RUNNER_TEMP 'mesa'
+        $url = "https://github.com/pal1000/mesa-dist-win/releases/download/$version/mesa3d-$version-release-msvc.7z"
```
ambroise-leclerc, line 108: **[P2] Enregistrer cette nouvelle dépendance de vérification dans le registre SOUP/outillage.**

Le checksum et la version sont bien épinglés, mais cette PR introduit et exécute une distribution tierce `pal1000/mesa-dist-win` pour produire l’évidence rendue commise. Elle n’est pas couverte par l’entrée actuelle `vulkan-sdk`, qui identifie le loader Khronos et son dépôt, pas cette distribution Mesa/son fournisseur.

Le périmètre de `docs/governance/soup-register.toml` inclut explicitement l’outillage de build et de vérification. Il faut y enregistrer le composant exact, sa version/source, sa licence, son rôle limité à la CI, son pin SHA-256 et les contrôles/limitations associés. Sans cela, un binaire qui influence directement l’évidence de vérification reste absent de la traçabilité gouvernée.

ambroise-leclerc (commented):
Revue détaillée au SHA `2b569a697e7ac2c7abeb5bbf695a4db74f6244c8`. **Recommandation : corriger avant fusion.** Les checks requis sont verts, mais quatre écarts restent à traiter : provenance du second outil dans le rapport, rollback incomplet de la publication par paire, dépendance manquante sur l’atlas de police et absence d’enregistrement du nouveau composant Mesa utilisé pour produire l’évidence. Les commentaires inline donnent les scénarios d’échec et les corrections attendues. GitHub n’autorise pas le compte auteur de la PR à lui appliquer lui-même le statut `Request changes`; cette revue est donc publiée comme commentaire sans affaiblir la recommandation.

AM-L (approved):

Issue #257, opened 2026-08-28
S2 — SignalTrace: ring-buffer binding and CPU polyline expansion


ambroise-leclerc:
Part of #17.

## Work

`SignalTrace`, bound to a **caller-owned ring buffer** through a binding slot. Expansion writes into
the pre-sized vertex budget from #13 — the buffer is sized once from the compiled screen's budget and
never grown, which is what keeps the no-heap property true in a frame.

Each segment becomes a quad with round-cap overdraw rather than mitred joins: cheaper, and visually
indistinguishable at 1–3 px stroke widths.

## Acceptance

- [ ] No allocation in the per-frame path, verified the three ways #63 established.
- [ ] A trace whose sample count would exceed the node's budget is refused, not truncated silently.
- [ ] The ECG demonstrator from #18 is a waiting consumer — wire it, so the component has a real
      caller rather than only a test.

PR #279, opened 2026-09-02
S5 — CI verification across all locales, with a diff image on failure [#255]

### LINKED ISSUE #255: S5 — CI verification across all locales, with a diff image on failure

ambroise-leclerc:
Closes #255. Closes epic #16 at 5/5.

`mdux-verify-ui --screen=generated/screen/<id> --locales=all` now runs over every committed screen
on the three render legs, and a failing check leaves a picture of itself behind.

## The blocker this had to settle first

Turning the gate on was not a workflow-file change. The committed screen's two golden entries name a
`NumericDisplay` and a `SignalTrace`, the runtime deferred both, and #254's `verification.json`
therefore recorded three `NothingPainted` findings:

```
[VUI101] node 'insufflation-pressure' … 'Bounds':    nothing was drawn there; expected (0, 496, 1280x96)
[VUI101] node 'insufflation-pressure' … 'ColorHash': …; expected rgba(33,184,107,255)
[VUI101] node 'ecg-lead-ii'           … 'Bounds':    nothing was drawn there; expected (0, 592, 1280x128)
```

A gate over that screen is red the day it is added, and the three cheap ways out — delete the
goldens, weaken their checks, verify a different screen — are the three this issue forbids by name.

## ADR-014 decision 5

A `NumericDisplay` and a `SignalTrace` paint the **field** they reserve: the node's whole resolved
rectangle, in the single colour token it carries. The *reading* inside it — the digits a template
expands to, the excursion a waveform makes — stays deferred to #258 and #257.

The rule is read off the artifact rather than invented, and that is the whole argument.
`collectGoldens()` applies ADR-011's predicate while the AST still holds its inputs and writes, per
selected node, that node's whole rectangle and the single token its author gave it. Decision 2 makes
those two values the verifier's expectation, and `goldenBounds()` reads the first as an equality. So
the compiled artifact — in a file four legs byte-compare — already asserts that this node's whole
rectangle carries that tint. A runtime painting nothing there was not declining to invent an
appearance; it was disagreeing with the artifact its own compiler emitted.

It is also the narrowest rule that closes the gap. A `Label`'s box is still not filled with its text
colour and a `Button` still has no face — neither has a golden that says so, and no golden can ever
name the synthetic `Panel` a `Row` produces. A `Clock` carries no token and a `StatusIndicator` one
per state, so neither has a single tint a field could be painted in, which is the same fact that
makes `collectGoldens()` refuse `ColorHash` for such a node.

**One consequence constrains #257 and #258 rather than being free**, so it is stated rather than
discovered: `colorHash()` admits only pixels that are a blend of the node's ground and its tint at
one coverage, so a reading drawn *inside* a field in some third colour fails the golden its own
screen was compiled with. Two ways to show a reading and no others — in the field's own tint, or
knocked out of it back to the ground.

`fieldColorToken()` returns an optional rather than a possibly-empty token, and that is a refusal
rather than a nicety. Absent means "paints no field", a deferral; engaged-and-empty means "paints one
and names no colour", which `validatePayload()` refuses at compile time and `render()` has always
refused at run time. Collapsing them would turn a malformed hand-built screen into a frame that
silently omits a node — which is what the first draft of this change did, and there is now a scenario
pinning it.

## The gate

`mdux_compile_screen()` registers `verify.screen.<id>` beside the `evidence.screen.<id>` it already
registers, so a second committed screen is gated by being committed, and `--no-tests=error` makes
that automatic rather than silent.

Its subject is the **committed** bundle, not the freshly baked one `mdux-verify-bake` renders during
the build. The two agree only because `evidence.screen.<id>` byte-compares them, and a gate that
assumed that agreement would rest on the check it is meant to be independent of.

| Leg | Asserted step | Device |
|---|---|---|
| Linux/GCC 16 | yes | lavapipe |
| Linux/Clang 21 libc++ | yes | lavapipe |
| macOS/Clang 21 | yes | MoltenVK |
| Windows/MSVC | no — runs as an ordinary ctest | Mesa for Windows |

Each asserted step carries the skip guard the pixel legs carry, since exit 77 means an absent device
and nothing else. Both sanitizer legs move the test to their **render** group, or it would create a
Vulkan device outside the leak suppression #253 put there for exactly this reason.

## The diff image

`--diff-image-dir=<dir>` writes `<screen>.<scope>.png` for each render scope that fails: the frame
dimmed, every failed obligation's expected rectangle outlined in magenta, what was actually found in
cyan. CI uploads it on failure.

It is an attachment and never a fifth file in the bundle. It *is* the frame, so decision 4's rule
excludes it — a byte-compared artifact holding a measurement becomes a property of the driver tuple
that produced it. The flag chooses a location, never an expectation: the same checks run and the same
status is returned whether or not it is given. Written inside the render loop, because the readback
is the target's own storage and the next scope overwrites it. A failure to write it is a **warning**,
not an error — whether that file exists says nothing about whether the screen verified.

The PNG encoder is written rather than linked, for ADR-007's zero-SOUP reason: a compression library
in a build tool's dependency graph would need qualifying, for an image no verdict depends on. Its
deflate stream is stored blocks only — legal by RFC 1951 §3.2.4, no Huffman coder, ~3.7 MB for a
1280x720 frame, paid only on a failing run. The tests **decode** it (signature, per-chunk CRC,
inflate, Adler-32, filter bytes, pixels) rather than comparing against a stored byte string: a golden
file would pass for a stream that is self-consistently wrong, and the point of writing a PNG is that
other software reads it.

## Test subjects

- **`tests/verify_ui/fixtures/overdrawn/`** lists a `Panel` after the `SignalTrace` it covers half
  of, so `Bounds` holds while `ColorHash` reports `ForeignColour`. One held and one failed obligation
  in one scope, because a fixture where everything failed could not show that the image marks the
  *failure* rather than the frame.
- **The `ScreenPixelTests` tripwire is gone.** It asserted the golden region was empty and said in
  its own comment that it had to fail the day a golden-eligible component drew. This is that day. Its
  replacement is the rendered golden consumer ADR-012 describes, comparing the frame against the
  numbers the committed sidecar carries rather than against the package the runtime drew from.

That replacement deliberately does **not** scan the ring outside each golden rectangle. This screen
packs `insufflation-pressure` and `ecg-lead-ii` edge to edge and both tokens resolve to the same
value today, so such a scan reports the neighbour's pixels as this node's spill — the attribution
limit `Verify.cppm` states for `goldenBounds()`, met here for the same reason. The "not one pixel
larger" half is the first scenario's job, which paints the whole expected surface and compares every
pixel of it.

## ADR-007 decision 6

Now records two scopes rather than one: four legs for byte-identity, three asserting this gate.
Windows runs it but asserts nothing, so a Windows runner that lost its ICD would report a skip and
stay green on that one check. That asymmetry is a statement about what each leg is for — the four-leg
claim is about *bytes* and Windows carries its full weight there, while this gate is about *pixels*
and the render legs are the ones this repository already treats as owning that question. Making it
four would mean adopting Mesa's Windows build as a **gate**-critical dependency rather than a build
one, which is worth doing in a diff of its own. The consequences paragraph moves with the decision,
since that pairing is what #255 asked for and what drifted once already in #246.

## Verified

Locally on macOS/Clang 21.1.8 + MoltenVK, from a clean build tree:

- `ctest --preset ninja-macos-clang` — **673/673**, including the new `verify` label.
- `mdux-verify-ui --screen=generated/screen/endoscope-monitor --locales=all` → `OK (5 obligations,
  1 render scope)`, exit 0. `verification.json` re-staged via `mdux-bake-update`: three
  `NothingPainted` → `Held`.
- The failure path end to end against the `overdrawn` fixture: exit 1, one PNG written, decoded
  independently with Python's `zlib` — CRCs, extent, filter bytes and pixel values all correct.
- The blank-token scenario mutation-checked: reintroducing the collapse fails it.
- `docs-lint`, `file-headers`, `named-mechanisms`, `schema-drift` and `evidence-lint` all clean.
- `git status --porcelain` empty after a full build — nothing writes into the source tree.

Not verified locally: lavapipe. The artifact records outcomes rather than samples, so it should be
byte-identical across drivers — this PR's own CI is what proves it.

## Regulatory impact

ADR-014 gains decisions 5 and 6; ADR-007 decision 6 and its consequences are updated together. The
committed `verification.json` changes from three `NothingPainted` findings to five `Held` outcomes,
which is a change to evidence content and is why the ADR records the reasoning rather than the
workflow file. The claim the artifact now supports is still bounded: it says the two safety-critical
nodes occupy their declared rectangles in their declared tints, and says nothing about what they will
display — the reading inside a field waits on #257 and #258, and ADR-014's consequences say so.

## Follow-up, not in this PR

`cmake/Cache.cmake` enables `ccache` whenever it is on PATH, and ccache is unsound with this
project's C++20 named modules: it hashes preprocessed source but not imported BMIs, so changing an
exported struct's layout silently serves objects compiled against the old one. It cost two false
diagnoses in this session (a `SEGFAULT` in `mdux-shaderbake`, a bogus `BakeReport::validate()`
rejection in `mdux-mlbake`), and it is very likely the real cause of the long-standing "local macOS
build is red while CI is green" puzzle — every clue fits, including that it survives `rm -rf` of the
build tree. CI is unaffected: runners have no ccache, so the launcher is never set. Worth its own
issue.

## Summary by CodeRabbit

* **New Features**
  * NumericDisplay and SignalTrace fields now render their configured color areas.
  * UI verification can generate PNG diff images highlighting expected and actual regions when checks fail.
  * Added verification coverage for committed screens across supported rendering environments.

* **Bug Fixes**
  * Empty color tokens are now reported as malformed rather than silently deferred.

* **Documentation**
  * Updated verification instructions, implementation status, architecture notes, and roadmap to reflect rendered-screen verification and field rendering.


tools/verify/Driver.cppm
```diff
@@ -118,8 +153,12 @@ struct RunResult {
 /// Test/integration overload with an explicit committed-artifact root (`generated/`).
 [[nodiscard]] RunResult run(const std::filesystem::path& screenDirectory, const std::filesystem::path& artifactRoot);
 
+/// The full form. `options.artifactRoot` must be set; the other members may be empty.
```
ambroise-leclerc: This public overload comment still says `artifactRoot` must be set, but the member contract immediately above explicitly defines an empty root as legal and meaningful: it denotes the current directory and is what the one-argument derivation yields for a relative `screen/<id>` bundle. The second commit says the stale “must be set” contract was corrected, but this declaration retains it. Please change this to state that `artifactRoot` may be empty, so callers are not encouraged to add a guard that rejects supported input.

cmake/MduXCompileScreen.cmake
```diff
@@ -156,4 +156,39 @@ function(mdux_compile_screen)
...
+        COMMAND mdux-verify-ui
+            "--screen=${CMAKE_SOURCE_DIR}/generated/screen/${ARG_ID}"
+            "--locales=all"
+            "--diff-image-dir=${CMAKE_BINARY_DIR}/verify-diff"
+    )
+    # 77 is the driver's absent-device status and nothing else's - every other impossibility exits 3
+    # and fails. A leg that skips this has no Vulkan device, which #254 already made a build failure
+    # for the bake; the CI steps assert on the skip as well, because a leg whose device disappeared
+    # must not report a green tick over a check that never ran.
+    set_tests_properties("verify.screen.${ARG_ID}" PROPERTIES LABELS "verify" SKIP_RETURN_CODE 77)
```
ambroise-leclerc: The configured skip code does not match the executable status. `mdux-verify-ui` maps `RunState::NoRenderDevice` to exit status 3 in `tools/verify/Driver.cppm`, while this test only treats 77 as skipped. I reproduced this on the current head with no usable ICD: `verify.screen.endoscope-monitor` was reported as `***Failed` with VUI008, not `Skipped`. This also contradicts the new ADR/workflow comments saying that Windows can lose its ICD and leave this test green as a skip, and it makes every new `SKIP_RETURN_CODE 77` comment ineffective. Please either map `NoRenderDevice` to 77 at the CLI boundary (and update the exit-status unit scenario/getting-started status table), or deliberately keep status 3 and remove the skip property plus all skip-specific claims. The asserted Linux/macOS legs should retain their explicit skip guard if 77 is chosen.

tools/verify/Driver.cpp
```diff
@@ -245,6 +259,76 @@ readGoldens(const std::filesystem::path& path, std::string& digestOut, std::vect
...
+    // unchanged. Dropping cannot collide two scopes of one screen: `RenderScope` is either the one
+    // locale-free name or a locale from a manifest `validate()` has already refused duplicates in.
+    std::string slug;
+    slug.reserve(scope.size());
+    for (const char character : scope) {
+        if (std::isalnum(static_cast<unsigned char>(character)) != 0 || character == '-' || character == '_' || character == '.') {
+            slug.push_back(character);
+        }
+    }
+    const std::filesystem::path path = directory / std::format("{}.{}.png", screenId, slug);
```
ambroise-leclerc: This slugging is not collision-free for values the current artifact pipeline accepts. `ScreenPackage::validate()` only rejects an empty or exactly duplicated approved locale, and the text/font bakers likewise require non-empty locale strings but do not enforce `[A-Za-z0-9-]`. Consequently two distinct approved scopes such as `en/US` and `enUS` both map to `enUS`, so the second failing render truncates and overwrites the first scope’s PNG. `(locale-free)` and `locale-free` are another collision. That violates the stated one-image-per-failing-render-scope behavior and can hide the useful frame from CI artifacts. Please use a reversible/collision-free filename encoding (for example percent-encoding bytes outside the portable set), append a stable digest of the original scope, or add and test the claimed locale grammar at the artifact boundary.

tools/verify/Driver.cppm
ambroise-leclerc: Fixed in e088f9f. The declaration now reads "every member of `options` may be empty; see `RunOptions` for what each empty value means, and in particular why an empty `artifactRoot` is supported input rather than a mistake." — so it points at the member contract instead of contradicting it.

cmake/MduXCompileScreen.cmake
ambroise-leclerc: You are right, and I reproduced it before changing anything: with the ICD unset, `ctest -L verify` reports `***Failed` with VUI008, not `Skipped`. The `SKIP_RETURN_CODE 77` was dead configuration and six comments were built on it.

I took the second option — keep status 3, remove the property and every skip-specific claim — rather than mapping `NoRenderDevice` to 77. Two reasons. #254 made the bake render, so a leg without a device fails to *build*; I checked that too (`mdux-verify-bake` exits 1 with VUI008 under a bad ICD), which means this test cannot be reached in a tree that has not already proved a device existed. An absent device at test time is therefore a device that disappeared, and #255 asks for that to read as an infrastructure failure rather than a skip. Adding a 77 would also push a CTest convention into a CLI whose 0/1/2/3 statuses are documented in getting-started and asserted in `distinctExitStatuses`.

So in e088f9f: the property is gone, the three `grep Skipped` guards are gone with it (a guard against an unreachable state asserts nothing), and the cmake comment, ADR-014 decision 6 and architecture.md now say why there is no skip. `verify_ui_pixel_test` keeps its 77 — different thing, a test binary returning 77 itself. Re-verified after the change: no device → Failed, device → passes.

tools/verify/Driver.cpp
ambroise-leclerc: Agreed on all three points — the filter is not injective, and `ScreenPackage::validate()` really does refuse only an empty or duplicated locale, so both collisions you name are reachable from an artifact the pipeline accepts.

Fixed in e088f9f with percent-encoding, the first of your three suggestions: reversible, so no two scopes of one screen can collide; `%` is legal in a filename on all three platforms and in an uploaded artifact; and an ordinary tag like `en-US` has nothing to escape and comes out unchanged. The locale-free scope now yields `<screen>.%28locale-free%29.png`, which is uglier but cannot be confused with a locale called `locale-free`.

I moved the rule out of the driver's anonymous namespace into `mdux.tools.verify.diff` as `diffImageName()`, beside the image it names, mainly so it is testable. The new scenario uses your two examples plus a check that no path separator survives, and I confirmed it is not vacuous by mutating the escape back into a drop — it fails.

I did not add the locale grammar at the artifact boundary. That would be a real tightening of what a compiled screen may carry and belongs in its own change against the schema, not smuggled in behind a filename.

AM-L (approved):

PR #285, opened 2026-09-03
Guard release branch topology [#284]

ambroise-leclerc:
## Summary

Add a `Branch Topology` workflow that detects master/develop drift, reruns when a PR is retargeted, and rejects non-release PRs targeting `master`. Update repository and release guidance for the release-only master policy and the green-when-unmergeable trap.

Closes #284

## Invitation and contributor agreement

- [x] This change was requested directly by the repository maintainer in issue #284.
- [ ] Independent contributor CLA acceptance is not applicable to this maintainer-directed agent change.

Invitation / CLA issue: #284

## Dependency

- **Base branch:** `develop`
- **Predecessor PR:** not stacked
- **Merge order:** mergeable now

## Shared registries touched

- [x] None of the listed shared registries

## Rebase state

- **Rebased onto:** `develop` at `180fc01625c07db37e537fdbd24b2e81649e687b`
- **Conflicts resolved as a union:** no conflicts

## Verification

- [x] Branch policy shell logic: aligned ordinary PR and `release/v1.2.3` pass; ordinary master PR and simulated master-only drift fail.
- [x] Workflow YAML parses locally and `Branch Topology` passes in GitHub Actions.
- [x] `mdux-docs-lint`: 83 files, 0 findings.
- [x] `mdux-named-mechanisms`: 79 documents/workflows, 0 findings.
- [x] AI-reference indexes and schema-drift checks pass.
- [ ] Full docs-lint unit discovery: one unrelated local failure because ignored `examples/build/CMakeFiles/.../CMakeCXXCompilerId.cpp` is scanned as source; the pre-existing build tree was preserved.
- [ ] C++ build not run locally: no product code, module, generated artifact, or rendering path changes. PR CI covers all build legs.

## Post-merge gate

- [x] I understand that the `develop` workflow must be green after this merges before the master protection is changed.
- **Post-merge `develop` run:** pending merge
- [ ] That run is green.

## Regulatory impact

Potentially safety-relevant: this changes release-control auditability and CI enforcement, but no governed runtime behavior, compliance metadata, generated evidence, or product API. `docs/release-process.md` is updated proportionately; no ADR, risk record, or SDF content changes.

## Summary by CodeRabbit

* **New Features**
  * Added automated branch-topology checks to enforce release branch rules and branch containment.
  * Added validation requiring pull requests targeting `master` to originate from same-repository `release/vX.Y.Z` branches.

* **Documentation**
  * Updated branch naming and release process guidance to reflect `master` as the default branch.
[...]

Issue #284, opened 2026-09-03
Nothing checks that master carries nothing develop lacks, and an unmergeable PR reports green


ambroise-leclerc:
`docs/release-process.md` step 9 states the invariant:

> The left column must be `0`: `master` carries nothing `develop` lacks. Skipping this step is how
> that number grows.

Nothing checks it. It was violated for two days and nobody noticed.

## What happened

#276 applied "Keep repository guidance tool-neutral" **directly to `master`** on 1 September, while
#275 applied the same work to `develop`. No back-merge followed, so
`git rev-list --left-right --count origin/master...origin/develop` read `2 55`.

The two independent edits to `.gitignore`, `AGENTS.md` and `ADR-003` then conflicted, which made the
v0.7.0 release PR unmergeable.

## The part worth fixing

**An unmergeable PR reports green, not red.** GitHub does not run `pull_request` workflows on a PR
whose merge commit it cannot compute, so PR #283 showed exactly one passing check — a review bot —
and no build, test or lint leg at all. Nothing said "blocked". It looked like a small, healthy PR
that happened to have few checks.

That is the same silent-no-checks failure `AGENTS.md` already warns about for branch naming
("A branch named anything else gets **no checks at all**, silently, which is worse than a failing
one"), reached by a different route, and neither `AGENTS.md` nor `release-process.md` mentions this
one. A release PR is currently the only thing that ever notices the drift, and it notices by
producing a green tick over nothing.

## Suggestion

A cheap workflow assertion, on `push` to `develop` and on `schedule`:

```sh
left=$(git rev-list --left-right --count origin/master...origin/develop | cut -f1)
test "$left" -eq 0 || { echo "::error::master carries $left commit(s) develop lacks; back-merge (release-process.md step 9)"; exit 1; }
```

It needs `fetch-depth: 0` and both refs fetched. Failing on `develop` rather than on the release PR
is the point: the drift is cheap to fix the day it appears and expensive to fix mid-release, when it
arrives disguised as a passing check.

Two secondary things worth deciding with it:

- **Should a direct push to `master` be possible at all?** The topology in `release-process.md` has
  `master` receiving only release PRs. #276 targeted it directly, which is what created the
  divergence.
- **Document the green-when-unmergeable trap**, in `AGENTS.md` beside the branch-naming warning and
  in `release-process.md` step 8. It cost a diagnosis this time; written down it costs a glance.

Found while cutting v0.7.0.

PR #287, opened 2026-09-03
Run the topology check from a trusted definition [#284]

ambroise-leclerc:
## Summary

`branch-topology.yml` polices the pull request it runs on, and on `pull_request` GitHub evaluates
the workflow from the merge commit — the head branch's copy. A pull request targeting `master`
could therefore delete the release-branch rule in the same diff that rule exists to reject: the
guard was enforced by a file the author being guarded controls. `pull_request_target` loads the
definition from the base branch instead, out of the author's reach until merged.

The event guards now match `pull_request*` rather than one exact name. Under `pull_request_target`
an exact `pull_request` test would skip the `master` rule silently, which is the failure the
trigger swap invites.

The trigger is safe here only because the job reads branch tips and event metadata and never checks
out or executes a pull request's code, so the checkout keeps its default base-branch ref. AGENTS.md
records that constraint, the resulting rule that edits to this workflow are checked by the previous
version, and that a green `Branch Topology` proves nothing about the build — it needs no merge
commit, so it is the one required check that reports green while the other eleven sit pending on an
unmergeable release PR.

Required status checks on `master` do not close this on their own: a pull request that keeps the
job name and empties the script reports present and green.

Raised by CodeRabbit on #285 (CWE-829, unresolved at merge). Follow-up to #284.

Invitation / CLA issue: #284

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [x] None of the above

## Rebase state

- **Rebased onto:** `9d53e3a42c04f42b65ec5dcb87c31eabed965910`
- **Conflicts resolved as a union:** no conflicts. The rebase over #286 applied cleanly because the
  AGENTS.md change is an append; #286's rewritten paragraph was verified intact afterwards rather
  than assumed from the clean exit.

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run; workflow and documentation only
- [ ] `ctest --test-dir build-gcc` — not run; workflow and documentation only
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 83 files checked, 0 findings
- [x] Other: `check_named_mechanisms.py` — 79 documents/workflows; `git diff --check` clean; workflow
      parses under `yaml.safe_load`

The policy script was extracted from the parsed workflow and run locally against this repository for
every event path, since the trusted trigger cannot be exercised by the pull request that introduces
it — the check reported here runs `develop`'s current definition, not this diff's:

| event | base | head | result |
| --- | --- | --- | --- |
| `push` / `schedule` | — | — | exit 0, drift `0 3` |
| `pull_request_target` | `master` | own `release/v1.2.3` | exit 0 |
| `pull_request_target` | `master` | own `284-foo` | exit 1, rejected |
| `pull_request_target` | `master` | fork `release/v1.2.3` | exit 1, rejected |
| `pull_request_target` | `develop` | own `284-foo` | exit 0 |
| `pull_request` (legacy name) | `master` | own `284-foo` | exit 1, still caught |

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

None to runtime behaviour, governed API, compliance metadata, generated evidence, or any claim of
conformity. It does harden a release control: the rule restricting `master` to same-repository
`release/vX.Y.Z` branches can no longer be removed by the pull request it applies to. The ISO 13485
and IEC 62304 documents that cite branch protection as an enforcing mechanism are unchanged and
remain accurate.

### DIFF
--- a/.github/workflows/branch-topology.yml
+++ b/.github/workflows/branch-topology.yml
@@ -5,10 +5,16 @@ on:
     # A push to master is what creates drift, so observe it directly rather than waiting for the
     # cron: a release merged at 06:00 would otherwise go unverified until the next morning.
     branches: [ master, develop ]
-  pull_request:
-    # A pull request gets only the release-branch-only rule below; the drift leg belongs to `push`
-    # and `schedule`, because it measures branch tips no pull request can move. The filter stays
-    # wide so ordinary and stacked PRs still report a check. `edited` covers base retargeting.
+  pull_request_target:
+    # `pull_request_target`, not `pull_request`, because this workflow polices the pull request it
+    # runs on. On `pull_request` GitHub evaluates the definition from the merge commit — the head
+    # branch's copy — so a PR targeting master could edit this file in its own diff and delete the
+    # rule below. `pull_request_target` loads the definition from the base branch, which the author
+    # cannot reach without first merging. Two consequences worth knowing: the job never checks out
+    # or runs any pull request's code (it reads only the branch tips and the event metadata, and
+    # `permissions` stays read-only), which is what makes the trigger safe here; and a change to
+    # this file takes effect only once it is on the base branch, so a PR editing it is still
+    # checked by the previous version.
     branches: [ master, develop, '[0-9]+-*', 'feat/**' ]
     types: [ opened, synchronize, reopened, edited ]
   schedule:
@@ -27,6 +33,9 @@ jobs:
 
     steps:
     - name: Checkout repository history
+      # No `ref:`, deliberately. Under `pull_request_target` the default is the base branch, and
+      # pointing it at the pull request's head would check out author-controlled content into a
+      # trusted context — the hazard that makes this trigger worth warning about.
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         fetch-depth: 0
@@ -49,10 +58,10 @@ jobs:
         failed=0
 
         # Drift belongs to the branch tips, not to any one pull request: the counts below are the
-        # same whatever a PR contains. Measured on `pull_request`, this leg would redden every open
+        # same whatever a PR contains. Measured on a pull request, this leg would redden every open
         # PR for the window between a release merge and the back-merge of step 9 — demanding of
         # each contributor an admin push to develop that none of them can make.
-        if [[ "${EVENT_NAME}" != "pull_request" ]]; then
+        if [[ "${EVENT_NAME}" != pull_request* ]]; then
           # Capture the counts before splitting them. `read x y <<< "$(cmd)"` takes its exit status
           # from `read`, never from the substitution, so a `git rev-list` that fails — an unfetched
           # ref, a renamed branch — would leave both variables empty, and `(( master_only != 0 ))`
@@ -72,7 +81,10 @@ jobs:
           fi
         fi
 
-        if [[ "${EVENT_NAME}" == "pull_request" && "${BASE_REF}" == "master" ]]; then
+        # Matches `pull_request_target` and `pull_request` alike: only the former is configured
+        # above, but an exact test for one event name is how this rule silently stops running if
+        # the trigger is ever changed back.
+        if [[ "${EVENT_NAME}" == pull_request* && "${BASE_REF}" == "master" ]]; then
           if [[ "${HEAD_REPOSITORY}" != "${GITHUB_REPOSITORY}" || \
                 ! "${HEAD_REF}" =~ ^release/v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
             echo "::error::master accepts only same-repository release/vX.Y.Z pull requests; target develop for ordinary work"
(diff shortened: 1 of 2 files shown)


.github/workflows/branch-topology.yml
```diff
@@ -5,10 +5,16 @@ on:
     # A push to master is what creates drift, so observe it directly rather than waiting for the
     # cron: a release merged at 06:00 would otherwise go unverified until the next morning.
     branches: [ master, develop ]
-  pull_request:
-    # A pull request gets only the release-branch-only rule below; the drift leg belongs to `push`
-    # and `schedule`, because it measures branch tips no pull request can move. The filter stays
-    # wide so ordinary and stacked PRs still report a check. `edited` covers base retargeting.
+  pull_request_target:
```
ambroise-leclerc, line 16: The trigger migration itself currently falls into a coverage gap. GitHub reports zero `Branch Topology` runs for this PR head (`34e8040`), whereas the preceding PR head (`f8931bf`) has two `pull_request` runs. The head revision has removed `pull_request`, while the trusted default-branch revision does not yet declare `pull_request_target`, so neither event definition launches this workflow for PR #287.

That contradicts the statement below that this edit is checked by the previous definition, and it leaves the security-sensitive event change without live event verification before merge. Please stage the migration: first add `pull_request_target` while retaining `pull_request`, let that trusted definition land, then remove `pull_request` in a follow-up that will itself be checked through `pull_request_target`. If a two-stage migration is intentionally declined, the one-time gap and a concrete post-merge validation of the `master` rejection path should be recorded explicitly.

AGENTS.md
```diff
@@ -292,7 +292,15 @@ easiest to miss on review. So:
...
 - A branch named anything else (`fix-typo`, `wip`, `my-feature`) gets **no CI on a PR based on
   it**. If you need one, add its pattern to the `branches:` list of every workflow under
-  `.github/workflows/` that has a `pull_request:` trigger.
+  `.github/workflows/` that has a `pull_request:` or `pull_request_target:` trigger.
+
+`branch-topology.yml` is the one workflow on `pull_request_target` rather than `pull_request`,
+because it is the one that polices the pull request it runs on: `pull_request` would evaluate the
+head branch's copy of the definition, so a PR targeting `master` could delete the rule in the same
+diff the rule exists to reject. The practical consequence when editing that file is that **your
+change is checked by the base branch's version, not yours** — the new definition only takes effect
```
ambroise-leclerc: GitHub documents `pull_request_target` as running in the context of the **default branch of the base repository**, with `GITHUB_SHA` and `GITHUB_REF` pointing to that default branch. In this repository that is `develop`; it is not necessarily the pull request base branch. Therefore, for a release PR targeting `master`, and for a stacked PR targeting an issue branch, the trusted workflow and default checkout come from `develop`, not from `master` or the predecessor branch.

Please replace “base branch” with “default branch” here and in the workflow comments (including the checkout comment). This distinction affects when workflow edits on stacked/base branches become active and is important future-maintainer guidance for a privileged trigger. The security conclusion remains sound—the head revision is not executed—but it is sound because the default-branch definition is trusted.

AGENTS.md
```diff
@@ -276,7 +276,7 @@ issue number, a dash, then the slugified issue title.
 14-text-schema
 ```
 
-This is not cosmetic. Every workflow filters `pull_request` on the pattern `[0-9]+-*`, and
+This is not cosmetic. Every workflow filters pull requests on the pattern `[0-9]+-*`, and
```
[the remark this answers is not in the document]
ambroise-leclerc: Confirmed and reworded in a7c90e7. `scorecard.yml` triggers on `push`, `schedule` and `branch_protection_rule` only — it is the sole workflow in `.github/workflows/` with no PR trigger, which I checked by parsing all of them rather than by eye. The sentence now reads "Every workflow that runs on pull requests filters them on the pattern `[0-9]+-*` — Scorecard is the one exception, triggering only on pushes and a schedule".

.github/workflows/branch-topology.yml
ambroise-leclerc: Confirmed, and staged exactly as you describe in a7c90e7. `pull_request` is retained alongside `pull_request_target`; stage two removes it and will itself be checked by the trusted definition. Your diagnosis of the mechanism is right: GitHub decides whether to run a `pull_request_target` workflow from the definition on the default branch, which does not declare that trigger until this merges, so the head revision had removed the only event that could fire. `Branch Topology` now reports on this PR (run 33807622099, pass in 7s), so the trigger change no longer merges without live event verification. The AGENTS.md sentence that contradicted this now states the two-stage sequence explicitly rather than claiming the edit is checked by the previous definition.

AGENTS.md
ambroise-leclerc: Verified against the GitHub documentation and you are right — I had this wrong. `pull_request_target` "runs in the context of the default branch of the base repository", with `GITHUB_REF` the default branch and `GITHUB_SHA` its last commit; not the pull request base. Corrected in a7c90e7 in AGENTS.md, in the trigger comment and in the checkout comment. I also recorded the consequence you draw out, because it is what makes the conclusion sound rather than incidental: a release PR to `master` and a stacked PR to an issue branch are both checked by `develop`s copy, not by a branch their author controls. One addition — the checkout comment now notes that the fetch names `master` and `develop` explicitly, so the comparison never depends on which ref was checked out.

AM-L (approved):

PR #289, opened 2026-09-03
Make ccache BMI-aware for Clang modules [#280]

### LINKED ISSUE #280: ccache silently corrupts C++20 module builds, and its symptoms look like a host miscompilation
`cmake/Cache.cmake` sets `CMAKE_CXX_COMPILER_LAUNCHER=ccache` whenever ccache is found on PATH.
ccache is **not sound for this project's C++20 named modules**: it hashes the preprocessed source but
not the BMIs the translation unit imports, so changing an exported struct's layout silently serves an
object compiled against the *old* layout.

The symptoms are ODR-mismatch symptoms, and none of them names the cause.

## Observed

Both of these appeared on a clean local build of `develop` and were cured by `ccache -C`:

- `mdux-shaderbake` segfaulting in `BakeReport::validate()`, reading a `std::vector<ToolStage>` that
  did not exist in the layout its caller was compiled against (the member #254 added).
- `mdux-mlbake` failing with `bake report is not valid: a production stage names an output the report
  does not list`, over a `stages` vector that is empty in the source.

## Why it is hard to recognise

Every clue points away from the cache:

- **It survives `rm -rf <build dir>`** — the stale objects are in `~/.ccache`, not the build tree.
- **`touch` does not dislodge it; only a real content change does** — a content change moves the
  ccache hash, a timestamp does not.
- **Adding a `std::cerr` line "fixes" it** — that is a cache miss, hence a correct compile.
- **CI is green** — runners have no ccache, so `CACHE_BINARY` is never found and the launcher is
  never set.

Taken together those read as a host-specific miscompilation. This repository's own notes carried
exactly that conclusion for a week ("local macOS red while CI is green, therefore macOS 26 vs the
verified macOS 15 lane"), including two incidents that now fit this cause precisely: three
`medui_spec` scenarios segfaulting over generated screens, and five `GoldenTests` resolving a
diagnostic code to the previous registry row after an enumerator was added.

## Options

1. **Default `ENABLE_CACHE` to `OFF`.** Simplest and safest; costs incremental rebuild time for
   contributors who opt back in knowingly.
2. **Keep it on and make it correct** — `CCACHE_DEPEND=1` (depend mode) and/or an explicit
   `sloppiness` set, if that is actually sufficient for CMake's explicit-module builds. This needs
   verifying rather than assuming: the failure mode is silent, so "it seems fine now" is not evidence.
3. **Refuse it where it is unsafe**: leave the option, but disable the launcher when
   `CMAKE_CXX_MODULE_STD`/`FILE_SET CXX_MODULES` targets are configured.

Whichever is chosen, the check that matters is a *negative* one: change an exported struct's layout in
a `.cppm`, rebuild without clearing the cache, and confirm the result is correct rather than merely
green.

## Impact

CI is unaffected. This is a local-development correctness issue, and its cost is entirely in
misdirected debugging — it produced two false diagnoses in the session that landed #255, and probably
several days of earlier archaeology.

Found while implementing #255.

ambroise-leclerc:
## Summary

Keep compiler caching enabled while making it sound for MduX's C++20 named-module builds.

The compiler launcher now classifies module commands before invoking the selected cache:

- module providers compile directly because a cache hit cannot restore their BMI side output;
- Clang consumers using ccache 4.8+ remain cacheable, with every imported BMI supplied through
  ccache's per-compilation `extra_files_to_hash` setting;
- unknown module command formats, older ccache versions, and other cache programs fail closed to a
  direct compile, while ordinary translation units remain cache-eligible.

The new negative integration test changes an exported struct layout without touching its consumer
and verifies that the rebuilt executable observes the new layout. It then touches the unchanged
consumer and requires a real ccache hit, proving that correctness was not obtained by disabling
consumer caching. The Linux Clang workflow installs ccache 4.8+ and runs this test explicitly.

This follows ccache's still-open named-module limitation
(https://github.com/ccache/ccache/issues/1252) while retaining safe incremental-build acceleration.

Closes #280

Invitation / CLA issue: #280

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [ ] `FILE_SET CXX_MODULES` lists — a module interface added, removed or moved
- [ ] None of the above

## Rebase state

- **Rebased onto:** `b0b0bbc39bc6df8ae6a81ccafcc13b2ffc321b11`

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — used an equivalent fresh out-of-tree
      GCC configuration under `/tmp/mdux-280-gcc`; configure and all 595 build steps passed
- [x] `ctest --test-dir /tmp/mdux-280-gcc --output-on-failure` — 676/676 passed
- [x] `python3 tools/docs-lint/mdux_docs_lint.py` — 83 files checked, 0 findings
- [x] Other:
  - standalone Clang/ccache layout-change integrity test passed, including the required subsequent
    ccache hit;
  - Clang-configured `build.cache.classifier` and `build.cache.namedModuleIntegrity`: 2/2 passed;
  - GCC-configured `build.cache.classifier`: 1/1 passed;
  - `check_named_mechanisms.py`: 79 documents/workflows checked;
  - `git diff --check` clean;
  - `clang-build.yml` parses as YAML.

The repository's exact `ninja-clang` preset could not configure on this host because its required
`/usr/lib/llvm-21/lib/libc++.modules.json` manifest is absent. An alternate Clang 21/libstdc++
configuration exercised both cache tests successfully. Its wider `MduXCore` build reached step
74/76 before the existing 4096-byte stack-frame guard rejected `Compliance.cpp` under that
unsupported standard-library combination; the supported Clang/libc++ lane remains covered by CI.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

None — this changes compiler-cache orchestration, tests, CI, and build documentation only. It does
not alter runtime behaviour, governed APIs, risk controls, compliance metadata, traceability,
generated evidence, or any claim about a medical-device standard.

## Summary by CodeRabbit

* **New Features**
  * Improved compiler-cache support for C++20 named modules and imported module artifacts.
  * Added options to enable caching and select a custom cache executable.
  * Added validation for cache availability and ccache version compatibility.
  * Cache bypasses unsafe module-related compilations while continuing to cache eligible translation units.

* **Documentation**
  * Documented compiler-cache configuration, version requirements, and named-module behavior.

* **Tests**
  * Added coverage for cache classification, module integrity, and correct rebuilds after module changes.

### DIFF
--- a/.github/workflows/clang-build.yml
+++ b/.github/workflows/clang-build.yml
@@ -108,6 +108,7 @@ jobs:
         sudo apt-get install -y \
           ninja-build \
           build-essential \
+          ccache \
           clang-21 \
           libc++-21-dev \
           libc++abi-21-dev \
@@ -137,6 +138,11 @@ jobs:
         major="$("$clangxx" --version | sed -n 's/.*clang version \([0-9]*\).*/\1/p' | head -1)"
         if [ "$major" != "21" ]; then
           echo "::error::${clangxx} reports major version ${major}, expected 21."
+          exit 1
+        fi
+        ccache_version="$(ccache --version | sed -n 's/^ccache version \([0-9][0-9.]*\).*/\1/p' | head -1)"
+        if ! dpkg --compare-versions "$ccache_version" ge 4.8; then
+          echo "::error::ccache ${ccache_version:-unknown} cannot accept per-compilation BMI inputs; 4.8+ is required."
           exit 1
         fi
 
@@ -181,6 +187,14 @@ jobs:
         export DISPLAY=:99
         Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
         ctest --preset ninja-clang --output-on-failure
+
+    # #280's negative case changes an exported struct without touching its consumer, then proves
+    # that the new BMI invalidates the cached object. The second rebuild also has to be a cache hit:
+    # bypassing every module consumer would be correct but would throw away the compile-time benefit
+    # this launcher exists to preserve. Kept as a named step with --no-tests=error so losing the
+    # registration cannot masquerade as a green run.
+    - name: Verify ccache invalidates named-module consumers
+      run: ctest --preset ninja-clang -R '^build\.cache\.namedModuleIntegrity$' -V --no-tests=error
 
     # The step this workflow's header used to describe as future work. It matches the MSVC and GCC
     # legs in windows-build.yml / linux-gcc16-build.yml: re-derive every committed artifact and
--- a/cmake/Cache.cmake
+++ b/cmake/Cache.cmake
@@ -1,29 +1,82 @@
-option(ENABLE_CACHE "Enable cache if available" ON)
-if(NOT ENABLE_CACHE)
-  return()
-endif()
-
+option(ENABLE_CACHE "Enable a compiler cache if available" ON)
 set(CACHE_OPTION
     "ccache"
     CACHE STRING "Compiler cache to be used")
+set(CACHE_BINARY "" CACHE FILEPATH
+    "Compiler cache executable override; empty searches for CACHE_OPTION")
 set(CACHE_OPTION_VALUES "ccache" "sccache")
 set_property(CACHE CACHE_OPTION PROPERTY STRINGS ${CACHE_OPTION_VALUES})
-list(
-  FIND
-  CACHE_OPTION_VALUES
-  ${CACHE_OPTION}
-  CACHE_OPTION_INDEX)
+list(FIND CACHE_OPTION_VALUES "${CACHE_OPTION}" CACHE_OPTION_INDEX)
 
-if(${CACHE_OPTION_INDEX} EQUAL -1)
-  message(
-    STATUS
-      "Using custom compiler cache system: '${CACHE_OPTION}', explicitly supported entries are ${CACHE_OPTION_VALUES}")
+# These cache entries are the explicit contract consumed by tests/CMakeLists.txt. Reset them on
+# every configure so normal early-return paths cannot leave stale capability state behind.
+set(MDUX_CACHE_BINARY "" CACHE INTERNAL "Resolved compiler cache executable" FORCE)
+set(MDUX_CACHE_KIND "disabled" CACHE INTERNAL "Resolved compiler cache kind" FORCE)
+set(MDUX_CACHE_CAN_HASH_BMIS OFF CACHE INTERNAL
+    "Whether the compiler cache accepts per-compilation BMI inputs" FORCE)
+
+if(NOT ENABLE_CACHE)
+    return()
 endif()
 
-find_program(CACHE_BINARY ${CACHE_OPTION})
+if(CACHE_OPTION_INDEX EQUAL -1)
+    message(STATUS
+        "Using custom compiler cache system: '${CACHE_OPTION}', explicitly supported entries are ${CACHE_OPTION_VALUES}")
+endif()
+
 if(CACHE_BINARY)
-  message(STATUS "${CACHE_OPTION} found and enabled")
-  set(CMAKE_CXX_COMPILER_LAUNCHER ${CACHE_BINARY})
+    set(mdux_resolved_cache_binary "${CACHE_BINARY}")
 else()
-  message(WARNING "${CACHE_OPTION} is enabled but was not found. Not using it")
+    find_program(mdux_resolved_cache_binary NAMES "${CACHE_OPTION}" NO_CACHE)
+endif()
+if(NOT mdux_resolved_cache_binary)
+    message(WARNING "${CACHE_OPTION} is enabled but was not found. Not using it")
+    return()
+endif()
+set(MDUX_CACHE_BINARY "${mdux_resolved_cache_binary}"
+    CACHE INTERNAL "Resolved compiler cache executable" FORCE)
+
+# Ccache still does not understand C++20 named-module dependency state. In particular, neither
+# depend_mode nor sloppiness=modules adds imported BMIs to an object's key. The launcher below
+# therefore compiles module providers directly and, on Clang, hashes every imported BMI before
+# caching a consumer. Other named-module command formats fail closed to a direct compilation;
+# ordinary translation units remain eligible for the selected cache.
+set(MDUX_CACHE_KIND "other" CACHE INTERNAL "Resolved compiler cache kind" FORCE)
+execute_process(
+    COMMAND "${MDUX_CACHE_BINARY}" --version
+    RESULT_VARIABLE mdux_cache_version_result
+    OUTPUT_VARIABLE mdux_cache_version_stdout
+    ERROR_VARIABLE mdux_cache_version_stderr
+)
+string(CONCAT mdux_cache_version_banner
+    "${mdux_cache_version_stdout}" "\n" "${mdux_cache_version_stderr}")
+if(mdux_cache_version_result EQUAL 0 AND
+   mdux_cache_version_banner MATCHES "(^|\n)ccache version ([0-9]+(\\.[0-9]+)+)")
+    set(MDUX_CACHE_KIND "ccache" CACHE INTERNAL "Resolved compiler cache kind" FORCE)
+    set(mdux_cache_version "${CMAKE_MATCH_2}")
+    # Per-compilation configuration, used for extra_files_to_hash, arrived in ccache 4.8.
+    if(mdux_cache_version VERSION_GREATER_EQUAL "4.8")
+        set(MDUX_CACHE_CAN_HASH_BMIS ON CACHE INTERNAL
+            "Whether the compiler cache accepts per-compilation BMI inputs" FORCE)
+    endif()
+endif()
+
+set(CMAKE_CXX_COMPILER_LAUNCHER
+    "${CMAKE_COMMAND}"
+    "-DMDUX_CACHE_BINARY=${MDUX_CACHE_BINARY}"
+    "-DMDUX_CACHE_KIND=${MDUX_CACHE_KIND}"
+    "-DMDUX_CACHE_COMPILER_ID=${CMAKE_CXX_COMPILER_ID}"
+    "-DMDUX_CACHE_CAN_HASH_BMIS=${MDUX_CACHE_CAN_HASH_BMIS}"
+    -P "${CMAKE_CURRENT_LIST_DIR}/MduXCompilerCacheLauncher.cmake"
+    --
+)
+
+if(MDUX_CACHE_KIND STREQUAL "ccache" AND
+   CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
+   MDUX_CACHE_CAN_HASH_BMIS)
+    message(STATUS
+        "ccache ${mdux_cache_version} found: module providers compile directly; Clang module consumers hash imported BMIs; other compilations remain cache-eligible")
+else()
+    message(STATUS
+        "${CACHE_OPTION} found: named-module compilations bypass the cache; other compilations remain cache-eligible")
 endif()
--- a/cmake/MduXCompilerCache.cmake
+++ b/cmake/MduXCompilerCache.cmake
@@ -1,0 +1,98 @@
+include_guard(GLOBAL)
+
+# Classify a Clang CMake module-map response file. Only a list composed entirely of imported BMI
+# arguments is cacheable. Provider output or any syntax this project does not understand fails
+# closed, because a missed cache opportunity is preferable to an object built against a stale BMI.
+function(mdux_analyze_clang_module_map content base_directory out_cacheable out_bmis)
+    string(STRIP "${content}" stripped_content)
+    if(stripped_content STREQUAL "")
+        set("${out_cacheable}" TRUE PARENT_SCOPE)
+        set("${out_bmis}" "" PARENT_SCOPE)
+        return()
+    endif()
+
+    separate_arguments(module_arguments NATIVE_COMMAND "${stripped_content}")
+    set(imported_bmis)
+    foreach(argument IN LISTS module_arguments)
+        if(NOT argument MATCHES "^-fmodule-file=(.+)$")
+            set("${out_cacheable}" FALSE PARENT_SCOPE)
+            set("${out_bmis}" "" PARENT_SCOPE)
+            return()
+        endif()
+
+        set(module_file "${CMAKE_MATCH_1}")
+        if(module_file MATCHES "^[^=]+=(.+)$")
+            set(module_file "${CMAKE_MATCH_1}")
+        endif()
+        if(NOT IS_ABSOLUTE "${module_file}")
+            get_filename_component(module_file
+                "${module_file}" ABSOLUTE BASE_DIR "${base_directory}")
+        endif()
+        cmake_path(NORMAL_PATH module_file)
+
+        if(NOT EXISTS "${module_file}" OR
+           (NOT WIN32 AND module_file MATCHES ":"))
+            set("${out_cacheable}" FALSE PARENT_SCOPE)
+            set("${out_bmis}" "" PARENT_SCOPE)
+            return()
+        endif()
+        list(APPEND imported_bmis "${module_file}")
+    endforeach()
+
+    if(NOT imported_bmis)
+        set("${out_cacheable}" FALSE PARENT_SCOPE)
+        set("${out_bmis}" "" PARENT_SCOPE)
+        return()
+    endif()
+
+    list(REMOVE_DUPLICATES imported_bmis)
+    set("${out_cacheable}" TRUE PARENT_SCOPE)
+    set("${out_bmis}" "${imported_bmis}" PARENT_SCOPE)
+endfunction()
+
+# Encode ccache's platform-specific path list. The launcher bracket-quotes the completed setting,
+# so a Windows semicolon remains inside one process argument rather than becoming a CMake list.
+function(mdux_join_ccache_extra_files paths use_windows_separator out_argument)
+    if(use_windows_separator)
+        set(path_separator ";")
+    else()
+        set(path_separator ":")
+    endif()
+    list(JOIN paths "${path_separator}" extra_files)
+    set("${out_argument}" "${extra_files}" PARENT_SCOPE)
+endfunction()
+
+# Append one exact process argument as a CMake bracket argument. The delimiter grows until its
+# closing token is absent from the value, preserving semicolons, quotes, newlines, and empty argv
+# entries when the resulting execute_process call is evaluated.
+function(mdux_append_bracket_argument code argument out_code)
+    set(equals)
+    while(TRUE)
+        set(terminator "]${equals}]")
+        string(FIND "${argument}" "${terminator}" terminator_index)
+        if(terminator_index EQUAL -1)
+            break()
+        endif()
+        string(APPEND equals "=")
+    endwhile()
+    string(APPEND code " [${equals}[${argument}]${equals}]")
+    set("${out_code}" "${code}" PARENT_SCOPE)
+endfunction()
+
+# Identify arguments that either produce module state or can hide/use module state the launcher
+# cannot safely add to a cache key. A recognized CMake .modmap is parsed separately by the caller.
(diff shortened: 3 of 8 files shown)


cmake/MduXCompilerCacheLauncher.cmake
```diff
@@ -0,0 +1,112 @@
...
+    if(imported_bmis)
+        list(REMOVE_DUPLICATES imported_bmis)
+        # MduX supports upstream Clang only on Unix-like hosts, where ccache's path-list separator
+        # is ':'. A colon in a BMI path was rejected by the parser above rather than encoded
+        # ambiguously here.
+        list(JOIN imported_bmis ":" imported_bmi_argument)
+        list(APPEND command "extra_files_to_hash=${imported_bmi_argument}")
+    endif()
+    list(APPEND command ${compiler_command})
+endif()
```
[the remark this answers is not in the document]
ambroise-leclerc: Corrigé dans 5b4cf4f : le lanceur encode désormais la liste avec `;` sous Windows (échappé uniquement pour préserver un argument CMake unique) et `:` ailleurs. Le test du classifieur couvre les deux séparateurs et vérifie explicitement que la forme Windows reste un seul argument ; les tests ciblés classifieur et intégrité ccache passent.

cmake/MduXCompilerCacheLauncher.cmake
```diff
@@ -0,0 +1,112 @@
...
+    message(FATAL_ERROR
+        "MduXCompilerCacheLauncher.cmake: missing compiler command after --")
+endif()
+
+set(compiler_command)
+set(bypass_cache OFF)
+set(imported_bmis)
+foreach(index RANGE ${command_start} ${last_argument_index})
+    set(argument "${CMAKE_ARGV${index}}")
+    list(APPEND compiler_command "${argument}")
```
ambroise-leclerc: **medium — compiler arguments containing `;` are silently split, and empty arguments are dropped.**

Each argv entry is stored as a CMake *list element* here, and `execute_process(COMMAND ${command})` at line 104 re-splits every element on unescaped `;`. Verified against this launcher:

- passing `-DX=a;b -c` delivers `<-DX=a>` `<b>` `<-c>` to the child — three arguments instead of two;
- passing `x '' y` delivers only `<y>` — the empty argument and everything positional around it is lost.

The previous arrangement (`CMAKE_CXX_COMPILER_LAUNCHER=ccache`, expanded by Ninja) preserved both.

This is not hypothetical for this repo: any `target_compile_definitions` / `target_compile_options` value or include path containing a semicolon — from MduX itself or from a CPM dependency (SpecLab, Vulkan-Headers) — now reaches the compiler as two arguments, and either errors out or, worse, mis-parses into something that compiles differently. Note the blast radius is wider than named modules: this hits *every* compilation routed through the launcher, including the ordinary TUs the design deliberately keeps cache-eligible.

Fix: escape before appending, e.g.

```cmake
string(REPLACE ";" "\\;" escaped_argument "${argument}")
list(APPEND compiler_command "${escaped_argument}")
```

and keep the same treatment for the `extra_files_to_hash=` element. A test in `CompilerCacheClassifierTest.cmake` asserting argv round-trip for `a;b` and for an empty argument would pin this down.

cmake/MduXCompilerCacheLauncher.cmake
```diff
@@ -0,0 +1,112 @@
...
+
+    # Module providers produce a BMI in addition to their object. Ccache cannot restore that
+    # additional output on a cache hit, so providers always compile directly.
+    if(argument MATCHES "\\.(cppm|ixx|mpp)$" OR
```
ambroise-leclerc: **medium — the "fail closed" policy is fail-*open* for module information the launcher never sees in argv.**

Bypass and BMI hashing are both decided by pattern-matching visible arguments. That is only sound if every module-affecting input is visible. Two concrete holes:

**(a) `-fprebuilt-module-path=<dir>`.** This is a real, supported Clang way to supply BMIs, and it appears in neither the bypass patterns here (lines 39-44), nor the second bypass block (lines 81-82), nor `mdux_analyze_clang_module_map`. A consumer built that way is handed straight to ccache with **no BMI in its key** — which is exactly the #280 corruption this PR exists to prevent, reintroduced through a different flag spelling. CMake does not emit it today, but the whole point of the surrounding comments is not to depend on that.

**(b) Response files.** I confirmed that with `CMAKE_NINJA_FORCE_RESPONSE_FILE=ON` (and automatically on Windows once the command line overflows), CMake emits `rspfile_content = $DEFINES $INCLUDES $FLAGS` and the compile line becomes:

```
clang++ @$RSP_FILE ... @$DYNDEP_MODULE_MAP_FILE ...
```

The launcher never opens `@*.rsp`, so any module-affecting flag a target or dependency puts in `$FLAGS` becomes invisible: it neither triggers a bypass nor gets hashed. The `.modmap` handling still works because that file is matched by name, which makes the failure quiet rather than obvious.

Suggested fix, in order of preference:

1. Invert the default — treat a TU as cacheable only when *every* argument is recognized as cache-neutral. That matches the comment at lines 3-6 ("fail closed") and turns unknown future flags into a lost cache hit rather than a corrupt object.
2. Failing that, at minimum add `-fprebuilt-module-path`, `-fmodules-cache-path`, `-fmodule-map-file` to the bypass set, and bypass on any `@` response file that is not the recognized `.modmap`.

cmake/MduXCompilerCacheLauncher.cmake
```diff
@@ -0,0 +1,112 @@
...
+        # setting in one CMake command-list element; execute_process passes a literal semicolon.
+        mdux_join_ccache_extra_files(
+            "${imported_bmis}" "${WIN32}" imported_bmi_argument)
+        list(APPEND command "extra_files_to_hash=${imported_bmi_argument}")
+    endif()
+    list(APPEND command ${compiler_command})
+endif()
+
+execute_process(COMMAND ${command} RESULT_VARIABLE command_result)
+if(command_result MATCHES "^[0-9]+$" AND command_result LESS_EQUAL 125)
```
ambroise-leclerc: **low — compiler exit codes 126-255 are collapsed to `1`.**

Verified: a child exiting 130 makes the launcher exit 1. Ninja still fails the build, so nothing is silently accepted, but the real status is lost — a wrapper or toolchain that distinguishes 126 (found but not executable), 127 (not found), or 128+N (killed by signal, when reported numerically) now reports a generic failure.

That is precisely the class of diagnostic #280 needed: the original bug was hard to find because the symptom did not look like its cause, and this flattens a second such signal. A compiler killed by SIGKILL (OOM on a big module build) is worth telling apart from a compile error.

`cmake_language(EXIT)` accepts 0-255, so the `LESS_EQUAL 125` cap has no apparent justification — was it carried over from `xargs` semantics? Suggest:

```cmake
if(command_result MATCHES "^[0-9]+$" AND command_result LESS_EQUAL 255)
    cmake_language(EXIT ${command_result})
```

with the existing `FATAL_ERROR` branch still catching the non-numeric "could not start" case.

cmake/Cache.cmake
```diff
@@ -1,29 +1,67 @@
...
+set(CACHE_OPTION_VALUES "ccache" "sccache")
+set_property(CACHE CACHE_OPTION PROPERTY STRINGS ${CACHE_OPTION_VALUES})
+list(FIND CACHE_OPTION_VALUES "${CACHE_OPTION}" CACHE_OPTION_INDEX)
+
+if(CACHE_OPTION_INDEX EQUAL -1)
+    message(STATUS
+        "Using custom compiler cache system: '${CACHE_OPTION}', explicitly supported entries are ${CACHE_OPTION_VALUES}")
+endif()
+
+find_program(CACHE_BINARY NAMES "${CACHE_OPTION}" NO_CACHE)
```
ambroise-leclerc: **low — `NO_CACHE` silently disables the `-DCACHE_BINARY=...` override.**

`NO_CACHE` stores the result in a *normal* variable, which shadows any cache entry of the same name. Before this change `CACHE_BINARY` was a cache variable, so a user or a CI lane could pin a specific binary with `-DCACHE_BINARY=/opt/ccache-4.10/bin/ccache`. Now that value lands in the cache, is immediately shadowed by a fresh `PATH` search, and is ignored — with no warning that the override was discarded.

It is not a cosmetic loss: the discovered value is baked into `MDUX_CACHE_BINARY` for every compile and passed as `-DCACHE_BINARY=` to the integrity test, so bisecting a ccache version regression against #280 no longer works from the command line.

If the intent was to re-run discovery when `CACHE_OPTION` changes, the usual idiom is to keep the cache variable and unset it when the option changes, rather than to bypass the cache entirely:

```cmake
if(DEFINED MDUX_CACHE_OPTION_LAST AND NOT MDUX_CACHE_OPTION_LAST STREQUAL CACHE_OPTION)
    unset(CACHE_BINARY CACHE)
endif()
set(MDUX_CACHE_OPTION_LAST "${CACHE_OPTION}" CACHE INTERNAL "")
find_program(CACHE_BINARY NAMES "${CACHE_OPTION}")
```

tests/CMakeLists.txt
```diff
@@ -7,6 +7,35 @@
...
+# Clang's explicit BMI paths; the Linux Clang workflow installs ccache so this negative layout-change
+# case is continuously exercised rather than left as a local-only reproduction.
+add_test(NAME build.cache.classifier
+    COMMAND ${CMAKE_COMMAND}
+        -DSOURCE_DIR=${CMAKE_SOURCE_DIR}
+        -DBINARY_DIR=${CMAKE_CURRENT_BINARY_DIR}
+        -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/CompilerCacheClassifierTest.cmake)
+set_tests_properties(build.cache.classifier PROPERTIES LABELS "build")
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
```
ambroise-leclerc, line 21: **low — the #280 regression test's registration depends on undeclared variable leakage.**

`mdux_cache_kind`, `mdux_cache_can_hash_bmis` and `CACHE_BINARY` are unprefixed locals set by `include(Cache)` at the root `CMakeLists.txt:188`; they reach this directory only through directory-scope inheritance, and nothing declares that contract.

The failure mode is quiet. If `Cache.cmake` returns early — `ENABLE_CACHE=OFF`, or ccache simply not installed, both of which are normal — or if it is ever wrapped in a function/macro, then `mdux_cache_kind` is undefined and `if(mdux_cache_kind STREQUAL "ccache")` degrades to comparing the *literal string* `mdux_cache_kind`, which is false. The negative test then disappears from the suite and `ctest` reports green with the regression guard absent.

The `--no-tests=error` step added in `clang-build.yml` is a good catch for exactly this, but it only covers that one lane — every other CI lane and every local `ctest` run would show a clean pass with no test.

Promoting these to `MDUX_`-prefixed cache entries or global properties in `Cache.cmake`, set unconditionally (including on the early-return paths), would make the contract explicit and keep the skip deliberate rather than accidental.

.github/workflows/clang-build.yml
```diff
@@ -139,6 +140,11 @@ jobs:
           echo "::error::${clangxx} reports major version ${major}, expected 21."
           exit 1
         fi
+        ccache_version="$(ccache --version | sed -n 's/^ccache version \([0-9][0-9.]*\)$/\1/p' | head -1)"
```
ambroise-leclerc: **low — the version gate is stricter than the CMake logic it mirrors, and hard-fails on a banner it should accept.**

The `sed` expression is `$`-anchored, so it only matches a banner that ends immediately after the digits. Any suffix — `ccache version 4.10.1-dirty`, or a distro/build tag — yields an empty `ccache_version`. `dpkg --compare-versions "" ge 4.8` then returns 1 and the job aborts with:

```
::error::ccache  cannot accept per-compilation BMI inputs; 4.8+ is required.
```

i.e. a confusing "too old" message with a blank version, for a ccache that is perfectly capable.

`cmake/Cache.cmake:40` uses the unanchored `ccache version ([0-9]+(\.[0-9]+)+)` and would have accepted that same ccache and enabled BMI hashing — so CI and the build system disagree about what counts as a supported ccache. Dropping the `$` (or, better, reusing the same regex shape) removes the divergence:

```sh
ccache_version="$(ccache --version | sed -n 's/^ccache version \([0-9][0-9.]*\).*/\1/p' | head -1)"
```

tests/cmake/CompilerCacheIntegrityTest.cmake
```diff
@@ -0,0 +1,140 @@
...
+    "${CMAKE_COMMAND}" -E env
+    --unset=CCACHE_DISABLE
+    --unset=CCACHE_RECACHE
+    "CCACHE_DIR=${fixture_cache}"
+    "CCACHE_CONFIGPATH=${fixture_config}")
+
+run_checked("Fixture configure"
+    ${cache_environment} "${CMAKE_COMMAND}"
+    -S "${fixture_source}" -B "${fixture_build}" -G "${GENERATOR}"
+    "-DCMAKE_CXX_COMPILER=${CXX_COMPILER}")
```
ambroise-leclerc: **low — the fixture does not exercise the toolchain the project actually builds with.**

The fixture is configured with only `-DCMAKE_CXX_COMPILER`, inheriting no `CMAKE_CXX_FLAGS` or standard-library selection. But the `ninja-clang` lane builds MduX against **libc++** — `libc++-21-dev` and `libc++abi-21-dev` are installed in this very workflow for exactly that reason — while the fixture compiles against whatever `clang++-21` defaults to, which is libstdc++ on the Ubuntu image.

Two consequences:

1. The #280 negative case is validated on a configuration CI never ships. libc++ and libstdc++ do not produce identical BMIs or identical `import std` behaviour, so a corruption that only manifests under libc++ would not be caught here.
2. The test breaks outright on any image that has Clang + libc++ but no libstdc++ headers. That failure would surface as `build.cache.namedModuleIntegrity` failing — reading as "ccache module caching is broken" when the real cause is "the fixture picked the wrong standard library".

Passing the preset's `CMAKE_CXX_FLAGS` (and `CMAKE_EXE_LINKER_FLAGS`, if the fixture links) through the `add_test` command in `tests/CMakeLists.txt` and forwarding them here would keep the guard aligned with the build it guards.

ambroise-leclerc (commented):
## Review: BMI-aware ccache launcher

I exercised this locally with ccache 4.12.3 + clang++-21 + Ninja before commenting, and the core mechanism holds up:

- Both new tests pass. The Clang modmap really does list **transitive** BMIs, so a layout change three modules deep still invalidates the consumer object — the `.modmap`-only strategy is sufficient, not just sufficient for depth 1.
- Providers bypass and consumers cache as intended (`cache_miss 3`, `direct_cache_hit 1` on the second identical build), so the launcher keeps the compile-time benefit rather than degenerating into "bypass everything".
- `CMAKE_MATCH_1` **is** cleared on a failed `string(REGEX MATCH)` in CMake 4.2, so the `preprocessed_cache_hit` fallback in the integrity test is not stale-prone.
- `compile_commands.json` does not pick up the launcher, so clangd/clang-tidy consumers are unaffected.
- The Windows `\;` list-escaping in `mdux_join_ccache_extra_files` round-trips correctly through `execute_process`.

Seven findings below, anchored inline. I'd treat the two **medium** ones as merge blockers: both are cases where the stated "fail closed" policy does not actually hold, and one of them can silently corrupt ordinary (non-module) compilations too.

Findings 3-7 are cheap follow-ups and could land separately.

cmake/MduXCompilerCacheLauncher.cmake
ambroise-leclerc: Corrigé dans 4c688d4. Les statuts numériques 0 à 255 sont propagés sans transformation ; seuls les résultats non numériques de démarrage de processus passent par FATAL_ERROR. Le test du classifieur lance un enfant qui sort avec 130 et exige exactement 130 en retour.

cmake/Cache.cmake
ambroise-leclerc: Corrigé dans 4c688d4. CACHE_BINARY est maintenant une surcharge FILEPATH explicite et vide par défaut ; elle est utilisée telle quelle lorsqu’elle est fournie, sinon CACHE_OPTION est recherché à chaque configuration. La fixture épingle ce chemin et vérifie les entrées CACHE_BINARY et MDUX_CACHE_BINARY dans son CMakeCache. La nouvelle option est aussi documentée.

.github/workflows/clang-build.yml
ambroise-leclerc: Corrigé dans 4c688d4. Le sed CI accepte désormais un suffixe après la version numérique, comme la regex CMake. Le cas ccache version 4.10.1-dirty a été vérifié localement et extrait 4.10.1.

tests/cmake/CompilerCacheIntegrityTest.cmake
ambroise-leclerc: Corrigé dans 4c688d4. tests/CMakeLists.txt transmet maintenant CMAKE_CXX_FLAGS et CMAKE_EXE_LINKER_FLAGS à la fixture, qui les applique à sa configuration imbriquée. Le check Linux Clang exécutera donc le cas avec -stdlib=libc++ comme le projet principal. Localement, la fixture Clang/libstdc++ reste verte ; ce poste ne possède pas libc++ 21, donc la combinaison officielle sera validée par CI.

tests/CMakeLists.txt
ambroise-leclerc: Corrigé dans 4c688d4. MDUX_CACHE_BINARY, MDUX_CACHE_KIND et MDUX_CACHE_CAN_HASH_BMIS sont désormais des entrées INTERNAL préfixées, réinitialisées avant tout retour anticipé puis consommées par tests/CMakeLists.txt. ENABLE_CACHE=OFF ou une absence de binaire produit donc explicitement kind=disabled au lieu de dépendre de variables locales absentes.

cmake/MduXCompilerCacheLauncher.cmake
ambroise-leclerc: Corrigé dans 4c688d4. Toute option contenant module, y compris -fprebuilt-module-path, -fmodules-cache-path, -fmodule-map-file et -fno-module-lazy, force désormais le contournement. Tout fichier de réponse @... autre que le .modmap CMake reconnu contourne également le cache. Le classifieur couvre explicitement ces formes ; la fixture confirme encore que le consommateur .modmap reconnu reste réellement caché.

cmake/MduXCompilerCacheLauncher.cmake
ambroise-leclerc: Corrigé dans 4c688d4. Le lanceur ne reconstruit plus argv dans une liste CMake : chaque entrée devient un bracket argument avec un délimiteur choisi pour ne pas apparaître dans la valeur, puis execute_process est évalué. Le test traverse maintenant le lanceur sur la branche cache et vérifie a;b, un argument vide, la liste Windows et une valeur contenant ]].

cmake/MduXCompilerCache.cmake
```diff
@@ -0,0 +1,97 @@
...
+    set(requires_bypass OFF)
+    if(argument MATCHES "\\.(cppm|ixx|mpp)$" OR
+       argument MATCHES "^[-/].*[Mm][Oo][Dd][Uu][Ll][Ee]" OR
+       argument STREQUAL "--precompile" OR
+       argument MATCHES "^[-/]ifc(Output|Only)($|:)" OR
+       argument MATCHES "^[-/]interface$" OR
+       argument MATCHES "^[-/](reference|headerUnit)(:|$)")
+        set(requires_bypass ON)
+    elseif(argument MATCHES "^@" AND NOT argument MATCHES "^@(.+\\.modmap)$")
+        set(requires_bypass ON)
```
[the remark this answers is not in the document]
ambroise-leclerc: Corrigé dans 035e25f. Les formes -working-directory=dir et -working-directory dir déclenchent maintenant toutes deux le contournement du cache avant toute résolution de BMI relative. Le classifieur couvre explicitement les deux syntaxes, et les tests classifieur plus intégrité ccache passent.

AM-L (approved):

PR #291, opened 2026-09-04
Assert the rendered-truth gate on the Windows leg [#282]

### LINKED ISSUE #282: Decide whether the rendered-truth gate should be asserted on the Windows leg too
#255 registers `verify.screen.<id>` and asserts it as a named CI step on three legs: lavapipe on
Linux/GCC 16 and Linux/Clang 21, MoltenVK on macOS. Windows/MSVC **runs** it — it is an ordinary
ctest and that leg's full suite executes it — but no step there asserts it separately.

ADR-007 decision 6 records this deliberately rather than by omission, and states the reasoning: the
four-leg claim is about *bytes*, and Windows carries its full weight there; this gate is about
*pixels*, and the render legs are the ones this repository already treats as owning that question
(ADR-013 makes skipping the pixel suite a failure on macOS, and both Linux legs guard theirs).

## What the asymmetry costs

Since #254 the bake renders, so a Windows runner with no ICD fails to build and nothing is hidden.
The residual gap is narrower: a device that works well enough to build but produces a *wrong* frame on
Windows would be caught by `evidence.screen.<id>`'s byte comparison only because the artifact records
outcomes — and if a Windows-specific misrender ever changed an outcome, that comparison is what would
catch it, not this gate. So the practical exposure is small and already argued in ADR-007.

## What making it four would mean

Adding an asserting step on Windows means treating Mesa's Windows build (pinned by release tag and
SHA-256, registered in the SOUP register by #254) as **gate-critical** rather than build-critical. That
is a real change in what this project depends on for a green CI, and it deserves its own diff and its
own line in the SOUP register's controls column — which is exactly why #255 did not do it in passing.

## If it is taken

- Add the step to `windows-build.yml` mirroring the other three: `ctest --preset ninja-msvc -L verify
  -V --no-tests=error`.
- No skip guard: `mdux-verify-ui` has no 77 status, so an absent device fails the step as itself.
  (See #255's second review round; the `SKIP_RETURN_CODE` that was there initially was dead
  configuration.)
- Update ADR-007 decision 6 **and its consequences paragraph in the same diff**. That pairing is what
  #255's acceptance asked for, and that paragraph has drifted from its own decision once already
  (#246).
- Consider whether the diff-image upload should be added to that leg too.

Found while implementing #255.

ambroise-leclerc:
## Summary

**Taken, not deferred.** #255 registered `verify.screen.<id>` and asserted it as a named step on
three legs — lavapipe on Linux/GCC 16 and Linux/Clang 21, MoltenVK on macOS. Windows/MSVC ran it as
an ordinary ctest inside its full suite and asserted nothing separately.

**Corrected after review.** An earlier revision of this description, and of the records it changed,
claimed that a Windows runner which lost its ICD would have reported the check as skipped and left
the leg green. That was wrong, and so was the promotion claim below it. The leg's `Run Tests` step
has always run `ctest --preset ninja-msvc` unfiltered, the `ninja-msvc` test preset excludes
nothing, and `verify.screen.<id>` is a plain `add_test()` with no `SKIP_RETURN_CODE` — an absent
device exits 3 and CTest reports Failed. The check has therefore always run on this leg and always
failed it. d194ee2 corrects the claim everywhere it was made.

All four bullets from the issue's "If it is taken":

- **The step.** `ctest --preset ninja-msvc -L verify -V --no-tests=error`, placed after the pixel
  step like the other three legs.
- **No skip guard**, because there is no skip to guard against: `mdux-verify-ui` has no 77 status,
  an absent device exits 3, and `cmake/MduXCompileScreen.cmake` therefore sets no
  `SKIP_RETURN_CODE`. A device that vanished between the bake and the tests fails this step as
  itself.
- **ADR-007 decision 6 and its consequences paragraph, in the same diff.** That pairing is what
  #255's acceptance asked for, and that paragraph has drifted from its own decision once already
  (#246). A new risk-and-mitigation bullet covers the dependency this adds.
- **The diff-image upload is added to this leg too** (`verify-diff-msvc`, `build/verify-diff/`),
  last in the job and `if: failure()` — a gate that can now fail on pixels needs its picture
  attached, and `if: failure()` is evaluated where the step sits, so it goes after the last step
  that can fail rather than before it.

### What it actually changes

Not the dependency. Mesa's Windows build — pinned by release tag and SHA-256, registered in the SOUP
register by #254 — has been **gate-critical since #255**, because that is when `verify.screen.<id>`
became a test the Windows full-suite step runs. A green CI has depended on that third-party
rasterizer producing a correct *frame*, not merely on it presenting a device, ever since.

What the named step adds is diagnosis:

- **The gate fails as itself**, rather than as one line inside a full-suite log — what the other
  three legs already give a reviewer.
- **`--no-tests=error` guards the selection.** The one thing the previous arrangement genuinely
  could not catch: if the `verify` label ever selects nothing — a renamed label, a screen dropped
  from the bake — this step fails, whereas an unfiltered run cannot notice, because a test that no
  longer exists is not a test that failed.
- **Diff images are uploaded on failure**, which this leg did not collect before.

A future reversal gives back those three and leaves the rendered check running inside the full
suite. That is recorded in that component's `usage`, `integration_path` and a fourth `risk_controls`
entry in `docs/governance/soup-register.toml`, and in ADR-007 decision 6.

#255's reasoning is **kept rather than overwritten**, in this ADR's house style. The argument was
sound and the exposure it left was genuinely narrow: since #254 the bake renders, so a Windows
runner with no ICD fails to *build*. What was uncovered is a device that builds but *misrenders* —
which `evidence.screen.<id>`'s byte comparison catches only if the misrender changed a recorded
outcome, because ADR-014 decision 4 keeps measurements out of `verification.json`. Asserting the
gate here is what makes a Windows-specific misrender fail as itself.

Leg count also corrected where it is stated elsewhere: ADR-014 decision 6 and its review date,
AGENTS.md's test-label list, and README's target table. `docs/roadmap.md`'s entry for #255 is left
alone — it records what #255 did, which is still true.

Closes #282

Invitation / CLA issue: #282 (maintainer-authored)

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [x] None of the above

## Rebase state

- **Rebased onto:** `48f679d67d23282a7c77b2d7fd5bbafd2ac5e4fc` (develop)

## Verification

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run: no C++, CMake or test source
      changes in this PR.
- [ ] `ctest --test-dir build-gcc` — not run, same reason. The invocation the new step uses was
      exercised in its GCC form on this develop SHA: `ctest -L verify -V --no-tests=error` →
      `verify.screen.endoscope-monitor` Passed, `mdux-verify-ui: OK (5 obligations, 1 render scope)`.
- [x] Other: `check_named_mechanisms.py` OK (79 documents/workflows), `check_file_headers.py`,
      `check_schema_type_drift.py` and the docs-lint unit suite (153 tests) all OK.
      `windows-build.yml` parses as YAML with the expected 19 steps in order;
      `soup-register.toml` parses with 7 entries. **The new MSVC step itself cannot be run
      locally — the Windows leg in CI is what proves it.**

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

Yes, and it is the substance of the change. This alters what a green CI depends on, which is a
change to a verification control rather than to product behaviour:

- **SOUP register** (`docs/governance/soup-register.toml`): `mesa-lavapipe-win` is recorded as
  gate-critical and as having been so since #255, rather than being promoted here. `usage`,
  `integration_path` and `risk_controls` all updated; `classification` stays `build-tool` and
  `runtime_deployment` stays `false`, since nothing about this reaches a device or a shipped
  binary.
- **ADR-007** decision 6, its consequences paragraph, and a new risks-and-mitigations bullet naming
  the failure mode this adds — a defect in a third-party redistribution turning CI red on an
  unchanged commit — and its mitigation.
- **ADR-014** decision 6 and its review date.

No safety behaviour, risk control or traceability record changes, and no certification or
regulatory-compliance claim is introduced: the gate verifies rendered output against committed
goldens, which is evidence of a check having run, not evidence of conformance.


docs/adr/ADR-007-evidence-pipeline-doctrine.md
```diff
@@ -181,21 +181,35 @@ The byte-identity claim survives the addition because the artifact records outco
...
+question. Windows ran the test as an ordinary ctest inside its full suite but asserted nothing
+separately, so a Windows runner that lost its ICD would have reported that one check as skipped and
+the leg would have stayed green on it.
+
+What #282 changed is not that argument but its cost. Adding the fourth assertion promotes Mesa's
+Windows build (pinned by release tag and SHA-256, registered in the SOUP register by #254) from a
+**build-critical** dependency to a **gate-critical** one: a green CI now depends on a third-party
+redistribution of a software rasterizer producing a correct frame, not merely a device. That is the
+whole of the change, it is recorded in that component's controls column in
+`docs/governance/soup-register.toml`, and it is what a future reversal would be giving back.
```
ambroise-leclerc: **[P2] Correct the claimed promotion from build-critical to gate-critical**

The pre-PR implementation already makes successful Windows CI depend on this rendered check passing:

- `.github/workflows/windows-build.yml` already has an unconditional `Run Tests` step invoking `ctest --preset ninja-msvc`.
- The `ninja-msvc` test preset has no filter excluding `verify` tests.
- `cmake/MduXCompileScreen.cmake` already registers `verify.screen.<id>` as an ordinary test invoking `mdux-verify-ui` against the committed bundle, and explicitly sets no `SKIP_RETURN_CODE`. Its existing comment states that an absent device exits 3 and CTest reports it as failed.

Consequently, a missing device at test time or a rendered-check failure already makes the full-suite step fail. The new named step repeats that check; it does not introduce the first gate on correct rendered output. In particular, lines 195–197 retain an incorrect claim that the old test would skip and leave CI green, and lines 206–211 incorrectly describe Windows-specific misrenders as previously uncovered. Those claims also contradict the no-skip behavior correctly described just above in lines 188–190.

Please retain the workflow additions but describe their actual benefits: explicit visibility in the job, `--no-tests=error` protection when the `verify` selection becomes empty, and uploading diff images on failure. Correct the matching build-critical/gate-critical promotion claims in ADR-014, `docs/governance/soup-register.toml`, the workflow comments, and the PR description as well. The SOUP register can record the additional invocation and diagnostics without claiming a newly introduced dependency on correct frames.

This matters because the ADR and SOUP register describe what verification coverage a future reversal would remove: removing only this named step would still leave the existing full-suite rendered check in place.

ambroise-leclerc (commented):
Reviewed the workflow changes and their verification-control documentation at 51fe00f0beed4b8ce6eb1f11025fa95d2e507730. One documentation correctness finding is detailed inline. I found no functional regression in the workflow additions. GitHub reports all four build legs passing, including Windows; I did not run local builds or tests, and Windows execution is unavailable in this Linux workspace.

docs/adr/ADR-007-evidence-pipeline-doctrine.md
ambroise-leclerc: Confirmed and fixed in d194ee2. The finding is right on every point, and I verified each one rather than taking it on the summary:

- `.github/workflows/windows-build.yml:172` — the `Run Tests` step runs `ctest --preset ninja-msvc` unconditionally and unfiltered.
- `CMakePresets.json` — the `ninja-msvc` *test* preset carries only `output.outputOnFailure`; no `filter`, so nothing excludes `verify`.
- `cmake/MduXCompileScreen.cmake:183` — `verify.screen.<id>` is a plain `add_test()`, and the comment at `:189` already states the reasoning: `mdux-verify-ui` has no 77 status, an absent device is `RunState::NoRenderDevice` and exits 3, "CTest reports the run as Failed, which was verified with the ICD unset."

So a lost ICD or a misrendered frame already took the Windows leg red, the ADR contradicted its own neighbouring paragraph, and the dependency has been gate-critical since #255 registered the gate — not since this PR.

Rewritten to claim only what the step actually adds:

- **A named failure** instead of one line inside a full-suite log, which is what the other three legs already give a reviewer.
- **`--no-tests=error` over the selection.** This is the one thing the old arrangement genuinely could not catch: if the `verify` label ever selects nothing — a renamed label, a screen dropped from the bake — the step fails, whereas the unfiltered run cannot notice, because a test that no longer exists is not a test that failed.
- **Diff images uploaded on failure**, which this leg did not collect before.

And the reversal note now says what it should: removing the step gives back those three and leaves the rendered check itself running inside the full suite.

Corrected in all five places rather than only the anchor:

| Where | Was | Now |
| --- | --- | --- |
| ADR-007 decision 6 | "would have reported that one check as skipped and the leg would have stayed green"; promotion to gate-critical | the check always ran; the step is diagnostic, with the three gains listed |
| ADR-014 gate description | "#282 added that step, and ADR-007 records what it costs — becomes gate-critical" | the step did not add the check; names what it does add |
| ADR-014 review timeline | "records as the promotion … to a gate-critical dependency" | "a gain in diagnosis rather than in coverage" |
| SOUP register `usage` | "That second use makes it gate-critical rather than only build-critical" | gate-critical "since #255 registered that gate as an ordinary test"; #282 "changes where the failure shows rather than whether it fails" |
| SOUP register `risk_controls` | "which is what promoted this component to gate-critical" | records the no-skip behaviour, that it has run since #255, and the `--no-tests=error` gap it closes |
| Workflow step comment | "#282 takes the opposite decision … promotes Mesa's Windows build" | "It did not leave the leg unchecked … Mesa's Windows build was already gate-critical" |

`AGENTS.md` and `README.md` needed no change: both say the gate is asserted as a named step on all four legs, which is true and does not assert a promotion.

Checks: `mdux-docs-lint` OK (83 files, 0 findings), `check_file_headers` / `check_named_mechanisms` / `check_schema_type_drift` OK, SOUP register parses with the Mesa entry's four controls intact, workflow YAML parses. No C++ changed.

AM-L (approved):

PR #292, opened 2026-09-04
S2 — SignalTrace: ring-buffer binding and CPU polyline expansion [#257]

### LINKED ISSUE #257: S2 — SignalTrace: ring-buffer binding and CPU polyline expansion

ambroise-leclerc:
## Summary

A `SignalTrace` now draws the waveform inside the field #255 gave it, expanded on the device from a
**caller-owned ring buffer** into the vertex budget the compiled screen already declares.

Three pieces:

- **`mdux.medui.trace`** (new governed module) — the expansion, split from the screen runtime for the
  reason `mdux.text.draw` is split from it: geometry that can be tested against numbers rather than
  only against a rendered frame. `SampleRing` is a *view* of the producer's storage — never copied,
  never owned, and validated on **every** frame rather than once, because the producer moves `oldest`
  and `count` between them. A `create()`-guarded type would be proving things about a snapshot that
  has already moved.
- **`mdux.draw`** — `addSolidQuad()`, the first primitive here that is not axis-aligned, and
  `Point2F` to carry the one coordinate that cannot be integral: a stroke's normal is irrational for
  all but a handful of slopes, and rounding it makes a 1px stroke alternate between one and two
  pixels wide as its slope changes. `addRect()` and `addSolidQuad()` now share a private
  `appendQuad()`, so "past the checks nothing can fail, so the list never holds a half-recorded
  primitive" lives in one place instead of two.
- **`mdux.medui.screen`** — `SignalBinding`, `TextBinding`'s counterpart for signals and deliberately
  a much cheaper object. There is nothing to hash: samples are produced at run time and no committed
  artifact describes them. What *is* checkable it checks once — every slot names a `SignalTrace` this
  screen carries, no two name the same stream, every slot has a usable ring and style — and it
  retains the screen id, so slots validated for one screen cannot render another.

Segments become quads with square joint caps rather than mitred joins, as the epic's design note
asks. The module comment is exact about what that is rather than letting the name carry a claim the
code does not make: a true round join is a disc and this is the disc's bounding square, which at 1–3
px differ by at most the four corner pixels of a 3×3 block. That is why `maxStrokeWidth` is 3 and a
wider stroke is refused. Mitres were rejected for a different reason: the intersection of two offset
lines is unbounded as the angle between segments closes, so a near-reversal produces a spike of no
fixed length — and a spike is exactly the artefact a waveform must not invent.

Closes #257

## Two decisions worth reviewing rather than skimming

**Refused, never truncated.** A ring past `maxSamplesPerTrace` (256) is `TraceTooLong` and the whole
frame rolls back — enforced at the expansion (`trace-refuses-an-oversized-ring`) and again at the
screen (`screen-oversized-trace-refuses-the-frame`). Showing the newest 256 of 300 samples would draw
a perfectly plausible waveform, in the tint that says it is the window the caller asked for, and
nothing on a monitor distinguishes that from a correct reading of different data.

**One tint at two coverages.** This was the hard part, and `Screen.cppm` records the whole argument
rather than only its conclusion. `SignalTraceSpec` carries **one** colour token; an additive draw
list cannot knock a stroke back to the ground; a third colour fails a `ColorHash` golden. So the two
options that file's #255 paragraph names were both unavailable as written, and what was left is the
third reading of the same rule: a bound trace paints its field at `boundTraceFieldCoverage` (25%)
under a full-tint stroke. Every pixel is then still a blend of ground and tint at one coverage, the
stroke supplies the fully-covered pixels `colorHash()` additionally requires, and the field keeps
`goldenBounds()` seeing the node's whole rectangle as painted — which a stroke alone, whose box is
data-dependent, would not.

Arguments about what a check admits belong to the check, so that is also a scenario:
`verify-golden-two-coverage-composition` in `tests/verify/GoldenCheckTests.cpp` paints exactly this
composition, asserts `goldenBounds()` and `colorHash()` both hold, and asserts that the field *alone*
is the `TintAbsent` the stroke exists to answer — so it is not a scenario that cannot fail.

An **unbound** trace is unchanged. That is why the committed screen's pixel test and its `verify` leg
are unchanged too: both render it without signals, and nothing under `generated/` moved.

## No allocation in the per-frame path, verified the three ways #63 established

1. **Runtime interposition** — a new `medui_noheap_spec` scenario writes a new sample into the ring
   and advances `oldest` *between* frames, so a per-frame scratch for the expanded polyline — the
   obvious implementation of a waveform, and the one this measurement exists to refuse — would show
   up as eight allocations rather than none.
2. **Object-file symbol scan** — needed nothing. `screen.noheap.symbolScan` filters MduXCore's
   objects on `/medui/` and picked up `Trace.cpp.o` and `Trace.cppm.o` on its own; both are in the
   generated object list and the test passes.
3. **Static analysis and a stack bound** — this was the layer genuinely missing for the screen
   runtime. #199 gave it layers 1 and 2 and no `.clang-tidy`; `src/medui/.clang-tidy` adds it, with
   the same checks as `src/ml/.clang-tidy` and a header saying why it is a sibling file rather than a
   shared one. `-Wframe-larger-than=4096` already covers these sources via
   `mdux_enforce_fp_determinism(MduXCore)`.

## The ECG demonstrator is wired, not just referenced

`EcgClassifierExample` binds the **same** `SampleRing` its classifier reads its window out of to
`EndoscopeMonitor`'s `ECG_LEAD_II` trace, and records a frame:

```
Frame from the committed screen 'endoscope-monitor' (1280x720):
  samples bound  : 180 (the window the classifier read, not a copy of it)
  traces expanded: 1
  nodes / drawn  : 6 visited, 362 rectangles, 3 deferred
  budget used    : 1448/4096 vertices, 2172/6144 indices, 1/64 commands
```

Nothing is copied between the two readers — `readWindow()` hands the classifier a contiguous ordered
window and `view()` hands the trace a description of the same storage — so a monitor showing a
waveform beside a classification derived from *different* samples is unspellable rather than
unlikely. That file's own comment had asked for exactly this ("a matter of giving them the same
buffer") and could not yet do it; the paragraph saying so has been replaced by the thing it wanted.

Still no Vulkan and no window in that example: a draw list is geometry.

Invitation / CLA issue: #257

## Dependency

- **Base branch:** `develop`

## Shared registries touched

- [x] `CMakeLists.txt` (root) — `mdux.medui.trace` added to `MduXCore`'s module and source lists
- [x] `FILE_SET CXX_MODULES` lists — `include/mdux/medui/Trace.cppm` added; `examples/CMakeLists.txt`
      gains a second file set on `EcgClassifierExample` for the generated screen module
- [x] `tests/CMakeLists.txt` — `medui/TraceTests.cpp` added to `medui_spec`

Nothing under `generated/` changed, and `ctest -L evidence` confirms it: the unbound render path is
untouched, so the committed screen bundle still re-bakes byte-identically.

## Rebase state

- **Rebased onto:** `180fc01` (current `develop`)

## Verification

Run on the verified macOS tuple (upstream LLVM/Clang 21.1.8, `ninja-macos-clang`,
`-DENABLE_CACHE=OFF`) rather than the GCC preset the boxes name, so the GCC lines are left unticked
rather than ticked for a run that did not happen.

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run locally, CI covers it
- [ ] `ctest --test-dir build-gcc` — not run locally, CI covers it
- [x] Other, all on `ninja-macos-clang`:
  - `cmake --build build-macos-clang` — clean, no warnings
  - `ctest` — **698/698 passed**, including `-L noheap` (11), `-L pixel` (2), `-L verify` (1),
    `-L evidence` (7) and `-L determinism` (5)
  - `python3 tools/governed-lint/mdux_governed_lint.py` — OK (36 files, 0 findings)
  - `python3 tools/evidence-lint/mdux_evidence_lint.py` — OK (75 files, 0 findings)
  - `check_file_headers.py` (226 sources), `check_named_mechanisms.py` (78 documents),
    `check_schema_type_drift.py` (7 schemas) — all OK
  - `clang-tidy -p build-macos-clang src/medui/Trace.cpp src/medui/Screen.cpp` — no
    `cppcoreguidelines-no-malloc` or `-owning-memory` findings, which is what the new `.clang-tidy`
    is there to make an error. The `readability-identifier-naming` warnings on enum constants are
    repo-wide and pre-existing (CONTRIBUTING.md's `UpperCamelCase` against the root config), not
    introduced here.
  - `clang-format --dry-run -Werror` clean on every new file and on both `medui` files; the `draw`
    files were left in their existing local style rather than reformatted, which would have been
    unrelated churn.
  - `./build-macos-clang/examples/EcgClassifierExample` — output quoted above

**Not run here:** the Windows/MSVC and Linux/GCC legs, and the lavapipe render legs. The macOS leg
does run `verify.screen.endoscope-monitor` under MoltenVK, and it passes.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

**Safety-relevant, and the impact is deliberate rather than incidental.** This adds a path by which
live physiological data reaches a rendered frame, which is the first such path in this repository.
Three properties carry that, all of them checks rather than review discipline:

- **Refused, never truncated.** A trace whose sample count exceeds the cap refuses the frame, and the
  frame rolls back whole. A partial frame is the worst outcome available on a medical display,
  because it looks like a reading — the rule the screen runtime already applied to text, applied to a
  waveform.
- **Bounded before the device runs.** Per-node work is `maxSamplesPerTrace`, a constant, exactly as a
  `Label`'s is `maxGlyphsPerRun`; the storage is the screen's declared `DrawBudget`, sized once and
  never grown; and no allocation occurs in the frame, verified the three independent ways #63
  established.
- **Excursions are shown, not refused.** A sample outside the declared display range is pinned to the
  band's rail rather than failing the frame — blanking a monitor at the moment the waveform becomes
  most interesting is the wrong failure direction here, and `trace-clamps-excursions` pins the
  decision. A *non-finite* sample is refused, because it makes a position undefined rather than
  extreme.

No new claim of certification, validation or production readiness is made or implied. The ECG
demonstrator's existing notice — synthetic weights, no training, no clinical validity — is unchanged
and still the first thing it prints.

**Artifacts updated:** `AGENTS.md` § 2 (the parity note said `SignalTrace` remains deferred, which is
no longer true), `docs/architecture.md` (module table, the compiled-screen section, the planned
table), `README.md` (two status rows), and
`.agents/skills/medui-authoring/SKILL.md` (its "one limit" paragraph was stale as of #255 and is now
accurate). No ADR needed a change: ADR-011, ADR-012 and ADR-014 all already say what this
implements, and the one constraint ADR-014 imposed on #257 is met rather than amended.

## Summary by CodeRabbit

- **New Features**
  - Added live waveform rendering for `SignalTrace` components using caller-provided sample data.
  - Added signal bindings with validation for stream names, sample ranges, styles, and screen compatibility.
  - Bound traces display a dimmed field beneath a full-color waveform.
  - Added safeguards for oversized or invalid sample data, preserving the frame without partial output.
  - Updated the ECG example to demonstrate classifier data rendered as a trace.

- **Documentation**
  - Updated implementation and architecture documentation to reflect SignalTrace support and deferred components.

### DIFF
--- a/include/mdux/medui/Screen.cppm
+++ b/include/mdux/medui/Screen.cppm
@@ -25,9 +25,14 @@
  *   `textKey`, not glyphs (ADR-011), so drawing one is a join with a baked text package for the
  *   locale the device is running. Without a binding there is nothing to join to and the node is
  *   deferred exactly as before, which is also what a screen with no text costs: nothing.
- * - `NumericDisplay` and `SignalTrace` draw their field, and the **reading inside it** is still
- *   deferred: the digits a template expands to and the excursion a waveform makes are functions of a
- *   sample this module is not given, and they arrive with #258 and #257.
+ * - `NumericDisplay` draws its field, and the **reading inside it** is still deferred: the digits a
+ *   template expands to are a function of a sample this module is not given, and they arrive with
+ *   #258.
+ * - `SignalTrace` draws its field, and since #257 the **waveform inside it** as well, when the
+ *   caller supplies a `SignalBinding` naming that node's stream. Without one the node is exactly
+ *   what it was after #255: the opaque field it reserves. The expansion itself is
+ *   `mdux.medui.trace`'s, split out for the reason `mdux.text.draw` is - geometry that can be tested
+ *   against numbers rather than only against a rendered frame.
  * - `Button`, `CriticalButton`, `TextInput` carry text keys or live sources as well, and are still
  *   deferred whole. Their text is not the whole of their appearance - a button has a face, a text
  *   input has a caret and a selection - and this module will not invent those. They arrive with #17.
@@ -68,6 +73,22 @@
  * colour fails the golden its own screen was compiled with. A component whose field a golden pins
  * with `ColorHash` therefore has two ways to show a reading and no others: in the field's own tint,
  * or knocked out of it back to the ground.
+ *
+ * #257 met that constraint head on and neither option was available as written. An additive draw
+ * list cannot knock a stroke back to the ground - there is no erase - and a stroke in the field's
+ * own tint over an opaque field of that tint is invisible. What was available is the *third* reading
+ * of the same rule, which the sentence above did not have to spell out because nothing had needed it
+ * yet: one tint at **two coverages**. A bound trace paints its field at `boundTraceFieldCoverage`
+ * and its stroke at full tint, so every pixel is still a blend of ground and tint at one coverage -
+ * what `ColorHash` asks - the stroke supplies the fully covered pixels `ColorHash` additionally
+ * requires, and the field keeps `goldenBounds()` seeing the node's whole rectangle as painted, which
+ * a stroke alone would not. An unbound trace is unchanged, which is why the committed screen's
+ * pixels and its `verify` leg are unchanged too: both render it without signals.
+ *
+ * That paragraph is an argument, and arguments about what a check admits belong to the check. So it
+ * is also a scenario: `verify-golden-two-coverage-composition` in `tests/verify/GoldenCheckTests.cpp`
+ * paints exactly this composition and asserts that `goldenBounds()` and `colorHash()` both hold -
+ * and that the field alone, without the stroke, is the `TintAbsent` the stroke exists to answer.
  *
  * ## Where a label's glyphs go, and why that is not an invention
  *
@@ -147,6 +168,7 @@ import mdux.evidence.digest;
 import mdux.font.schema;
 import mdux.image.schema;
 import mdux.medui.schema;
+import mdux.medui.trace;
 import mdux.text.draw;
 import mdux.text.schema;
 
@@ -200,6 +222,14 @@ enum class ScreenError : std::uint8_t {
     TextOverflowsNode,     ///< a run's ink is wider or taller than the node that names it
     ImageSidecarMismatch,  ///< RGBA bytes differ from the baked image package
     ImageNotApproved,      ///< the screen did not approve this image id/digest/extent
+    UnknownStreamSource,   ///< a signal slot names a stream no `SignalTrace` on this screen carries
+    DuplicateStream,       ///< two signal slots name the same stream
+    MalformedTraceStyle,   ///< a slot's sample range is empty or not finite, or its stroke is not 1-3px
+    MalformedSampleRing,   ///< a bound ring's oldest index or live count is not a position in it
+    NonFiniteSample,       ///< a live sample is a NaN or an infinity
+    TraceTooLong,          ///< a bound ring holds more than `maxSamplesPerTrace` samples
+    TraceBandTooSmall,     ///< a bound trace's node is too small to hold its stroke
+    ScreenNotApproved,     ///< a signal binding built for one screen was offered to another
 };
 
 // The two token failures are kept apart because the schema keeps them apart, and for its reason: a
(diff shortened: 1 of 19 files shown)


src/medui/Screen.cpp
```diff
@@ -196,13 +238,60 @@ std::string_view describe(ScreenError error) noexcept {
             return "the screen was not compiled against this text package";
         case ScreenError::TextOverflowsNode:
             return "a run's ink is larger than the node that names it";
+        case ScreenError::UnknownStreamSource:
+            return "a signal slot names a stream no SignalTrace on this screen carries";
+        case ScreenError::DuplicateStream:
+            return "two signal slots name the same stream";
+        case ScreenError::MalformedTraceStyle:
+            return "a signal slot's sample range is empty or not finite, or its stroke width is out of range";
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in b280414, and fixed at the source rather than in the prose.

`SignalBinding::create()` now returns a new `ScreenError::MissingSampleRing` for `slot.ring == nullptr`, so `MalformedTraceStyle` means only what it says: the sample range or the stroke width. Updating the comment would have made the documentation true, but it would have left the two conditions sharing one enumerator, and this runtime states the opposite rule twice:

- `include/mdux/medui/Screen.cppm:235` — "The two token failures are kept apart because the schema keeps them apart... Collapsing them would tell an integrator to look in the wrong place."
- `src/medui/Screen.cpp:294` — `asScreenError()` maps six `TraceError`s one-for-one, because "sending all three to the same integrator would send two of them to the wrong person."

The branch's own test made the case: the scenario covering these refusals is titled *"each refusal names its own cause"* and its `expected` array listed `MalformedTraceStyle` twice. It now expects `MissingSampleRing` at that position.

An absent ring is a slot the caller never finished filling in; a malformed style is one filled in wrongly. `describe(MissingSampleRing)` returns "a signal slot names a stream but carries no ring to read samples from".

716/716 tests pass.

include/mdux/medui/Screen.cppm
```diff
@@ -196,6 +218,14 @@ enum class ScreenError : std::uint8_t {
     SidecarMismatch,      ///< the sidecar is not the one the text package describes
     PackageNotApproved,   ///< the screen was not compiled against this locale/package/digest
     TextOverflowsNode,    ///< a run's ink is wider or taller than the node that names it
+    UnknownStreamSource,  ///< a signal slot names a stream no `SignalTrace` on this screen carries
+    DuplicateStream,      ///< two signal slots name the same stream
+    MalformedTraceStyle,  ///< a slot's sample range is empty or not finite, or its stroke is not 1-3px
+    MalformedSampleRing,  ///< a bound ring's oldest index or live count is not a position in it
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in b280414, and fixed at the source rather than in the prose.

`SignalBinding::create()` now returns a new `ScreenError::MissingSampleRing` for `slot.ring == nullptr`, so `MalformedTraceStyle` means only what it says: the sample range or the stroke width. Updating the comment would have made the documentation true, but it would have left the two conditions sharing one enumerator, and this runtime states the opposite rule twice:

- `include/mdux/medui/Screen.cppm:235` — "The two token failures are kept apart because the schema keeps them apart... Collapsing them would tell an integrator to look in the wrong place."
- `src/medui/Screen.cpp:294` — `asScreenError()` maps six `TraceError`s one-for-one, because "sending all three to the same integrator would send two of them to the wrong person."

The branch's own test made the case: the scenario covering these refusals is titled *"each refusal names its own cause"* and its `expected` array listed `MalformedTraceStyle` twice. It now expects `MissingSampleRing` at that position.

An absent ring is a slot the caller never finished filling in; a malformed style is one filled in wrongly. `describe(MissingSampleRing)` returns "a signal slot names a stream but carries no ring to read samples from".

716/716 tests pass.

The enum comment is updated in the same commit: `MissingSampleRing` carries its own line, and `MalformedTraceStyle`'s is now accurate as written.

ambroise-leclerc (commented):
Two correctness findings reproduced against this commit using a standalone C++ program linked to the existing GCC build of MduXCore. The existing test binaries passed: medui_spec (54 scenarios), draw_spec (29), and medui_noheap_spec (5). I did not perform a full rebuild or run pixel tests. Details and regression-test suggestions are attached inline.

src/draw/Draw.cpp
```diff
@@ -139,6 +141,49 @@ ResultVoid<DrawError> DrawList::addSolidRect(const mdux::core::Rect& rect,
...
+
+    // Twice the signed area, by the shoelace formula over the ring. Zero means the four corners are
+    // collinear or coincident, which is `addRect()`'s DegenerateRect in the shape a quad has it.
+    // The sign is not checked: a caller may wind either way, and the fixed 0-1-2 / 0-2-3 split
+    // draws both - the pipeline disables face culling, so winding is not a visibility question.
+    float doubledArea = 0.0F;
+    for (std::size_t i = 0; i < corners.size(); ++i) {
+        const Point2F& a = corners[i];
+        const Point2F& b = corners[(i + 1) % corners.size()];
+        doubledArea += (a.x * b.y) - (b.x * a.y);
```
ambroise-leclerc: [P2] Compute quad area without cancellation from absolute screen coordinates

The shoelace products are evaluated in `float` using absolute positions. At sufficiently large coordinates, rounding removes the area of a valid small quad, so the subsequent zero-area check rejects it. This affects the new trace path because every sample emits a small cap quad.

Reproduced against the PR's existing GCC library with a fresh DrawList (32 vertices, 48 indices, 4 commands), samples `{0.0F, 1.0F}`, a default TraceStyle (range 0–1, width 1), and band `{4096, 4096, 100, 50}`. `recordTrace()` returns `TraceError::ListRejected` despite sufficient budget and valid geometry. The first cap is the nondegenerate square `(4096,4145)`–`(4097,4146)`, but the area calculation loses its area through cancellation. In the screen path this becomes `BudgetExhausted` and rolls back the entire frame, so merely moving a trace on a large surface can prevent rendering.

Please calculate the area from coordinates translated relative to one corner, using wider intermediates where needed. Add a regression that records the same 1px trace near the origin and at this offset, asserts both succeed, and checks equivalent translated geometry. A direct `addSolidQuad()` test for the translated unit square would also isolate the root cause.

src/medui/Trace.cpp
```diff
@@ -0,0 +1,242 @@
...
+ * is the belt to that pair of braces, because a waveform compared across toolchains cannot afford a
+ * last-bit difference in a vertex.
+ *
+ * Clamping rather than refusing is `TraceStyle`'s documented decision: an excursion past the display
+ * range is a real reading a monitor pins to its rail.
+ */
+[[nodiscard]] Px rowFor(float sample, const TraceStyle& style, Px rows) noexcept {
+    const float span       = style.maximum - style.minimum;
+    const float offset     = sample - style.minimum;
+    const float normalised = offset / span;
```
ambroise-leclerc: [P2] Prevent finite sample ranges from overflowing during normalization

Checking that minimum, maximum, and samples are individually finite does not ensure these float subtractions remain finite. A range of `[-FLT_MAX, FLT_MAX]` passes validation, but `maximum - minimum` overflows to infinity. At the maximum sample, `sample - minimum` also overflows, so normalization produces `inf / inf` (NaN). The following `!(t > 0.0F)` branch silently maps this to the bottom rail.

Reproduced against the PR's existing GCC library with two samples `{-FLT_MAX, FLT_MAX}`, style `{-FLT_MAX, FLT_MAX, 1}`, and band `{10, 20, 100, 50}`. `recordTrace()` reports success, but the first and last cap top edges are both y=69; the expected positions are y=69 and y=20. Thus a minimum-to-maximum transition becomes a plausible flat waveform rather than being rendered correctly or refused.

Please use overflow-safe normalization (for example, widen the operands before subtraction and division), preserving the documented rail-clamping behavior for finite excursions. If a range cannot be supported, reject it explicitly instead of accepting it and mapping an arithmetic NaN to a reading. Add regressions for opposite-sign extreme finite endpoints and interior samples, checking endpoint placement and monotonic mapping.

src/draw/Draw.cpp
ambroise-leclerc: Fixed in 4dacf98. Confirmed, with one correction to the repro.

**The defect is real.** The shoelace sum ran over absolute coordinates, where the two products are of similar magnitude and their difference is the area, so a small quad far from the origin loses that difference to rounding. It now measures relative to `corners[0]` and accumulates in `double`: a float minus a float is exact in double, translation does not change an area, and the products stay small.

**The stated coordinates do not reproduce.** At `(4096,4145)`-`(4097,4146)` the float sum is `4.0`, not `0.0` — a wrong area, but nonzero, so the quad is admitted and `recordTrace()` succeeds. I simulated the exact `float` sequence across offsets:

| offset of the unit square | doubled area in float | exact |
| --- | --- | --- |
| `(0, 0)` | `2.0` | 2 |
| `(4096, 4145)` | `4.0` | 2 |
| `(4096, 4096)` | **`0.0`** | 2 |
| `(65536, 65536)` | **`0.0`** | 2 |
| `(1e6, 1e6)` | `131072.0` | 2 |
| `(8e6, 8e6)` | **`0.0`** | 2 |

So the regression pins `(4096, 4096)`, which cancels to exactly zero and did fail before the fix. Worth noting the `1e6` row: the area is wrong by five orders of magnitude and the zero-check cannot see it — which is the argument for fixing the arithmetic rather than loosening the threshold, as you suggested.

The test records the same unit square at the origin and at the offset and asserts both are admitted, so it fails if the check ever becomes position-dependent again. 718/718 pass.

src/medui/Trace.cpp
ambroise-leclerc: Fixed in 4dacf98, exactly as diagnosed — including the reported y values.

`rowFor()` now widens before subtracting:

```cpp
const double span       = static_cast<double>(style.maximum) - static_cast<double>(style.minimum);
const double offset     = static_cast<double>(sample) - static_cast<double>(style.minimum);
const auto   normalised = static_cast<float>(offset / span);
```

The difference of two floats is exact in double, so nothing that was previously exact is now rounded, and the division is rounded once on the way back to float. Rail clamping is untouched and still handles finite excursions.

I took the widening rather than the reject-the-range option: `[-FLT_MAX, FLT_MAX]` is a degenerate configuration but not a meaningless one, and refusing it would trade a wrong reading for a refused screen when the arithmetic can simply be made to work.

The regression (`trace-extreme-range-separates-rails`) asserts the smallest sample lands on the bottom rail and the largest on the top. Reverting the fix reproduces your numbers precisely:

```
- the largest sample is higher than the smallest: y 69 against 69
- the largest sample is on the top rail, got y 69
```

Both rails now separate: y=69 and y=20.

On determinism, since this changes arithmetic in a path the FP-determinism rules cover: the determinism, evidence and verify labels all pass (5, 654 and 1 tests), and no committed artifact changed — the committed screen renders its trace unbound, so `rowFor()` is not on the bake path. 718/718 overall.

AM-L (approved):

PR #294, opened 2026-09-05
S3 — NumericDisplay and Clock live binding [#258]

### LINKED ISSUE #258: S3 — NumericDisplay and Clock live binding

ambroise-leclerc:
## Summary

A `NumericDisplay` now draws its digits and a `Clock` its time, both through a pattern whose shape a
compiler measured against the node before the device ran.

Closes #258

## Review this first: it needed an amendment to ADR-010

ADR-010 decision 4 forbade "on-device code that maps code points to glyphs, and no on-device code
that advances a pen by a runtime-computed width." Decision 3 of the same ADR permits dynamic text.
Nothing had to resolve that contradiction while no component drew a live value — these two are the
components that do, and `TextBudget.cppm` had already named the gap: `NumericDisplay`'s template
check "waits for the runtime rule rather than guessing at it."

Meanwhile #219 had built `checkClockFormat()`, which computes a clock's worst-case envelope *by
advancing a pen with baked advances and kerning* — a build-time bound that only means something if
the runtime advances a pen the same way.

**The amendment is narrow, and every clause is a check rather than a convention:**

- **The shape is a build-time constant.** A reading is drawn from a *pattern* — `HH:MM:SS`,
  `###.# mmHg` — in which the literals, the slot positions and the glyph count are fixed. Only which
  of ten digits stands in each slot varies.
- **The shape is measured against the node that will hold it.** `measurePattern()` walks the pattern
  considering all ten digits at every slot and returns the envelope of every reading it can ever
  produce; a screen whose box cannot hold that fails to compile.
- **One implementation, shared.** The pen arithmetic is `mdux.medui.reading`'s, and the host budget
  stage imports it rather than carrying its own copy. ADR-008 decision 1 applied to text, and the
  thing that makes the compile-time certificate mean something — a device-time clip the compiler had
  certified is exactly what two copies would eventually produce.
- **Bounded lookup, not a shaping decision.** `FontPackage::find()` over the baked table. No GSUB, no
  GPOS, no contextual form, no fallback — a code point the package lacks refuses the frame.
- **Bounded and allocation-free.** `maxPatternLength` caps per-node work; verified the three ways #63
  established.

What stays forbidden is unchanged, and alternatives A–D in that ADR are rejected still. The honest
cost is written into its **Negative** consequences rather than left to be discovered: a static
`Label` is attested by committed bytes and a live reading by a bound, and a reviewer should know
which they are looking at.

## What landed

**`mdux.medui.reading`** (governed, new) owns the slot model, the integer pen, the worst-case
envelope and the two concrete walks. `tools/medui/TextBudget.cpp` lost its local `clockSlot()`,
`isClockDigitSlot()`, `toPixels()` and envelope walk and imports them instead — so that diff is
mostly deletion.

**A `NumericDisplay`'s `template:` is measurable for the first time.** What it renders as comes from
a `[numericTemplates]` recipe table — the mechanism `charset:` already resolves through — so
`template:` stays a validated *name* in the artifact and the shared contract's compiled-screen
semantics are untouched. An unknown template is `MEDUI-E053`, fail-closed exactly as an unknown
charset name is. That check fired on the committed screen's own compile until the recipe declared
`TPL-PRESSURE-MMHG = "##.# mmHg"`, which is the behaviour working rather than an inconvenience.

**`ReadingBinding`** joins a node to a value and a pattern, and a screen to a time. Its font comes
from the `TextBinding`: carrying its own looked more independent until `needsTextPackageApproval()`
settled it — the schema already requires any screen with a `NumericDisplay` or `Clock` to approve a
text package, so a `TextBinding` is always obtainable and its font is one whose identity `create()`
authenticated. That reversal is recorded in the type's own comment rather than silently landed.

## Three decisions worth reading the code for

**Refused, never truncated.** A value with more digits than its slots is `ValueTooLarge` and nothing
is drawn. Dropping the leading digit shows `09.9` for 109.9 mmHg — a smaller, entirely plausible
number, in a box whose unit says what it means, with nothing on screen to say it was cut.

**The pattern travels with the value, and the runtime re-checks it.** The compiled node carries the
template's name, not its shape, so a device could hold a table the compiler never saw. That exposure
is closed the way a `Label`'s is: the drawn extent is measured against the node and a frame that
overflows is refused as `ReadingOverflowsNode`. `screen-reading-overflowing-its-node-is-refused`
pins it.

**A `Clock`'s tint is the host's.** The shared component model gives `Clock` no `color:` field, so
unlike every other drawn component no artifact names one. `ReadingBinding` supplies a *token*,
resolved through the governed table, so what a host chooses is which approved colour rather than any
colour — and it contradicts no golden, because `collectGoldens()` already refuses `ColorHash` for a
node with no single declared token.

`boundTraceFieldCoverage` is renamed `boundFieldCoverage`: a bound `NumericDisplay` uses the same
one-tint-at-two-coverages composition #257 introduced, and `insufflation-pressure` is in fact the
node `verify-golden-two-coverage-composition` already models.

## Acceptance

- **The widest rendering is measured at compile time against the node's bounds and refused if it
  does not fit** — `measurePattern()` for both components, reported as `MEDUI-E050`.
  `reading-envelope-bounds-every-value` checks the implication that bound rests on by drawing the
  extreme readings and comparing them against the envelope, rather than by arguing it follows.
- **Neither component allocates in a frame** — a new `medui_noheap_spec` scenario advances both the
  value and the clock between frames, so the obvious wrong implementation (a `std::format` scratch)
  would show as eight allocations. Layer 2 already covers `Reading.cpp.o` through
  `screen.noheap.symbolScan`'s `/medui/` filter, and layer 3 through `src/medui/.clang-tidy`.
- **A golden pins bounds and colour but never the varying value** —
  `screen-golden-facts-are-value-invariant` renders one screen with two different readings and
  asserts the panel and the field are byte-identical while the digits differ. A field that resized
  itself around its digits, or tinted itself by magnitude, would make a correct golden describe a
  screen that no longer exists.

Invitation / CLA issue: #258

## Dependency

- **Base branch:** `257-s2-signaltrace-ring-buffer-binding-and-cpu-polyline-expansion`
- **Predecessor PR:** #292
- **Merge order:** must not merge until #292 has merged

Stacked because both change `render()`, `ScreenError` and `Screen.cppm`'s module comment, and this
one renames a constant #292 introduced. A reviewer here sees only #258's diff.

## Shared registries touched

- [x] `CMakeLists.txt` (root) — `mdux.medui.reading` added to `MduXCore`'s module and source lists
- [x] `FILE_SET CXX_MODULES` lists — `include/mdux/medui/Reading.cppm` added
- [x] `tests/CMakeLists.txt` — `medui/ReadingTests.cpp` added to `medui_spec`
- [x] Committed artifacts under `generated/` — see below

**`generated/screen/endoscope-monitor/report.json` only**, and the diff is six lines: the newly
resolved `numericTemplates` option and the recipe's new digest. `package.json` and `goldens.json`
are byte-identical, which is the design working rather than luck — the template stays a name in the
compiled screen, so the artifact the device holds and the golden sidecar a verifier reads are both
unchanged.

## Rebase state

- **Rebased onto:** `155a5d1`, the head of #292's branch

## Verification

Run on the verified macOS tuple (upstream LLVM/Clang 21.1.8, `ninja-macos-clang`,
`-DENABLE_CACHE=OFF`) rather than the GCC preset the boxes name, so the GCC lines are left unticked
rather than ticked for a run that did not happen.

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — not run locally, CI covers it
- [ ] `ctest --test-dir build-gcc` — not run locally, CI covers it
- [x] Other, all on `ninja-macos-clang`:
  - `cmake --build build-macos-clang` — clean, no warnings
  - `ctest` — **716/716 passed**, including `-L noheap` (12), `-L pixel` (2), `-L verify` (1),
    `-L evidence` (7) and `-L determinism` (6)
  - `python3 tools/governed-lint/mdux_governed_lint.py` — OK (38 files, 0 findings)
  - `python3 tools/evidence-lint/mdux_evidence_lint.py` — OK (75 files, 0 findings)
  - `check_file_headers.py` (229 sources), `check_named_mechanisms.py` (78 documents),
    `check_schema_type_drift.py` (7 schemas) — all OK
  - `clang-tidy -p build-macos-clang src/medui/Reading.cpp` — no `cppcoreguidelines-no-malloc` or
    `-owning-memory` findings, which is what `src/medui/.clang-tidy` makes an error
  - `clang-format --dry-run -Werror` clean on every new file, on both `medui` runtime files, and on
    the three `tools/medui` files this touches (all of which were conformant before and are again).
    `tests/medui/ScreenNoHeapTests.cpp` was already non-conformant on `develop` and is left as it
    was rather than reformatted, which would be unrelated churn in an already large diff.

**Not run here:** the Windows/MSVC and Linux/GCC legs, and the lavapipe render legs. The macOS leg
does run `verify.screen.endoscope-monitor` under MoltenVK, and it passes.

**One gap worth naming rather than leaving to be found:** the bound reading path is unit-tested, not
pixel-tested. `ScreenPixelTests` renders the committed screen *unbound*, which is unchanged and still
passes; there is no scenario yet that puts live digits through lavapipe. The unit scenarios check the
glyph rectangles the renderer would consume, and `verify-golden-two-coverage-composition` (#292)
checks that the composition satisfies both golden checks — but neither is a photograph. Worth adding
when a screen with a bound reading is something CI renders.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

**Safety-relevant, and this is the change in the wave with the largest such impact so far**, because
it amends an architecture decision rather than only implementing against one. ADR-010 is a Class B/C
architectural control: "the device runtime contains zero text-shaping code" was a *measurable
absence*, and this PR narrows it to "contains bounded slot substitution in a build-measured pattern".
That narrowing is recorded in the ADR itself — in decision 4 as an amendment with its five
constraints, and in Consequences → Negative as a stated reduction in what committed bytes alone
attest.

The controls that keep the residual risk bounded, all of them checks:

- **The shape cannot vary with data.** Literals, slot positions and glyph count are compile-time
  constants; only digits vary. There is no path by which a value changes what sequence of glyphs is
  drawn.
- **The box is certified before the device runs.** The worst-case envelope over every reading the
  pattern can produce is measured against the node, and a screen that does not fit fails the build.
- **The certificate is not trusted alone.** The runtime measures what it drew against the node and
  refuses the frame on overflow, which covers the one case the build-time check cannot see: a device
  holding a template table the compiler never measured.
- **A wrong number is refused rather than shown.** A value exceeding its slots, a negative value, a
  character the font cannot draw, and a pen that would move behind the node all refuse the frame
  whole. There is no truncation and no substitute glyph — ADR-010 leaves no fallback, and a partial
  or silently-shortened reading on a medical display is indistinguishable from a correct one.
- **No allocation and bounded work per node**, verified three independent ways.

No new claim of certification, validation or production readiness is made or implied. The
demonstrator's synthetic-weights notice is untouched.

**Artifacts updated:** `docs/adr/ADR-010-no-on-device-text-shaping.md` (decision 4 amendment and a
new Negative consequence), `AGENTS.md` § 2, `docs/architecture.md` (module table, compiled-screen
section, planned table), `README.md` (two status rows), and
`.agents/skills/medui-authoring/SKILL.md` (the three things an author needs to know before writing
either component). No other ADR needed a change: ADR-011, ADR-012 and ADR-014 already say what this
implements.

## Summary by CodeRabbit

* **New Features**
  * Numeric displays now render bound values, clocks render time, and signal traces render live waveforms.
  * Numeric templates support measured, fixed-format rendering with validation for missing, oversized, or invalid patterns.
  * Signal traces support caller-provided sample rings, bounded rendering, configurable ranges, and clear validation errors.
  * Bound fields visibly dim beneath rendered readings or traces for improved clarity.
  * Added an endoscope monitor example demonstrating live ECG waveform and reading rendering.

* **Bug Fixes**
  * Invalid or oversized readings and traces are refused without partial frame updates or truncation.


ambroise-leclerc (commented):
Reviewed head f2beae2. Three reproducible correctness findings are detailed inline.

Verification: configured an isolated copy with GCC 16.1.0, CMake 4.2.3 and Vulkan 1.4.341, with examples and compiler caching disabled and the existing local SpecLab source supplied after the network fetch failed. Built `medui_spec` and `mdux-meduic`. Direct execution of `medui_spec` passed all 69 existing scenarios; the added clock-overflow regression failed as expected (70 total, 1 failure). The two compiler reproductions are described inline. CTest found no registered tests in this isolated configuration, so the scenario results reported here are from direct executable invocation. Full build, pixel and no-heap suites were not run in this review.

GitHub currently reports merge conflicts against develop. Earlier successful checks do not establish integration with the current base; resolve the conflicts and rerun the applicable checks. All review fixtures and the temporary regression test were kept outside the working checkout.

tools/medui/TextBudget.cpp
```diff
@@ -55,33 +56,24 @@ constexpr std::array dynamicFields{
...
-    }
-    return slot;
+/// Whether a field names a rendering in the product's numeric-template table (#258).
+///
+/// Separate from `namesDynamicText()` because the two tables answer different questions. A
+/// dynamic-text rule says which code points a name *can produce*, which is a charset claim; a
+/// numeric template says what shape a name *renders as*, which is a geometry claim. A single table
+/// carrying both would let a rule answer one and be consulted for the other.
+[[nodiscard]] bool namesNumericTemplate(std::string_view component, std::string_view field) noexcept {
+    return component == "NumericDisplay" && field == "template";
```
ambroise-leclerc, line 66: [P2] Include numeric templates in the gate that decides whether text packages are loaded

The new predicate is used by `Checker::checkNode()`, but `needsTextBudget()` still checks only `carriesFixedText()` and `namesDynamicText()` (line 624). Consequently, a screen whose only text-bearing component is a `NumericDisplay` is classified as having no measurable text. `Compile.cpp` then rejects the `[text]` table before this new template check can run. Adding a Label or Clock should not be necessary to compile a numeric display.

I reproduced this with the PR's GCC 16.1-built `mdux-meduic`: a 1280x720 screen containing one 300x80 NumericDisplay, using `TPL-PRESSURE-MMHG`, and a recipe with the existing font/text packages and numeric-template table. `mdux-meduic bake review-numeric.toml <output-dir>` exits 1 with:

```text
[MEDUI-E002] this screen carries no measurable text, so its recipe must not declare a [text] table
```

Please include `namesNumericTemplate(node.component, field.name)` in `needsTextBudget()` and add an end-to-end compiler test for a screen with only a NumericDisplay. That test should accept a fitting template with the required text packages and reject a missing template or an overflowing one. The committed screen's other text currently masks this gate omission.

src/medui/Reading.cpp
```diff
@@ -0,0 +1,412 @@
...
+    std::array<std::uint8_t, maxPatternLength> digits{};
+    std::size_t                                count = 0;
+
+    const auto push = [&digits, &count](std::int64_t value, std::size_t width) noexcept {
+        for (std::size_t position = width; position > 0; --position) {
+            std::int64_t scaled = value;
+            for (std::size_t drop = 1; drop < position; ++drop) {
+                scaled /= 10;
+            }
+            digits[count] = static_cast<std::uint8_t>(scaled % 10);
```
ambroise-leclerc, line 399: [P2] Refuse clock fields that exceed their digit slots instead of dropping leading digits

`push()` extracts exactly `width` digits using division and `% 10`, but never checks whether any higher digits remain. The month/day/hour/minute/second fields are `uint8_t`, so they can hold 100–255; their types do not restrict them to two digits. For example, `CivilTime{.year = 2026, .hour = 123}` with `TimeSeconds` successfully records `23:00:00`. The later year check does not cover these fields. This turns malformed clock input into a plausible different time, contrary to the documented refusal-on-digit-overflow contract.

I added a temporary regression scenario using this file's existing font and draw-list fixtures. It calls `recordClock()` with hour 123 and requires an error and an unchanged list. Both expectations fail: the call succeeds and appends glyph vertices. All 69 existing MedUI scenarios pass in the same run, so the current tests do not cover this boundary.

Please validate the digit capacity of every field actually rendered, before recording anything, or reuse a checked conversion equivalent to `digitsOf()`. This does not require calendar validation: allowing an hour of 99 under the documented policy is separate from silently displaying 123 as 23. Add overflow cases for each two-digit field and verify whole-frame rollback through the screen API.

tools/medui/TextBudget.cpp
```diff
@@ -443,103 +435,86 @@ class Checker {
...
-                penMin += minAdvance + minKerning;
-                penMax += maxAdvance + maxKerning;
-                if (penMin < 0) {
-                    report(Code::CharsetEscape,
-                           value.position,
-                           std::format("format '{}' can move its pen left of the clock origin with font package '{}'", value.text, inputs_.font->id));
-                    return;
-                }
-            }
+        measureAgainstNode(node, value, rule->rendering, mdux::medui::PatternKind::Numeric, std::format("template '{}'", value.text));
```
ambroise-leclerc, line 495: [P2] Validate numeric digit-slot limits before accepting the compiled template

The template check only calls `measurePattern()`, which validates pattern length and glyph geometry but does not require any `#` slots or enforce `maxDigitsPerField`. The runtime has stricter structural requirements: `digitsOf()` rejects zero slots with `NoDigitSlots`, and more than ten with `ValueTooLarge`, regardless of the supplied value. Thus a recipe can compile successfully even though its declared numeric rendering cannot display any reading.

Reproduced with the PR's `mdux-meduic`: copy `recipes/screen/endoscope-monitor.toml`, change only `renderings = ["##.# mmHg"]` to `renderings = ["mmHg"]`, and bake to a temporary directory. The compiler exits 0 and reports:

```text
mdux-meduic: OK (compiled endoscope-monitor: 6 nodes, 2 golden references)
```

Binding that declared pattern at runtime necessarily refuses the reading, and the screen renderer rolls back the whole frame. An eleven-slot pattern has the same structural mismatch if its measured extent fits the node.

Please enforce `1 <= number_of_digit_slots <= maxDigitsPerField` for numeric templates during compilation, preferably through a shared structural validator used by both compiler and runtime. Keep the generic clock/pattern measurement semantics separate where appropriate. Add compiler rejection tests for both zero and excessive slots so a successfully compiled template is structurally drawable.

tools/medui/TextBudget.cpp
ambroise-leclerc: Fixed in afb034c. Confirmed exactly as described — `needsTextBudget()` (`TextBudget.cpp:624`) asked only `carriesFixedText()` and `namesDynamicText()`, so a screen whose one text-bearing component is a `NumericDisplay` was classified as carrying no measurable text, and the driver refused its `[text]` table before the template could be measured at all.

`namesNumericTemplate(node.component, field.name)` now joins that condition, which is the fix you asked for and keeps the gate asking through the same predicates the measuring pass uses.

For the test: rather than a full bake, I added `tests/medui/fixtures/accepted-numeric-only.medui` — a screen whose only node is a `NumericDisplay` — and assert the predicate directly, beside the two assertions already there for a text-drawing and a textless screen. That is where this scenario already pins the rule ("the driver must not be the second place that knows what counts as text"), and it fails without the fix, which I checked by reverting it.

I did not add the accept/reject-a-fitting-template end-to-end case here, because the two new scenarios on the sibling thread already cover a missing template and an unmeasurable one through `md::run()` over the committed recipe. Say the word if you want the full numeric-only bake as well.

740/740 pass.

src/medui/Reading.cpp
ambroise-leclerc: Fixed in afb034c, and the second half of your reading was right too — the year check was wrong in the other direction.

`push()` now runs only after every field the format actually renders has been checked for digit capacity:

```cpp
const bool drawable = format == ClockFormat::TimeSeconds
                          ? fits(now.hour, 2) && fits(now.minute, 2) && fits(now.second, 2)
                          : fits(now.year, 4) && fits(now.month, 2) && ...;
```

`fits()` is a plain digit-capacity test, so this stays out of calendar validation exactly as you said: an hour of 99 still draws `99`, and only 100+ is refused.

The standalone `now.year < 0 || now.year > 9999` check is gone rather than kept beside it. It was subsumed for `DateTimeSeconds` (year at width 4 is the same bound) and it was wrong for `TimeSeconds`, which renders no year — so it refused a perfectly drawable time-only clock over a field that never reaches the pattern. That is the second point in Copilot's note on the same line.

Two regressions in `ReadingTests.cpp`: `reading-clock-field-overflow` offers hour 123, minute 199, second 100, month 200 and day 255 — each to a format that draws that field — and requires `ValueTooLarge` with an untouched draw list; `reading-clock-time-only-ignores-year` renders a `TimeSeconds` clock with year 70000 and requires it to draw.

Reverting the fix reproduces your result: the overflow scenario fails and the existing scenarios all still pass, so nothing else covered this boundary. 740/740 pass with it.

tools/medui/TextBudget.cpp
ambroise-leclerc: Fixed in afb034c, through the shared validator rather than a second copy — `countSlots()` is now exported from `mdux.medui.reading` and `checkNumericTemplate()` calls it before `measurePattern()`:

```cpp
const std::size_t slots = mdux::medui::countSlots(rule->rendering, mdux::medui::PatternKind::Numeric);
if (slots == 0) { ... }
if (slots > mdux::medui::maxDigitsPerField) { ... }
```

It already existed in `Reading.cpp`'s anonymous namespace, so this moves it to namespace scope and declares it in the module interface — the compiler and the device now count slots with the same walk, which is the property `measurePattern()` alone could not give. Your framing is what the doc comment on it now says: a compiler that measured only geometry could sign a template the runtime must refuse on every frame.

Generic clock/pattern measurement is untouched — the bound is applied only on the numeric-template path, so `HH:MM:SS` (six slots) and the clock walk keep their existing semantics.

`medui-compile-numeric-template-slots` covers both ends you named: it takes the committed recipe, substitutes `renderings = ["mmHg"]` and then `["###########"]` (eleven), and requires `md::run()` to refuse each. Both pass with the fix and fail without it.

740/740 pass.

tools/medui/Compile.cpp
```diff
@@ -386,6 +400,67 @@ std::optional<Recipe> parseRecipe(std::string_view text, std::string_view recipe
...
+                       std::string{recipePath},
+                       namesLine,
+                       std::format("[numericTemplates] entry '{}' renders as {} characters, and the runtime draws at most {}",
+                                   names[index],
+                                   renderings[index].size(),
+                                   mdux::medui::maxPatternLength),
+                       "the cap is mdux::medui::maxPatternLength, which bounds per-node work on a device");
+                return std::nullopt;
+            }
+            recipe.numericTemplates.push_back(NumericTemplate{.name = names[index], .rendering = renderings[index]});
```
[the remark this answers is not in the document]
ambroise-leclerc: Fixed in afb034c. Confirmed — the `[numericTemplates]` arrays were walked positionally with no name check, and the budget stage resolves a name with `std::ranges::find`, so only the first entry could ever be measured.

`parseRecipe()` now refuses a repeated name rather than de-duplicating it, since which entry the author meant is not knowable from the recipe:

```
[numericTemplates] declares 'TPL-PRESSURE-MMHG' twice, rendering '##.# mmHg' and '###.# mmHg'
```

Your "silently ignoring subsequent entries" is the part that made this worth refusing rather than warning: an author who edited the wrong duplicate would see a template that compiles and draws the other one's shape, with nothing in the output to say so.

`medui-compile-numeric-template-duplicate` duplicates the committed recipe's template name with a different rendering and requires refusal plus a diagnostic naming the repetition. It fails without the fix. 740/740 pass.

src/medui/Reading.cpp
```diff
@@ -0,0 +1,412 @@
...
+            return mdux::core::err(ReadingError::PatternEmpty);
+    }
+
+    // Every field is bounded by its own width above, so a host reporting an hour of 99 draws `99`
+    // rather than overflowing the slot run. What it cannot do is exceed the *pattern*, and a year
+    // past four digits is the one field where that is reachable - so it is checked rather than
+    // assumed, in the same direction as `digitsOf()`.
+    if (now.year < 0 || now.year > 9999) {
+        return mdux::core::err(ReadingError::ValueTooLarge);
+    }
```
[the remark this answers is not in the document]
ambroise-leclerc: Both points fixed in afb034c — see the detailed reply on the maintainer's thread at the same function.

Truncation: `recordClock()` now checks every field the format renders for digit capacity before `push()` runs, so hour 123 is `ValueTooLarge` instead of drawing `23`. The field types are what made this reachable — month, day, hour, minute and second are `std::uint8_t`.

The year check: you were right that it should not run for `TimeSeconds`. The standalone `now.year > 9999` test is removed; the year is now checked only as part of the `DateTimeSeconds` field set, where it is actually rendered. A time-only clock with an out-of-range year now draws, and `reading-clock-time-only-ignores-year` pins that.

740/740 pass, both regressions confirmed to fail without the fix.

ambroise-leclerc (commented):
Follow-up review of **be4c66097c3ca66976b43327710364e54663d173**, including the fixes and the final diff against `develop` after the rebase.

**All three findings from my previous review are addressed. I found no new actionable correctness issue in this pass.**

- **Numeric-only compilation:** `needsTextBudget()` now includes `namesNumericTemplate()`. I reran the original end-to-end reproduction with one NumericDisplay, its template table and the required text packages. It now compiles successfully, producing one node and no golden references. The new predicate regression also passes.
- **Clock overflow:** each field actually used by the selected format is checked before digit extraction. The new overflow scenario rejects oversized hour, minute, second, month and day values with `ValueTooLarge` and records no vertices. The time-only scenario also confirms that an unused out-of-range year does not reject an otherwise drawable time. Both pass locally.
- **Undrawable numeric templates:** the compiler now uses the governed `countSlots()` and the runtime's `maxDigitsPerField`. Rerunning the original `mmHg` reproduction now fails with MEDUI-E053; an eleven-slot rendering fails with the diagnostic explicitly naming the ten-slot limit. The added compiler regressions pass.
- **Additional duplicate-name fix:** duplicate template names now fail during recipe parsing. I independently exercised two different renderings under the same name and confirmed MEDUI-E002 identifies the duplicate.

I also checked the reading paths after integration with the image binding, the whole-frame refusal paths, and the unchanged runtime bounds/budget checks.

Local verification used an isolated copy, GCC 16.1.0, CMake 4.2.3 and the installed Vulkan 1.4.341 SDK. Configuration used Ninja, `MDUX_BUILD_EXAMPLES=OFF`, `ENABLE_CACHE=OFF`, and `FETCHCONTENT_SOURCE_DIR_SPECLAB` pointing to the existing local SpecLab checkout. Built `medui_spec`, `medui_tools_spec`, `medui_noheap_spec` and `mdux-meduic` successfully.

CTest results:

- `medui_spec`: **75/75 passed**.
- `medui_tools_spec`: **131/131 passed**.
- `medui_noheap_spec`: **6/6 passed**, including the live reading/clock allocation scenario.
- `screen.noheap.symbolScan`: **passed**.

Documentation lint passed over **83 files**, and `git diff --check` was clean. Recompiling the committed endoscope screen produced byte-identical `package.json` and `goldens.json`. The compiler-stage report content also matches; the complete committed report additionally contains metadata from the later `mdux-verify-bake` stage, which was not rerun locally.

Limits: this was a focused review, not a full local build/test run. I did not rerun local pixel/render verification, the complete staged evidence pipeline, Windows or macOS. GitHub reports this exact head as mergeable and reports successful Linux GCC/Clang, Windows, macOS, sanitizer and lint checks. No working-checkout files were modified.

AM-L (approved):

Issue #259, opened 2026-08-28
S4 — StatusIndicator with a closed states list


ambroise-leclerc:
Part of #17.

## Work

`StatusIndicator`, taking a `states: [t(…), …]` list. Every state is a locale-checked string, so the
compiler measures **the widest state across every approved locale** against the node's bounds — the
same worst-case rule #195 established, applied to a set rather than one run.

## Acceptance

- [ ] A state list whose widest member overflows the node is a compile error naming the state, the
      locale and the overflow.
- [ ] The `states` list is closed at compile time; a runtime index outside it is refused rather than
      drawn blank.
- [ ] The ECG demonstrator has a waiting consumer for this — wire it.

PR #296, opened 2026-09-05
S5 — TextInput, display and caret only [#260]

### LINKED ISSUE #260: S5 — TextInput, display and caret only
Part of #17.

## Work

`TextInput`, **display and caret only**.

## The scope cut, restated so it is not re-litigated by accident

**No IME.** Input-method editing is a platform concern that does not belong inside a governed
renderer, and pulling it in would breach the trust-zone boundary from #11: governed modules are
compiled without access to platform, graphics or OS headers, and an IME needs all three. The cut is
architectural, not a matter of effort.

The glyph set stays an open name (`charset`), unlike `ClockFormat` and `SystemEvent` in #219 —
TrustSC leaves it open too, so this is the one of the three where MduX already agrees.

## Acceptance

- [ ] Display and caret render; no input method, no composition, no candidate window.
- [ ] The restricted charset from #14 bounds what can be displayed, and a format that could escape
      it is a compile error rather than a runtime surprise.

ambroise-leclerc:
## Summary

A `TextInput` now draws the value a host gives it and the caret that host is editing at, on a grid
whose worst case a compiler measured against the node before the device ran.

Closes #260

## Review this first: it extends ADR-010 decision 4 a second time

#258 amended decision 4 to admit **slot substitution in a measured pattern** — literals and slot
positions fixed at build time, only the digit varying. That does not describe a `TextInput`: its
value is characters from an open-ended charset, and no fixed string of literals describes what an
operator types.

What does carry over is the clause the amendment actually turns on — *slot positions are build-time
constants* — applied to a **fixed pitch**. `mdux.medui.field` puts cell *k* at `k * cellWidth(font)`
and only which permitted character occupies a cell varies. The alternative was a proportional pen,
which is precisely what the unamended decision forbids: cell 5 would sit where characters 0 to 4 put
it, which is a runtime-computed width.

Two clauses are worth reading in their new form:

- **The shape is fixed at build time, and more strictly than a pattern's** — a cell's position
  depends on nothing but its index and one number derived from a committed font package.
- **The measurement and the placement share one implementation *and one input*.** `cellWidth()` is
  derived from the font package rather than supplied by a host, so the compiler's number and the
  device's are the same number whenever the bytes are. No product table travels beside the screen —
  which is the exposure `ReadingSlot` documents at length for a `templateId`, avoided here rather
  than mitigated.

The honest cost is in the ADR rather than left to be found: a proportional font on a fixed pitch
looks monospaced. That is the appearance price of a placement a compiler can certify.

## `max_length` is measured, which closes the last gap TextBudget.cppm names

That file has carried this admission through three issues: the check "waits for the runtime rule
rather than guessing at it", then `TextInput` "still waits, and now for a smaller reason", then a
pointer to this issue. The obstacle it named was that "the worst case is the widest twelve glyphs the
charset admits, in an order nothing constrains".

The grid removes the order from the question. The worst case is twelve cells of the font's widest
permitted glyph plus the caret's column, and `measureField()` is that number — computed in the budget
stage through the same function the device calls. A box that cannot hold it is `MEDUI-E050`; a
`max_length` past `maxFieldCells` (64) is `MEDUI-E053` instead, because no box makes it drawable.

The node's own `charset:` deliberately does not narrow the measurement. A named charset narrows what
may *appear* and never widens it, so a box holding the font's widest glyph holds every glyph a
narrower set admits — conservative in the only direction that cannot produce a device-time overflow.

## Three decisions worth reading the code for

**Refused, never truncated or substituted.** A value longer than its field draws nothing: a shortened
patient identifier is a different identifier that looks like a whole one. A character the package
cannot draw refuses the frame *including the characters before it*, because showing the prefix is
truncation by another route.

**The value crosses as code points, not bytes.** Decoding UTF-8 is parsing, the governed zone does
not parse (ADR-004), and a decoder here would be the first step of the on-device text machinery
ADR-010 exists to keep out. The host has the bytes and the encoding; this takes what they mean.

**No box, no border.** A `TextInput`'s single token is its *text* colour, exactly as a `Label`'s is,
so this module does not paint a face for it. That is the line #255 drew — it painted a field for the
two components whose golden entry pins their whole rectangle against one token, and a `TextInput`'s
does not.

## Acceptance

- **Display and caret render; no input method, no composition, no candidate window** — the component
  has no key handling of any kind, and ADR-010's "No IME integration" consequence now records that
  the landed component honours the cut. `An authored screen's bound text field reaches the pixels`
  renders the committed screen's field under lavapipe and checks the caret's column three pitches
  from the node's left edge, spanning the full envelope height, absent when no caret is bound, with
  glyphs that do not move when it appears.
- **The restricted charset bounds what can be displayed, and a format that could escape it is a
  compile error rather than a runtime surprise** — the charset-escape check has been the budget
  stage's since #195 (`medui-textbudget-charset-escape`, `-charset-gap`, `-charset-noncharacter`);
  what this PR adds is the geometry half, `medui-textbudget-field-length` and
  `-field-length-cap`. At run time the second line is `GlyphNotInPackage`, with no fallback:
  `medui-field-refuses-a-glyph-the-package-lacks`.

Thirteen new scenarios: ten in `tests/medui/FieldTests.cpp`, one no-heap scenario editing the value
between frames, two in the budget suite, plus the pixel scenario. **Six mutations confirm the
behavioural claims** — truncating a long value, substituting an unknown glyph, a proportional pen, a
half-height caret, and dropping the compile-time check each fail exactly the scenario that names it.

Invitation / CLA issue: #260

## Dependency

- **Base branch:** `develop`

Cut from the `develop` head that contains #259 (PR #295), which merged before this branch was made.

## Shared registries touched

- [x] `CMakeLists.txt` (root) — `mdux.medui.field` added to `MduXCore`'s module and source lists
- [x] `FILE_SET CXX_MODULES` lists — `include/mdux/medui/Field.cppm` added
- [x] `tests/CMakeLists.txt` — `medui/FieldTests.cpp` added to `medui_spec`
- [x] Committed artifacts under `generated/` — see below

The screen gained a node, so `generated/screen/endoscope-monitor/` (`package.json`, `report.json`,
`verification.json`) is re-baked; the text package is untouched, because a `TextInput` carries no
text key. **`goldens.json` is byte-identical** — the new node is neither positioned nor
`@safety_critical`, and the label whose width changed never had an entry. `verification.json`'s five
outcomes are unchanged for the same reason `StatusIndicator`'s were: a `TextInput` contributes no
`textKeyOf()` and its `colorTokenOf()` is not pinned by any golden on this screen.

Re-baked with `mdux-bake-update` and driven to a fixpoint.

**One edit beyond the issue's scope, called out rather than slipped in:** AGENTS.md's module table
listed no `mdux.medui.*` module at all, so adding only the new one would have implied it was the
only one. The five that exist are now listed. That table is documented as verified against
`CMakeLists.txt`, and it was not.

## Rebase state

- **Rebased onto:** `2c72a9b`, the head of `develop`

## Verification

Linux, GCC 16.1.0, CMake 4.2.3, Ninja — the `ninja-gcc` preset's own `build-gcc` tree, generator,
compiler and build type, configured before this branch rather than freshly by preset, so the first
box is left unticked rather than ticked for a command that was not the one typed.

- [ ] `cmake --preset ninja-gcc && cmake --build build-gcc` — `cmake --build build-gcc` on the
      equivalent existing tree, clean, no warnings
- [x] `ctest --test-dir build-gcc` — **765/765 passed**, including `-L noheap` (14), `-L pixel` (2),
      `-L verify` (1), `-L evidence` (8), `-L determinism` (6) and `-L governed` (2)
- [x] Other:
  - `check_named_mechanisms.py` (79 documents), `check_schema_type_drift.py` (7 schemas) — OK.
    `check_file_headers.py` reports one finding, in the gitignored `examples/build/` tree, present
    before this branch.
  - `mdux-bake-update` run to a fixpoint; `ctest -L evidence` green over the committed artifacts.
  - `./build-gcc/examples/EcgClassifierExample` — runs; its new field is deferred there, and its
    printout says so, because that program opens no files and so binds no locale.
  - `git clang-format` applied; clean on every line this branch touches.
  - The six mutations above, each reverted after.

**Not run here:** the Windows/MSVC, macOS/Clang and Linux/Clang legs. The Linux lavapipe render legs
*were* run — `verify.screen.endoscope-monitor` and `offscreen_tests` both pass locally under radv.

**A limit worth naming rather than leaving to be found:** the pixel scenario pins the caret's
geometry and that the glyphs do not move when it appears, but it does not prove the *glyphs* are on
the grid — `A`, `B` and `7` have equal advances in the committed font, so a proportional pen would
render that scenario identically. `medui-field-grid-is-fixed-pitch` is what covers the grid, over a
fixture font whose `i` and `W` differ, and the proportional-pen mutation fails there and only there.

## Post-merge gate

      dependent PR may merge.
- [ ] That run is green.

## Regulatory impact

**Safety-relevant, and it amends an architecture decision rather than only implementing against
one** — the second such extension of ADR-010 decision 4, after #258's. The narrowing is recorded in
the ADR itself, in decision 4 as a `#260` extension with its clauses, and in the "No IME integration"
consequence.

The controls that bound the residual risk are checks rather than conventions:

- **The placement cannot vary with data.** A cell's position is its index times one number derived
  from a committed font package. There is no path by which a value moves a character.
- **The box is certified before the device runs**, and by the same arithmetic the device uses: the
  budget stage calls `measureField()`, the same function whose grid `recordField()` places on.
- **The certificate is not trusted alone.** The runtime measures what it drew against the node and
  refuses the frame on overflow, which covers the case the build-time check cannot see — a device
  holding a font package the compiler never measured.
- **A wrong value is refused rather than shown.** Longer than the field, a caret outside it, or a
  character the package cannot draw all refuse the frame whole. No truncation, no substitute glyph,
  no partial field.
- **The scope cut is architectural and now mechanically true**: there is no key handling, no
  composition state and no candidate list anywhere in the governed zone, and `mdux_verify_trust_zones()`
  is what would make adding one a configure-time error rather than a review question.

**No new certification or compliance claim.** A bounded input field that displays what a host tells
it to display is a rendering mechanism; it is not a claim about data entry, validation, or the
clinical safety of any identifier it shows.

## Summary by CodeRabbit

- **New Features**
  - Added bounded `TextInput` fields with fixed-width text and caret rendering.
  - Added support for displaying host-provided values, including patient ID on the endoscope monitor.
  - Added validation for field length, supported characters, caret positions, and available space.
  - Text fields render without editing, keyboard handling, or input composition.

- **Bug Fixes**
  - Prevented oversized or undrawable text fields from rendering partially.

- **Documentation**
  - Updated architecture, roadmap, and implementation status documentation to reflect shipped `TextInput` support.


ambroise-leclerc:
Thanks — four of the five findings were real and are fixed in `9380a71`. The first was a defect on the committed screen, not a hardening: a patient identifier beginning with `J`, `T`, `Y`, `_` or `j` blanked the display, because `measureField()` reserved room for a negative left bearing and `recordField()` did not spend it, so the frame was refused as `FieldOverflowsNode`. The overhang is now computed once and used by both, and `fieldOriginX()` is exported so a caller checking a caret's column uses the same number instead of deriving it again.

The fixture font had no negative-bearing glyph, which is why the suite missed it; it has one now, modelled on the committed package.

Also fixed: `recordField()` now asks `permits()` rather than `find()`, so the charset bounds what can be displayed at run time and not only at compile time; a charset whose every glyph is blank is `CharsetHasNoInk` from both entry points, instead of measuring as height zero and then silently drawing no caret; and the `max_length` cap diagnostic no longer blames the font package for a limit no font can change.

**Not taken: enforcing a node's declared `charset:` inside `render()`.** The compiled node carries `charset` as a validated *name*, not a set, so the runtime would need a product charset table shipped beside the screen — the exposure `Field.cppm` and ADR-010 decline for the reason `ReadingSlot` gives about a `templateId`. The split is deliberate: the compiler checks the named charset against the font package (`MEDUI-E053`, since #195), and the runtime's bound is the package's own charset, which is now enforced. A device holding a package the compiler never measured is caught by the drawn-extent re-check rather than by a table it would have to be told about.

768/768 pass, and two mutations confirm the functional fixes are load-bearing: reverting the origin fails the new regression plus both grid scenarios, and reverting the `permits()` check fails the charset scenario.

ambroise-leclerc:
Both confirmed. `5b1fd86`.

**P2 #1 — TextInput-only screens cannot compile.** Right, and worse than reported: the screen was uncompilable *both* ways. With a `[text]` table it was `MEDUI-E002` "carries no measurable text"; without one it was `MEDUI-E000`, an **internal error**, because `needsTextPackageApproval()` puts `TextInput` among the components whose screen must approve a text package. Two predicates disagreeing about the same component, telling the author to report a compiler defect whichever way they turned. Reproduced with `mdux-meduic` exactly as you did, both directions, before fixing.

`namesFieldLength()` is now in the predicate, and `medui-compile-textinput-only-screen` is the end-to-end test you asked for: it compiles such a screen through `md::run()`, then narrows the box and requires `MEDUI-E050`, so it fails if the field stops being measured as well as if it stops compiling. Reverting the one-line fix fails it.

This is the second time the hole has been dug — #258 hit it with `NumericDisplay`, and the comment recording that sits three lines above the list I failed to extend. The comment now states the rule for the next component rather than only the history: the measuring pass, this predicate and `needsTextPackageApproval()` are one decision asked in three places.

**P2 #2 — the node's `charset:` does not restrict displayed values.** Correct, and the documentation implied otherwise, which is the part I've fixed here. The skill said "your `charset:` narrows what may appear"; what it narrows is what the *source* may produce, which the compiler checks against the font package. A compiled node carries the charset's name and not its set, so the runtime's bound is the package's own charset — wider than the node's whenever an author narrowed it — and your digits-only-field-displays-`A` case is exactly right.

I have not implemented runtime enforcement in this PR, and I want to be straight about why rather than resting on the reason I gave last round. "Preserve the resolved set" means the compiled screen carrying resolved ranges: a change to `TextInputSpec`, the package JSON, both emitters, `validate()`, and therefore to the shared contract's compiled-screen semantics. That is the class of change #258 declined for `templateId`, and #17 records that `charset:` deliberately stays an open name rather than a closed set like `ClockFormat`.

But the analogy to `templateId` is weaker than I claimed last round, and I should say so: a template name stands for a product's *rendering*, while a charset name stands for a set the compiler has **already resolved and already validated** — carrying it in the artifact is closer to carrying resolved `bounds` than to carrying a product table. That is a real argument for doing it, and it is upstream's as much as this repository's.

So it is #297 rather than a comment thread, with the decision written out. Meanwhile the limit is stated in `Field.cppm` and in the authoring skill instead of being left for an author to discover from a screen.

769/769 pass.

AM-L (approved):

